PR #3, opened 2019-07-01
forked regional_workflow for non-CONUS domains

MatthewPyle-NOAA:
Most changes are isolated to new domain specific parameter files and rocoto drivers, but does touch the jobs to add a domain variable to certain path definitions.  Should be safe to merge, but would appreciate someone else looking it over.


MatthewPyle-NOAA:
> 
> 
> Wondering if we need a separate XML for each domain or one XML template that build on the fly based on whatever domain we run with. I think this would make maintenance easier.

I'd had a similar thought - just put together a prototype that I've now pushed to my fork.

PR #21, opened 2019-07-15
update script for gitflow naming

JacobCarley-NOAA:
Updates the update_fork.pl script and moves it to the top level of the directory.


JacobCarley-NOAA:
Reference most recent commit - changes present there.  Forgot to perform a git add.

PR #22, opened 2019-07-16
Feature/buildport

BinLiu-NOAA:
Pull Request from the feature/buildport branch.
Modify and standardize the build system and add support for multi-platforms.
    a. Use the naming convention of regional_*.fd for the workflow source code components.
    b. Switch to checkout the proper branches for the source code components with added support for other platforms.
    c. Modify and make the build scripts working
    d. Use the naming convention of regiong_*.x for executables

README for how to build and install:
cd sorc
./checkout.sh
./build_all.sh
./install_all.sh
./link_fix.sh


ywangwof:
Will it be better to call "link_fix.sh" in "install_all.sh"?
In "checkout.sh", there is still not tags or hash # specified. It cannot guarantee the same source code between any two calls of the script. Please add hash # (or tag) for each "*.fd" cloning command.

Issue #42, opened 2019-07-23
Import the regional_grid into the build system


BinLiu-NOAA:
Import the regional_grid into the build system and the corresponding source code will be included in a branch of UFS_UTILS. Make sure it supports theia, wcoss_dell_p3 and wcoss_cray.

JacobCarley-NOAA:
Was merged into develop at [94f30e7](https://github.com/NOAA-EMC/regional_workflow/commit/94f30e7fb5f03ab69a75b933e5ce5f472ff949ae)

PR #43, opened 2019-07-23
Feature/buildregionalgrid

BinLiu-NOAA:
This PR is related to #42.

Import the regional_grid into the build system. And the corresponding source code has been added into the UFS_UTILS branch. Have tested building on theia, wcoss_cray, and wcoss_dell_p3. But, have not tested running the executable for a test case.


JacobCarley-NOAA:
I went ahead merged it in since there's no scripts yet that run these execs in develop and therefore safe to merge.

PR #71, opened 2019-08-23
Merge feature/merge_comm_to_emc into develop

BenjaminBlake-NOAA:
Since the workflow with the new xml template in the feature/merge_comm_to_EMC branch has been successfully tested on both the WCOSS Dell and Theia, I would like to completely overwrite what is currently in develop.  We have an old version of the develop branch saved as a release (the version of the EMC workflow before the code sprint).  There will probably be conflicts in this merge process that will need to be resolved.


BenjaminBlake-NOAA:
Thanks for doing that Bin!  You were faster than I was.  I'll go ahead and merge the two branches now.  I retrieved a copy of the current develop branch and saved it on WCOSS in case something happens.

PR #91, opened 2019-09-12
Update CCPP build script and remove duplicate namelist

BenjaminBlake-NOAA:
Hi Matt, when you get a chance can you check out the code from the develop branch of my fork and see if you are able to compile the CCPP forecast executable?  I had to add STATIC=Y and SUITES=FV3_GFS_2017_gfdlmp_regional in order for it to compile.  I also removed input_sar.nml because it is identical to input_sar_conus.nml.  Thanks!


BenjaminBlake-NOAA:
Thanks Matt!  Yes that's correct it will become the default build option at some point.

PR #98, opened 2019-09-14
Change dnats and consv_te to 0 in GFS physics namelist template

JeffBeck-NOAA:


climbfuji:
See my comment to your email, dnats has a physical meaning and as far as I know should be set to one for GFDL-MP and to zero for all other MP schemes.

PR #122, opened 2019-10-09
Changes to run with latest NEMSfv3gfs master code

BenjaminBlake-NOAA:
A new section has been added to the namelists.  Also, the path to NDATE and NHOUR on Hera was updated in JREGIONAL_ENVIR.


BenjaminBlake-NOAA:
This commit also includes changing dt_atmos to 30 and k_split to 2.  This will result in the physics being called 3 times as frequently (dt_atmos was previously set to 90 seconds).

PR #125, opened 2019-10-10
Update version of ESMF from 7.1.0r to 8_0_48b.netcdf47.

JulieSchramm:
Hera update from EMC uses ESMF 8_0_48b.netcdf47

### DIFF
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -1,14 +1,16 @@
 [ufs_utils_develop]
-branch = dtc/develop
 protocol = git
 repo_url = https://github.com/NCAR/UFS_UTILS
+#Working hash of develop branch
+hash = 4643e2c
 local_path = sorc/UFS_UTILS_develop
 required = True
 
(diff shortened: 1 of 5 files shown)


gsketefian:
Just wondering why this wasn't a problem for me on hera.

llpcarson:
I think this is needed for the latest NEMSfv3gfs, so may show up for you when that's updated.

climbfuji:
The target date for the offical 8.0.0 release is mid October - we should update to this version. You can use bs50 for now, which is pretty much identical to what will be released.

JulieSchramm:
Note that this update is only for the make_sfc_climo task.  This is so the version of EMSF used for the build from the regional workflow matches what is run from the UFS_UILTS develop branch.

climbfuji:
> Note that this update is only for the make_sfc_climo task. This is so the version of EMSF used for the build from the regional workflow matches what is run from the UFS_UILTS develop branch.

Got it, thanks for the clarification.

PR #139, opened 2020-01-30
Maintain functionality of regional workflow develop branch

BenjaminBlake-NOAA:
Summary of the changes:

- Update modulefiles to load esmf/8.0.0 instead of esmf/8.0.0bs21
- Need to update Hera modulefiles to load correct version of netcdf - Ratko can you do this?
- Remove tpp=2 from forecast job resource request.  Don't need it since OMP_NUM_THREADS is set to 2.
- Fix install_all.sh to only retrieve executable if it was built.  Previously, the script would fail if it couldn't find it.
- Set realtime = "F" in regional_workflow.xml.in.  Otherwise if it is true, it will likely say the cycle has expired.


BenjaminBlake-NOAA:
Hey Matt, I simply grabbed the latest suite definition file from the ufs-weather-model repository and added the fast_physics_nml section since we need that for our runs.  I also had to change ozphys to ozphys_2015 because in our namelist we have ozphys = F and ozphys_2015 = T.  Not sure which one we should be using.

BenjaminBlake-NOAA:
Ratko I just approved the changes, so once you approve the pull request I will merge the changes into the develop branch.

PR #154, opened 2020-02-19
Add predefined hrrr domains to latest workflow

mkavulich:
This replaces PR #132.  

Adding the pre-defined domains for HI, PR, and GU to set_predef_grid_params.sh, and adding JPgrid write component configuration for AK. 

The values for these domains have been taken from fix files generated for old output grid, originally located on Hera at /scratch4/NCEPDEV/fv3-cam/save/Benjamin.Blake/fix_fv3cam/fix_sar but have since been scrubbed. I archived a copy here: /scratch2/BMC/det/kavulich/fix/fix_sar

## Tests run:
- [x] Test end-to-end through post
  - [x] Alaska
  - [x] Hawaii
  - [x] Puerto Rico
  - [x] Guam
- [x] View all domains to ensure they are geographically "sane"

## Feedback requested:

 - This AK domain may be a somewhat duplicate of the "GSD_HRRR_AK_3km" domain; let me know if it should be omitted
 -  I went with a pretty conservative number of nodes so currently all of these cases take a few hours to run. If the number of nodes should be increased let me know.

### DIFF
--- a/ush/UFS_plot_domains.py
+++ b/ush/UFS_plot_domains.py
@@ -161,7 +161,7 @@ def get_lambert_points(gnomonic_map, lambert_map,pps):
 
 # Call the function we just defined to generate a polygon roughly approximating the lambert "rectangle" in gnomonic space
 
-verts3,codes3=get_lambert_points(map2, map3,10)
+verts3,codes3=get_lambert_points(map1, map3,10)
 
 # Now draw!
 
--- a/ush/UFS_plot_domains.py
+++ b/ush/UFS_plot_domains.py
@@ -1,0 +1,175 @@
+#!/usr/bin/env python
+
+import matplotlib.pyplot as plt
+from mpl_toolkits.basemap import Basemap
+from matplotlib.path import Path
+import matplotlib.patches as patches
+import numpy as np
+
+#### User-defined variables
+
+
+# Computational grid definitions
+JPgrid_LON_CTR=-153.0
+JPgrid_LAT_CTR=61.0
+JPgrid_DELX=3000.0
+JPgrid_DELY=3000.0
+JPgrid_NX=1344
+JPgrid_NY=1152
+
+# Write component grid definitions
+
+WRTCMP_nx=1340
+WRTCMP_ny=1132
+WRTCMP_lon_lwr_left=151.5
+WRTCMP_lat_lwr_left=42.360
+WRTCMP_dx=JPgrid_DELX
+WRTCMP_dy=JPgrid_DELY
+
+# Plot-specific definitions
+
+plot_res='i' # background map resolution
+
+#Note: Resolution can be 'c' (crude), 'l' (low), 'i' (intermediate), 'h' (high), or 'f' (full)
+#      To plot maps with higher resolution than low,
+#      you will need to download and install the basemap-data-hires package
+
+
+#### END User-defined variables
+
+
+
+JPgrid_width  = JPgrid_NX * JPgrid_DELX
+JPgrid_height = JPgrid_NY * JPgrid_DELY
+
+big_grid_width=np.ceil(JPgrid_width*1.25)
+big_grid_height=np.ceil(JPgrid_height*1.25)
+
+WRTCMP_width  = WRTCMP_nx * WRTCMP_dx
+WRTCMP_height = WRTCMP_ny * WRTCMP_dy
+
+fig = plt.figure()
+
+#ax1 = plt.axes
+ax1 = plt.subplot2grid((1,1), (0,0))
+
+map1 = Basemap(projection='gnom', resolution=plot_res, lon_0 = JPgrid_LON_CTR, lat_0 = JPgrid_LAT_CTR, 
+               width = big_grid_width, height=big_grid_height)
+
+map1.drawmapboundary(fill_color='#9999FF')
+map1.fillcontinents(color='#ddaa66',lake_color='#9999FF')
+map1.drawcoastlines()
+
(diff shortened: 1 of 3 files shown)


mkavulich:
If reviewers want to see my test runs, they can be found here on Hera: /scratch1/BMC/gsd-fv3-dev/kavulich/new_predef_domains/expt_dirs/

mkavulich:
After viewing the AK domain for geographical sanity, I realized the parameters are all off, and we're getting a weird US domain instead. Also, I have some questions about the contents of the output files; the contents do not seem to be intuitive; I am not sure which fields are which, and the ones that I think I know what they are do not look realistic. It's possible that I haven't configured the write component parameters correctly; I will look into it further.

mkavulich:
I have confirmed that the write component parameters were not set correctly; not sure how I missed that but I am now testing the corrected versions; should be complete by tomorrow morning

gsketefian:
Ok, thanks Mike.  Let me know when the tests complete.  One thing that would make your life easier is to also create coarse versions of these since those run a lot faster, then after everything works with those, try the 3km versions (but you don't have to if you don't think that's necessary).

mkavulich:
@gsketefian @JeffBeck-NOAA  Is there a better way to determine the write component parameters than guess-and-check methods? 

I recall in the past we used a script to read the pre-processing files and determine the best write component parameters. I'm pretty sure these parameters I'm using came from that script, but I can't remember how to re-run it to confirm (and I'm not sure it would even still work).

I've used the same resolution and taken the WRTCMP_lon_lwr_left and WRTCMP_lat_lwr_left directly from the fix_sar/C3338_oro_data.tile7.halo0.nc data that is created during the preprocessing (which I thought should work) and yet I'm still not getting the write component grid correct: it has a large white bar on the left (actually the east 😭) indicating that at the very least my longitude is incorrect in some way:

ncview ugrd

JeffBeck-NOAA:
Hey Mike,

Can you plot the landmask instead?  Gerard was maintaining the code that
would provide write component information, and it was working for
everything recently, but only for the Lambert projection (not for rotated
lat/lon or regional lat/lon), which we should get working in the future.

- Jeff

gsketefian:
Hi Mike,
It would be nice to have a short description of how to use the plotting script (but not required).  Otherwise, if this is all done (I can't quite tell), I can approve.

mkavulich:
@gsketefian I am working on a full description of the PR as we speak so there's a more permanent record of these changes and the details. Once that is posted I'll ask for a final review. 

The plotting script will remain unfinished for now as I need to track down the technical document with the exact details of the JP_grid modified gnomonic projection, and I don't want to delay this PR any further.

mkavulich:
Okay, I have finally figured out everything I think I need to figure out for this specific PR. More work needs to be done, but that's the case everywhere in the system I guess.

Dusan provided me with this tool: /scratch2/NCEPDEV/fv3-cam/Dusan.Jovic/dbrowse/fv3grid

It can be painfully slow at times over ssh but it is very useful, so long as you know what all of the parameters are and what they do. I used it to determine by eye the best write-component grid for each compute grid; here are plots of the 4 domains, compute domain in black, write domain in light blue:

## EMC_AK (Alaska)
EMC_AK_domain
ncview vgrd
ncview ugrd
ncview tmp
ncview pressfc

## EMC_HI (Hawaii)
EMC_HI_domain

## EMC_PR (Puerto Rico)
EMC_PR_domain

## EMC_GU (Guam)
EMC_GU_domain

As you can see in the example fields I supplied for EMC_AK, the "striping" pattern I mentioned earlier is nowhere to be found; this seems to happen only with severely mismatched resolutions, and certain input fields. However, there are a few fields that look weird, though this may be an artifact of the input data; for example, surface roughness:
ncview sfcr


Potential remaining issues:
 - These domains were designed just taking numbers from other fix files I found something like a year ago. They were not designed to mimic the existing HRRR domains. The write-component parameters are basically a "best fit" determined by mapping with Dusan's tool.
 - For smaller domains (HI and GU), we are assigning too many cores to EMC_post, causing a segfault presumably due to over-decomposition. I'm not sure of a good way to handle this aside from adding "post_cores" as a new configuration variable or something similar.
 - Dusan does not seem keen on supporting/providing this tool to the public, so it is probably still necessary to work on our own tools. This PR commits a script that gets us partway there, but is not yet useful for matching compute and write domains because it does not incorporate the JP_grid scaled/stretched gnomonic projection

@gsketefian @JeffBeck-NOAA @JulieSchramm let me know if you think there is anything left to discuss here, otherwise I think it's ready for merging.

JeffBeck-NOAA:
Hi Mike, I was going to merge this in, but I noticed the EMC AK compute domain (1344 x 1152) is the same number of grid points as the Lambert conformal write component output for that domain.  Normally this should be smaller than the compute domain.  Is that a mistake?

mkavulich:
@JeffBeck-NOAA It was a bit of an experiment to see if the workflow would be happy with domains if they were the same size (write component extends beyond physical domain). It worked and I never thought to change it back. If you think it's best I could trim the write domain to be entirely within the physical domain, but as I mentioned these domains are not setting out to emulate operational domains anyway. If you or others at EMC would prefer that I could work on doing that, but I could do it in a follow-up PR to get this moving, since fiddling with domains is a slow and tedious process over ssh even with Dusan's tool.

JeffBeck-NOAA:
@mkavulich No, I think it's completely fine to leave as is.  There will be a few lines of NaN in the output, but that's not an issue at all.  It's more important to get this in and focus on other things for the release at the moment.  As you said, we can always come back to this in the future.

Going to merge!

PR #155, opened 2020-02-19
Feature/users guide rst More cleanup of old docs Users Guide

JulieSchramm:
- Delete year-old Acknowledgement.rst, Chapter1.rst, Chapter3.rst, Chapter4.rst
- Rename Chapter2.rst to Quickstart.rst
- Add FAQ.rst and SystemRequirements.rst
- Add updates to Introduction.rst
- Update project name and year in conf.py

html and pdf build to completion


jwolff-ncar:
@JulieSchramm @gsketefian To follow the MRWeather App, I think the end-to-end system will be called something like the UFS Short-Range Weather Application. I think the "application" would include workflow generation, build steps, etc. That is what our UG will cover. Does that makes sense @llpcarson? So what we currently have as regional_workflow would be part of the full application.

PR #158, opened 2020-02-28
Update module load for hera to be consistent with what is in post repo

jwolff-ncar:
This update allows the latest hash for post to run successfully on hera using the regional_workflow.

### DIFF
--- a/modulefiles/tasks/hera/run_post
+++ b/modulefiles/tasks/hera/run_post
@@ -21,10 +21,10 @@ module load g2/3.1.1
 module load gfsio/1.1.0
 module load ip/3.0.2
 module load sp/2.0.3
-module load w3emc/2.3.1
 module load w3nco/2.0.7
 module load crtm/2.2.5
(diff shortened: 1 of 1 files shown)


JeffBeck-NOAA:
Gerard and I talked about the potential to point the run shell script to the EMC_post module file, thus avoiding this problem in the future altogether, and allowing us to have a single module file for the build and execution.

jwolff-ncar:
Did you want to do that instead of this PR? That is fine with me - do @JeffBeck-NOAA  or @gsketefian want to do that? If we want this to be a future enhancement we can create a new issue for that and approve/merge this one for now so that the repo actually runs.

JeffBeck-NOAA:
Let's get this PR accepted and then make a new enhancement issue to take
care of things separately.  Sound good?

gsketefian:
Yes that sounds good.  I'll let Mike or Julie approve so we have one GSD and one NCAR approver.

Issue #160, opened 2020-03-02
Make post modules files consistent


jwolff-ncar:
Gerard and I talked about the potential to point the run shell script to the EMC_post module file, thus avoiding this problem in the future altogether, and allowing us to have a single module file for the build and execution.

PR #163, opened 2020-03-09
Eliminate duplication of module load for EMC_post

JulieSchramm:
- Remove run_post file that duplicates module loads from EMC_post
- Modify scrpits to create link to already existing module load in EMC_post
- Addresses issue #160

Workflow runs to completion, these mods are non-answer changing.


mkavulich:
Julie, the way that libraries work or do not work with the EMC build of post at runtime is a bit of a mystery to me, so as long as it runs fine I'm okay with it.

PR #168, opened 2020-03-11
Adding Python utility to manage namelists.

christinaholtNOAA:
These changes provide a standalone Python utility (set_namelist.py) that takes some combination of an input Fortran namelist, a YAML configuration file, and a dict-formatted string to write out a fully functional Fortran namelist. The `-h` functionality is supported for a full description of command line argument parameters.

The following tasks have been completed:

- The existing Fortran namelists for FV3 have been reduced to a single "baseline" configuration in `input.nml.FV3` and all differences have been captured in a YAML configuration file, organized into sections corresponding to the naming convention of the namelist files. 
- The make_grid namelist template was removed.
- The set_namelist.py script has been incorporated into the generate_FV3SAR_wflow.sh and exregional_make_grid.sh scripts, replacing calls to set_file_param for namelists.
- The modulefile for make_grid was updated to load the conda module and activate the necessary environment. The README was updated to reflect the same requirement for running generate_FV3SAR_wflow.sh.

Open questions:
- Where should the information that currently exists in templates/regional_grid.nml live? This file is no longer needed (but hasn't been explicitly removed here)
- Where does the appropriate information about required modules go for running the test suite?
- Should I run a baseline of the test suite and compare bit-for-bit results?

Ongoing:
- I am currently in the process of running the test suite on my branch after pulling in top of community_develop. Many of the tests already passed on the older version, but I updated before continuing to all of them.


environment.yml
```diff
@@ -0,0 +1,6 @@
+name: regional_workflow
+channels:
+  - conda-forge
+dependencies:
+  python: 3.7
+  f90nml: 1.1.2
```
christinaholtNOAA, line 6: I added this file to make the transition to other platforms easier. It should be feasible to create the required conda environment using:

`conda env create -f environment.yml`

ush/templates/FV3.input.yml
```diff
@@ -0,0 +1,393 @@
...
+    fv_debug: False
+    range_warn: False
+    k_split: 6
+    n_split: 6
+    dnats: 1
+    d4_bg: 0.15
+    vtdm4: 0.075
+    delt_max: 0.008
+    do_sat_adj: True
+    nord_zs_filter: !!python/none
```
christinaholtNOAA, line 53: This is a YAML built-in that allows us to specify a Python "None" type, and choose to delete this item from the resulting namelist.

ush/templates/FV3.input.yml
```diff
@@ -0,0 +1,393 @@
...
+    icloud_bl: !!python/none
+    bl_mynn_tkeadvect: !!python/none
+    bl_mynn_edmf: !!python/none
+    bl_mynn_edmf_mom: !!python/none
+    do_deep: False
+    lgfdlmprad: True
+    effr_in: True
+  nam_stochy: !!python/none
+  nam_sfcperts: !!python/none
+  gfdl_cloud_microphysics_nml: &gfs_gfdl_cloud_mp
```
christinaholtNOAA, line 94: The YAML anchor, specified with `&` here lets us reference the section later.

ush/templates/FV3.input.yml
```diff
@@ -0,0 +1,393 @@
...
+    xkzminv: 0.3
+    xkzm_m: 1.0
+    xkzm_h: 1.0
+  namsfc:
+    ldebug: False
+    landice: True
+  nam_stochy: !!python/none
+  nam_sfcperts: !!python/none
+  gfdl_cloud_microphysics_nml:
+    <<: *gfs_gfdl_cloud_mp
```
christinaholtNOAA: This notation means that we are filling in the entire section by pointing to the referenced by the anchor, any additional config items included here will overwrite the referenced section.

ush/templates/input.nml.FV3
```diff
@@ -8,11 +8,9 @@
 /
 
 &atmos_model_nml
-     blocksize = <blocksize>
```
christinaholtNOAA, line 11: Removing all template entries. These are handled in the script by providing the command-line option `-u`

README.md
```diff
@@ -15,11 +15,16 @@ This step will checkout EMC\_post, NEMSfv3gfs and its submodules, UFS\_UTILS\_ch
...
 ./build_all.sh
 ```
+NOTE: You must not have the conda module loaded for the build to succeed.
+
 This step will also copy the executables to the `exec` directory and link the fix files.
 4. Create a `config.sh` file in the `ush` directory (see Users Guide).
 5. Generate a workflow:
 ```
 cd ush
+module use /scratch2/BMC/zrtrr/cholt/modulefiles
```
JulieSchramm: This will be a problem with portability and stability.  It is generally advised not to put code used by the community in a users directory.  With the location on scratch, which could possibly be cleaned or scrubbed, this is a problem.  This module would have to be made available on every pre-configured platform, or installed by the user.

modulefiles/tasks/hera/make_grid
```diff
@@ -8,3 +8,6 @@ module load intel/18.0.5.274
 module load netcdf/4.7.0
 module load hdf5/1.10.5
 
+module use /scratch2/BMC/zrtrr/cholt/modulefiles
```
JulieSchramm: Same as comment above.

README.md
christinaholtNOAA: I totally agree! I can absolutely work toward getting this module moved to contrib on Hera. I can also provide the modulefile to help with portability to other supported machines. Would it be acceptable for the short-term to keep it in this location until its eventual home is determined, or does that need to happen immediately?

JulieSchramm: I don't see a conda module on Cheyenne.  How much work would it be to install it there?

christinaholtNOAA: It's hard to say. I know that NOAA is restricting access to the outside world for installing conda. If these restrictions (in a different NCAR flavor) exist on Cheyenne, then it may not be trivial. If it is less restrictive, and they support tcl modulefiles, it might be relatively similar to the Hera implementation.

I am not currently seeing a set of modulefiles for Cheyenne. Is the plan to add Cheyenne soon?

JulieSchramm: I finally got a build working on Cheyenne, and conda would need to be installed somewhere for this to work.  I also requested installation of a rocoto module on Cheyenne about a month ago, but I haven't seen any progress yet.  Cheyenne would be the platorm most accessible to the community.

llpcarson: @JulieSchramm cheyenne has a more robust python environment, so if (I think?) conda is only being used to install the f90nml package, that entire step may not be needed.  I may be overlooking something, though!

Yes - plan to add cheyenne "soon", and wider portability to community users is also needed (including those who may not have "modules" available at all)

christinaholtNOAA: Conda is certainly not a requirement here. Conda is one way to manage the installation of f90nml. Users only need some way to install f90nml, and then access it when running the script.

llpcarson: would it make sense for it to be a user-installed pre-requisite?  Seems like each system might have a different solution, although perhaps something could be automated - I'm not a python package expert!

JulieSchramm:
I am building Christina's fork now on Cheyenne, I'll see what I can do
tomorrow.

README.md
christinaholtNOAA: @llpcarson I think it makes sense to make it a user-requirement, although I think it would be nice to have common-use installations on our main platforms so that we don't have to walk every Hera/Cheyenne through that process.

Providing the `environment.yml` file makes this pretty straightforward for having a user install a conda environment we suggest on their own platform (they'd be responsible for installing conda). We could similarly provide a pip `requirements.txt` for users who prefer to make a pip virtual environment, and maybe that will be what we need for the Cheyenne solution, anyway. Outside those two options, I'd say folks would be on their own for installing the packages.

JulieSchramm:
The build on Cheyenne still works.  I can't get the conda environment to
work:

module load ncarenv
module load python
ncar_pylib
pip install conda

Not sure what to do now.

christinaholtNOAA:
@JulieSchramm I believe you should be able to skip the conda step on Yellowstone. You'd only really need Python 3.6+ with f90nml. While I don't know exactly how the ncar_lib command works, it may be something as easy as:
```
module load python/3.6.8
ncar_pylib f90nml
```
I was following along with the docs here: https://www2.cisl.ucar.edu/resources/computational-systems/cheyenne/software/python-%E2%80%93-ncar-package-library

It seems that if f90nml is not already available, you can install it with pip. Conda is not necessary if Cheyenne already provides a Python package/environment management tool like ncar_pylib.

christinaholtNOAA:
As conversations have progressed with @JacobCarley-NOAA , it seems that he is in favor of moving forward with the set of proposed utilities since they could prove useful now, and an NCO implementation is not scheduled until 2023. He also suggests that since the workflow has the potential to evolve significantly over the next several years, the risk of incorporating functionality that is not fully NCO compliant _now_ (i.e., the use of a small set of Python libraries) is low. He seems confident that we can work out the details of using these libraries in operations, if it is still relevant at that time.

We have also discussed adding a couple of EMC folks to the PR to review what I'm adding earlier, rather than later, if you all are okay with that. I don't have the permission to add reviewers, so if someone on the team could add @RatkoVasic-NOAA and @BenjaminBlake-NOAA, that would be great!

Let me know your thoughts!

JacobCarley-NOAA:
Added @RatkoVasic-NOAA

gsketefian:
I have one more pr to submit that NCAR is waiting on and then will get to this, probably tomorrow.

ywangwof:
I found a few problems/suggestions.

1.  In file "environment.yml" a dependency for "pyyaml" should be added. Otherwise, you will get an error, "ModuleNotFoundError".
2. If users only care for one specific run, there should be a way to avoid providing all configurations as in file _ush/templates/FV3.input.yml_. If users provide one configuration in the yams file,  option "-c" in command line should accept one argument. I prefer to using multiple documents in one Yaml file to separate those configurations and give each configuration a specific name, or individual yaml files for these configuration templates,
```
ush/templates/input.nml.FV3_CPT_v0
ush/templates/input.nml.FV3_GFS_v16beta
ush/templates/input.nml.FV3_GSD_SAR
ush/templates/input.nml.FV3_GSD_v0

```

Furthermore, two arguments with the same name is also confusing.

> usage: set_namelist.py [-h] [-c **CONFIG CONFIG**] [-n NML] -o OUTNML [-q] [-u USER_CONFIG]

3. The instructions in the script header do not mention configuration (section) name. I made a mistake at the first try by providing a yaml file contains only one set of configuration. Furthermore, it may be a typo, _fv_core_0nml_ should be **fv_core_nml**.

Overall, the python script is well written and documented. I have working with _namelist/yaml_ file a lot using python, but I do not know the existence of module **f90nml** and have to write my own namelist processing functions. This script "set_namelist.py" is much simple and clear.

christinaholtNOAA:
@ywangwof Thanks for the review! Great catch on #1 and I updated some of the doc strings to hopefully help ease the confusion you were feeling for the next folks who are using the script.

```
$ python set_namelist.py
usage: set_namelist.py [-h] -o OUTNML [-c [FILE, SECTION]] [-n NML] [-q] [-u YAML STRING]
```
or
```
$ python set_namelist.py -h
usage: set_namelist.py [-h] -o OUTNML [-c [FILE, SECTION]] [-n NML] [-q] [-u YAML STRING]

Update a Fortran namelist with user-defined settings.

optional arguments:
  -h, --help            show this help message and exit
  -o OUTNML, --outnml OUTNML
                        Required: Full path to output Fortran namelist
  -c [FILE, SECTION], --config [FILE, SECTION]
                        Full path to a YAML config file containing multiple configurations, and the top-level section to
                        use. Optional.
  -n NML, --basenml NML
                        Full path the input Fortran namelist. Optional.
  -q, --quiet           If provided, suppress all output.
  -u YAML STRING, --user_config YAML STRING
                        Command-line user config options in YAML-formatted string. These options will override any
                        provided in an input file. Optional.
```

I would like to address some of the benefits of using a single YAML file to manage all the separate namelist options. I will say that it is my intention to fully incorporate all the available options in the repo at the time that my PR is cleared, but want to hold off until the PR is approved so that there is only one merge of that kind. So, here are the advantages I see in setting all in a single file:

- Duplication of information is reduced by using native pyyaml features like anchors/aliases (`&` and `*` syntax), as well as key merging (`<<` syntax), to refer to previous settings in the same file. It is a bit more difficult and convoluted to gain the same functionality when splitting into multiple files. 
- Duplication is also reduced by maintaining only differences from a "base configuration" in the `input.nml.FV3`. That file contains all the settings needed for FV3, and it is the canonical source for the settings that are common to all supported FV3 configurations. For example, if a required FV3 namelist option gets added, but it is set to the same thing across all the SAR applications, it only needs to be added to this single file.
- Maintenance of configurations becomes more straightforward when like settings are grouped in a single, readily-accessed location.

I'm happy to hear more feedback on the topic, especially if there are considerations I am missing about the advantages of breaking these configurations out into separate files.

scripts/exregional_make_grid.sh
```diff
@@ -361,8 +361,18 @@ Setting parameters in file:
...
 
-  ${USHDIR}/set_namelist.py -q -o ${rgnl_grid_nml_fp} -u "{$settings}" ||
-     (echo "set_namlist.py failed!" && exit 1)
+  ${USHDIR}/set_namelist.py -q -o ${rgnl_grid_nml_fp} -u "{$settings}"
+  if [[ $? -ne 0 ]]; then
+    echo "
+    !!!!!!!!!!!!!!!!!!!!!!
+  
+    set_namelist.py failed!
+    Check documentation to ensure that the proper python environment is available
```
christinaholtNOAA: I'm not so sure this comment is helpful in the ex-scripts. To my knowledge, there is no documentation about this, and it should be handled by the workflow task loading the module files.

ush/load_modules_run_task.sh
```diff
@@ -292,7 +292,7 @@ Call to \"module use\" command failed."
    #
    case "${task_name}" in
    #
-   "${MAKE_ICS_TN}" | "${MAKE_LBCS_TN}" | "${MAKE_SFC_CLIMO_TN}")
+   "${MAKE_GRID_TN}" | "${MAKE_ICS_TN}" | "${MAKE_LBCS_TN}" | "${MAKE_SFC_CLIMO_TN}")
```
christinaholtNOAA: This won't work for the changes I made to make_grid. @mkavulich Can you please retain this file as a true modulefile?

christinaholtNOAA:
@mkavulich Thanks for adding the Cheyenne capability. Your changes look good to me. Happy to get this one merged. :)

mkavulich:
Documentation here has been updated: https://docs.google.com/document/d/1V5F3bJNDBoe5QnpmhNlYC-Mo69mpG-vptxn4wYOtBpA/edit?usp=sharing

The top-level README.md file has also been updated with the latest compile/run instructions.

PR #169, opened 2020-03-13
Fix to build_post.sh for EMC_post compile instead of DTC_post for Her…

JeffBeck-NOAA:
…a, Jet, WCOSS, Cray, and Dell


JeffBeck-NOAA:
Just an FYI, the configure/compile files don't exist in EMC_post, which is
why we use build_ncep_post.sh.

JulieSchramm:
We are now using the DTC_post branch of the EMC_post repository, instead of the develop branch.  The DTC_post branch can be built on more platforms, including Cheyenne.  If you look in the EMC_post directory of your fork, you will see the ./compile and ./configure scripts.

JeffBeck-NOAA:
OK, I didn't realize you'd changed the default branch for all systems.  I
was under the assumption that this change only applied for Cheyenne.  We
need to discuss this in more detail during our next meeting, since I don't
know whether EMC is willing to use this branch instead of develop.  In
addition, the develop branch is where major SAR commits from EMC have been
pushed, including bug fixes.  How often is the DTC_post branch updated?

JulieSchramm:
The DTC_post branch is in the same repository as develop:
https://github.com/NOAA-EMC/EMC_post/tree/DTC_post
 I'm not sure how often the develop branch is merged in.  @mkavulich might know more than I do about this repo.

mkavulich:
To try to answer all questions in one reply (let me know if I missed anything):

There is no checking out different versions of code for different machines. That's all handled by manage_externals, which is *before* the build step. Even if it were possible, it's something we should not do. We updated the EMC_post hash to point to the DTC_post branch, which is still in the same repository and has all of the same scripts for building on "official" platforms, but has the additional ./configure and ./compile scripts for the "community" version of the build.

The source code of the DTC_post branch should be identical to the post release branch. The only difference should be in the build system.

@JeffBeck-NOAA Can you point me to your failing build on Hera? I have a hunch that the problem is you are not fully updating all your externals and so you are not getting the correct hash in post, but it's possible something else is going on.

mkavulich:
Also, as a side note, on the "official" platforms, the ./compile script simply calls the build_ncep_post.sh script, so there is no functional difference in the build. And however we set it up now for official platforms probably doesn't matter, as the build system for that will likely change in the coming months.

JeffBeck-NOAA:
The reason my build fails is because I have been building top-of-the-trunk
code and don't use hashes, so my manage_externals is set to check out the
develop branch of EMC_post, which doesn't contain the configure and compile
scripts.

We still need to talk with EMC to see if they will go along with changing
branches to DTC_post.  If the develop branch is merged into DTC_post
regularly (as in weekly updates), then they may OK with the change, since
there are frequent SAR updates to develop.

- Jeff

JulieSchramm:
@JeffBeck-NOAA If you have a set of top of the trunk repos that work well with the workflow, would it be possible to commit their corresponding hashes to the community_develop branch Externals.cfg file, keeping the EMC_post at the 172bab0 hash.  That way, we can all use the same code.  Does anyone see any problem with this?

Issue #134, opened 2019-12-18
Port to Jet


jwolff-ncar:

JeffBeck-NOAA:
Working on this now...

climbfuji:
Hey Jeff, for your info, support for jet within the ufs-weather-model itself (i.e. regular regression testing) will be added back in the current PR (https://github.com/ufs-community/ufs-weather-model/pull/18). The code should compile and run even without the PR.

JeffBeck-NOAA:
Thanks, Dom!  I confirmed that it does compile even without the PR, but
it's good to know there will be regular regression testing there too.

- Jeff

jwolff-ncar:
Should this issue now be closed?

JeffBeck-NOAA:
Yes, I was going to mark that we should resolve it in our task notes, but
please feel free to close it!

PR #173, opened 2020-03-22
Port to Jet

JeffBeck-NOAA:
This PR covers three topics:

1) Port to Jet: https://github.com/NOAA-EMC/regional_workflow/issues/134
- End-to-end tests succeeded on both Hera and Jet aside from the following:
     - Coarse Alaska grid (Existing problem with MPI grid decomposition for coarse resolutions)
     - NCO test (location of fix_sar directory unknown or may not exist yet on Jet)
     - Rotated_lat/lon grid with GFDL mp (regional test 007) fails in post task (a result of using DTC_post instead of EMC_post; the former needs to be updated to the newest code in EMC_post/develop). Regional 007 runs to completion when using EMC_post/develop.
     - DTC_post must be built using the v8.0.0-jet modulefile from EMC_post/develop.  DTC_post needs to be updated to use this modulefile. 
2) Update Externals.cfg to top-of-the-trunk repos aside from EMC_post (uses DTC_post hash)
3) Modify all diag_table templates to avoid writing out fv3_history.nc file


JeffBeck-NOAA:
I updated the DTC_post hash to the one in Mike's PR, so this is ready to merge.

PR #183, opened 2020-03-30
Adding a Jinja Templating tool for Rocoto.

christinaholtNOAA:
This PR make the following modifications:

- Add the Python script `create_xml.py`
- Change `templates/FV3SAR_wflow.xml` to a Jinja template. 
- Update `generate_FV3SAR_wflow.sh` to call the Python script.
- Add jinja2 to `environment.yml`

On Hera, the miniconda3 contrib module has been updated to include Jinja support. Directions for using that module are in the README of PR #168. As a reminder, you can access that module/environment on Hera using:

```
module load contrib miniconda3
conda activate regional_workflow
```

I have not run the regressions tests just yet given that this set of updates is still in a "work in progress" state. I have tested that a usable XML file is generated as expected, and am happy to run the regression tests once we are ready to merge the PR.

Could someone please add @RatkoVasic-NOAA as a reviewer here, too? Thanks!


gsketefian:
Hi Christina,
I have a couple of PRs that I need to get in pretty urgently and will then look at this and your other namelist PR carefully.  I want to understand how things work so that I can use this approach with any new changes we want to make to the workflow.  I will probably get to it next week.  Thanks!
G

christinaholtNOAA:
After yesterday's merge, I resolved conflicts with this branch and tested on Hera. Test suite succeeded. I'm ready for a merge when you all are.

ush/generate_FV3SAR_wflow.sh
```diff
@@ -144,115 +126,48 @@ fi
...
-#
-# Perform the subsutitution.
-#
-sed -i -r -e "s|${regex_search}|${all_cycledefs}|g" "${WFLOW_XML_FP}"
 
+settings="
+  'account': $ACCOUNT
+  'sched': $SCHED
+  'queue_default': $QUEUE_DEFAULT
+  'queue_default_tag': $QUEUE_DEFAULT_TAG
```
gsketefian, line 134: If one wanted to, is there a way to include a comment within this settings block, e.g. something like

```
...
'queue_default': $QUEUE_DEFAULT
# Some comments here.
'queue_default_tag': $QUEUE_DEFAULT_TAG
...
```

gsketefian:
Hi Christina,
This looks good to me.  If all the WE2E tests passed, we can go ahead and merge (some of them like the coarse Alaska domain one probably still won't pass with this version).

ush/generate_FV3SAR_wflow.sh
christinaholtNOAA: @gsketefian I just ran a test and looks like it works with syntax like this:

```
    'npx': ${npx},
    # A comment
    'npy': ${npy},
```

I assume that's what you had before the markdown took control.

gsketefian: Ok, thanks Christina.  Yes, it was something like that before the change.  I will go ahead and merge your PR.

PR #187, opened 2020-04-01
Community develop

JeffBeck-NOAA:
Modify Externals.cfg for top-of-the-trunk hashes.  Tested on both Jet and Hera.


JeffBeck-NOAA:
Thanks, Julie!

PR #188, opened 2020-04-07
Running on Cheyenne 

mkavulich:
This PR introduces the capability to run on Cheyenne with Intel compilers. Several minor changes needed to be made to the build process to allow this to happen.

# Summary of changes

 - J jobs and exscripts were updated with stanzas for Cheyenne
 - Some Cheyenne modulefiles were created
 - The workflow generation script and XML template needed to be updated in several places to allow new variables, mostly related to Cheyenne using a PBSpro queueing system rather than slurm:
   - NPROCS_RUN_FCST (total number of processors for running the forecast step) needed to be passed for the mpirun -np command used for running with PBSpro
   - NCORES_PER_NODE is now specified, as Rocoto complains when <nodesize> is not specified for PBSpro jobs
   - Three new variables, QUEUE_DEFAULT_TAG, QUEUE_HPSS_TAG, and QUEUE_FCST_TAG were added: for slurm, Rocoto uses the \<partition> tag rather than the \<queue> tag for certain job submissions, but this tag is not supported for PBSpro. 
   - UFS_WTHR_MDL_DIR: allows us to find the location of the fv3 ccpp library files and add that to LD_LIBRARY_PATH at model runtime; otherwise the executable fails due to not finding these .so files
 - Modifications were made to the workflow generation script and module loading script to allow EMC_post to run on Cheyenne with no modulefile, as it is not necessary
 - Set the new location for Cheyenne fix files: /glade/p/ral/jntp/UFS_CAM/fix/

I have copied over the global input files for running a forecast starting at 2019090118; these can be found in /glade/p/ral/jntp/UFS_CAM/COMGFS/gfs.20190901/18/. If you want to run a test case for a different initial date/time you will have to bring over the files yourself; there are some other sample datasets on Hera here: /scratch1/NCEPDEV/hwrf/noscrub/hafs-input/COMGFS. 

# Tests run:

 - Full workflow on Cheyenne (compilation and running through post). Used GSD_HRRR25km pre-defined domain. Results here: /glade/work/kavulich/UFS/get_running_on_cheyenne/expt_dirs/test_community
 - Full workflow on Hera (compilation and running through post). Used GSD_HRRR25km pre-defined domain. Results here: /scratch2/BMC/det/kavulich/cheyenne_branch_testing/expt_dirs/test_cheyenne_mods_25km

# Potential problems that need further investigation

 - The make_lbcs and make_ics steps are intermittently and randomly segfaulting in the final stages of the executable, on Cheyenne only. They are succeeding after re-running every time, but this should probably be investigated.
   - See issue #190 
 - ~~The make_sfc_climo task is taking more than 10 times longer on Cheyenne; I'm not sure what the reason is but it should probably be investigated.~~
   - Resolved, I had not properly set up this task to use MPI, now fixed
 - I originally tried to test the GSD_RAP13km pre-defined domain but it turns out that the model segfaults, even in the current community_develop version. ~~Perhaps not enough processors?~~
  - See issue #191  

# Other notes
I also took the liberty of removing stanzas for the old, defunct Theia machine in this PR as well.


llpcarson:
Nice work, Mike.  This looks good to me.  

Your notes reminded me that the "dynamic build" of the ufs-weather-model is going to be deprecated/retired sometime in the next couple of months, and that will impact the regional build and workflow (no need for LD_LIBRARY_PATH updates, for example, but SDFs specified at build time).  I will open an issue in ufs-srweather-app so we don't forget this one!

modulefiles/tasks/cheyenne/make_orog.hardcoded
```diff
@@ -0,0 +1,11 @@
+#%Module#####################################################
+## Module file for make_grid task.
```
gsketefian: In the comment, change make_grid to make_orog.  Another option is to get rid of the comment lines 2 and 3 since it's already obvious from the file name what task this module file is for (and we would do that for all module files).  I had the comment in there originally because I copied it over from a sample EMC module file, and they had it.

scripts/exregional_make_grid.sh
```diff
@@ -171,6 +171,11 @@ case $MACHINE in
   ulimit -a
   ;;
 
+"CHEYENNE")
+
+  export APRUN="time"
+  export topo_dir="/glade/p/ral/jntp/UFS_CAM/fix/fix_orog"
```
gsketefian, line 163: Do these need to be exported?

scripts/exregional_make_orog.sh
```diff
@@ -202,6 +180,11 @@ case $MACHINE in
   ;;
 
 
+"CHEYENNE")
+  export APRUN="time"
+  export topo_dir="/glade/p/ral/jntp/UFS_CAM/fix/fix_orog"
```
gsketefian, line 146: Don't export?

scripts/exregional_make_sfc_climo.sh
```diff
@@ -153,6 +146,10 @@ case $MACHINE in
   APRUN="srun"
   ;;
 
+"CHEYENNE")
+  export APRUN="time"
```
gsketefian: Don't export?

ush/config_defaults.sh
```diff
@@ -54,26 +54,41 @@ RUN_ENVIR="nco"
...
+# platforms this should be "queue"
+#
 # mach_doc_end
 #
 #-----------------------------------------------------------------------
 #
 MACHINE="BIG_COMPUTER"
 ACCOUNT="project_name"
 QUEUE_DEFAULT="batch_queue"
+QUEUE_DEFAULT_TAG="queue"
```
gsketefian, line 87: If these three "_TAG" variables are only machine dependent and should just be set automatically depending on the scheduler and/or machine, then we can move from here to setup.sh (which is the script that sets secondary variables, i.e. the ones derived from the user-specified or primary variables).  That way they'll be hidden from the user, just like for example NCORES_PER_NODE.

ush/generate_FV3SAR_wflow.sh
```diff
@@ -147,10 +148,13 @@ fi
...
-set_file_param "${WFLOW_XML_FP}" "QUEUE_DEFAULT" "${QUEUE_DEFAULT}"
-set_file_param "${WFLOW_XML_FP}" "QUEUE_HPSS" "${QUEUE_HPSS}"
-set_file_param "${WFLOW_XML_FP}" "QUEUE_FCST" "${QUEUE_FCST}"
+set_file_param "${WFLOW_XML_FP}" "QUEUE_DEFAULT" "<${QUEUE_DEFAULT_TAG}>${QUEUE_DEFAULT}</${QUEUE_DEFAULT_TAG}>"
+set_file_param "${WFLOW_XML_FP}" "QUEUE_HPSS" "<${QUEUE_HPSS_TAG}>${QUEUE_HPSS}</${QUEUE_HPSS_TAG}>"
+set_file_param "${WFLOW_XML_FP}" "QUEUE_FCST" "<${QUEUE_FCST_TAG}>${QUEUE_FCST}</${QUEUE_FCST_TAG}>"
+set_file_param "${WFLOW_XML_FP}" "NCORES_PER_NODE" "${NCORES_PER_NODE}"
 set_file_param "${WFLOW_XML_FP}" "PROC_RUN_FCST" "${PROC_RUN_FCST}"
+set_file_param "${WFLOW_XML_FP}" "NPROCS_RUN_FCST" "${NPROCS_RUN_FCST}"
+set_file_param "${WFLOW_XML_FP}" "UFS_WTHR_MDL_DIR" "${UFS_WTHR_MDL_DIR}"
```
gsketefian: The setting of UFS_WTRH_MDL_DIR can be moved down to the next section with the other directories.

jobs/JREGIONAL_MAKE_ICS
```diff
@@ -134,7 +134,7 @@ case "$MACHINE" in
   ;;
 #
 "CHEYENNE")
-#
+  APRUN="mpirun -np ${NPROCS}"
```
gsketefian, line 137: Is NPROCS set in the workflow scripts somewhere or is it an environment variable set by pbspro?

ush/templates/FV3SAR_wflow.xml
```diff
@@ -351,14 +373,16 @@ MODULES_RUN_TASK_FP script.
...
     <jobname>&RUN_FCST_TN;</jobname>
     <join><cyclestr>&LOGDIR;/&RUN_FCST_TN;_@Y@m@d@H.log</cyclestr></join>
 
     <envar><name>GLOBAL_VAR_DEFNS_FP</name><value>&GLOBAL_VAR_DEFNS_FP;</value></envar>
     <envar><name>CYCLE_DIR</name><value><cyclestr>&CYCLE_DIR;</cyclestr></value></envar>
     <envar><name>CDATE</name><value><cyclestr>@Y@m@d@H</cyclestr></value></envar>
     <envar><name>PDY</name><value><cyclestr>@Y@m@d</cyclestr></value></envar>
-
+    <envar><name>NPROCS</name><value>&NPROCS_RUN_FCST;</value></envar>
+    <envar><name>LD_LIBRARY_PATH</name><value>&UFS_WTHR_MDL_DIR;/FV3/ccpp/lib/</value></envar>
```
gsketefian: For hera and jet, this is done in the scripts/exregional_run_fcst.sh script.  You can probably do the same for cheyenne.  That way you don't have to define UFS_WTHR_MDL_DIR in the XML and won't need this line (you'll have to add a line in that script though).

scripts/exregional_make_grid.sh
mkavulich: Most of the other stanzas have exported variables; I just followed the template of other machines. APRUN probably doesn't need exporting, but topo_dir probably does. Since that variable is only used in the executable, not the exscript, I'm surprised it works without exporting, as non-exported variables should only be available within the current process.

jobs/JREGIONAL_MAKE_ICS
mkavulich: The number of processors is available via an environment variable in slurm, but not in pbspro (as far as I can tell), so on Cheyenne we have to specify it ourselves in the xml.

ush/templates/FV3SAR_wflow.xml
mkavulich: Good call, I could not figure out where this was being set for Hera for some reason; I swear I looked for LD_LIBRARY_PATH but somehow I didn't find it. I'll change this to work the same way.

mkavulich:
I got individual emails for each comment, which is fine. For future reference there's the capability to "start a review" which will bundle all the comments into a single post. But it's no big deal. I'll address each comment as I get to them.

jobs/JREGIONAL_MAKE_ICS
gsketefian: Oh I see it now.  NPROCS gets set to the appropriate NPROCS_${TASK_NAME} variable in the xml.  Thanks.

scripts/exregional_make_grid.sh
gsketefian: I forgot that in my latest PR, I don't export for hera because everything becomes a local variable (I end up calling the executable directly instead of yet another script (fv3gfs_make_grid.sh), and that script is the one that requires topo_dir).  But that change isn't in the version you started out with.  Sorry for the confusion.

Also, I'd trust only the hera and jet stanzas since the other ones haven't been touched in months, maybe years!

ush/templates/FV3SAR_wflow.xml
gsketefian: Ok, thanks.

jwolff-ncar:
@mkavulich Please make new GitHub issues to keep track of the few things you mention under potential problems that need further investigation. Thanks!

ush/config_defaults.sh
mkavulich: I'm on the fence about this; theoretically people *could* specify queues rather than partitions for slurm machines as well. I'm not too much of a slurm expert though, so I'm not sure about how practical that would be, or if it's worth worrying about.

mkavulich:
@gsketefian I think I have resolved all your concerns except for one about queues which I replied to. Let me know what you think about that reply, otherwise let me know if I can merge the PR.

ush/config_defaults.sh
gsketefian: How do you specify a queue in slurm?

As I looked at this again, it seemed to me that a new user will be confused by the naming, thinking that QUEUE_DEFAULT_TAG (for example) is yet another category of queue.  Maybe we can rename to something like XML_TAG_QUEUE_DEFAULT, etc.  

So when you're running on cheyenne, you currently (re)set QUEUE_HPSS_TAG, for example, to "queue" (which is what you need for pbspro, right?) in your config.sh file?  If so, that's a lot to expect of a new user to understand; I think it would be better to hide it from the user and set it behind the scenes in setup.sh.

Feel free to take/reject my suggestions and then merge.  My PR will probably have some conflicts with yours when I try to merge yours into mine.  Please take a look once I submit it tomorrow to make sure you're ok with any changes.  Thanks.
G

mkavulich:
After resolving merge conflicts I'm now seeing segmentation faults in the run_fcst step; I need to resolve this before the PR gets merged

gsketefian:
Hey Mike,
FYI, I'm ready to submit my PR, but I'm going to wait until yours is merged so I can merge your changes into mine.
Gerard

PR #189, opened 2020-04-07
Feature/preproc

gsketefian:
* Fix bug where the orography filtering was failing but not in an obvious way.
* Simplify and clarify (as much as possible!) the preprocessing tasks.  
* No longer use mosaic_grid.fd code to make mosaic files for a JPgrid type of grid.  Instead, use GFDL's make_solo_mosaic code (which is the one used for a GFDLgrid type of grid).
* This PR needs to be accompanied by two PRs to the NCAR/UFS_UTILS repo.  I have those submitted but apparently will have to resubmit.  I'll link them here once I resubmit.

The major changes are as follows:
------------------------------------
1) Fix bug where the orography filtering was failing but not in an obvious way. 

2) To simplify the scripting and the need to interact with scripts in external repositories such as UFS_UTILS, do not use the scripts fv3gfs_make_grid.sh, fv3gfs_make_orog.sh, and fv3gfs_filter_topo.sh. in UFS_UTILS.  Instead, call the executables that these scripts would call directly in regional_workflow's ex-scripts.  Also, delete the local versions of these scripts (since they are available in UFS_UTILS).

3) Remove variables that were defined and exported only because they were needed by the above three scripts.

4) Change the grid and orography file names so that they contain the new user-specified workflow variable DOT_OR_USCORE after $CRES instead of a hard-coded underscore.  This variable should be set to either an underscore ("_") or a period (".").  If set to an underscore, the file names will not change.  If set to a period, the grid and orography file names will be updated to the naming convention used for the surface climatology files.  The goal is to eventually use a period so that all fixed files use the same naming convention (i.e. with a period following $CRES).

5) Modify process_args() so that it can process an empty array of arguments.  Then remove unused/dummy arguments in other functions that use process_args() that are never actually used (but were present because process_args() had to have at least one argument).

6) Don't use the mosaic_file.fd code to create mosaic files in the case of a JPgrid.  Instead, use GFDL's make_solo_mosaic code.  Previously, did not realize this could be used for both JPgrid and GFDLgrid types of grids.  

7) For clarity, now create three mosaic files (one for the wide-halo grid, one for the 3-cell-wide halo grid, and one for the 4-cell-wide halo grid) instead of just one for the wide-halo grid.

Tests performed:
------------------
This PR passed 12 of 14 workflow end-to-end tests.  The ones that failed are regional_005 and regional_013.  regional_005 is on the very coarse (50 km) Alaska domain; this test fails due to MPI nodes not being selected correctly for chgres_cube.  This is a problem with chgres_cube and must be addressed in another PR.  regional_013 does not work because the name of the physics suite that it uses changed the the ufs_weather_model repo.  This must be addressed in another PR.


scripts/exregional_make_grid.sh
```diff
@@ -133,39 +121,37 @@ case $MACHINE in
...
 #
 
 "JET")
+#
+  { save_shell_opts; set +x; } > /dev/null 2>&1
+
+  . /apps/lmod/lmod/init/sh
+  module purge
+  module load newdefaults
+  module load intel/15.0.3.187
```
JeffBeck-NOAA, line 138: @gsketefian I'm not sure how this happened, but my Jet module load changes were reverted in your PR, and things are now failing.

JeffBeck-NOAA: OK, I think I know what happened. I removed this Jet stanza in a previous PR because I have a module file for make_grid that is loaded instead. Maybe you thought these module loads needed to be put back in since it was empty?

Issue #193, opened 2020-04-13
Fix workflow end-to-end (WE2E) test configuration #13 that tries to run the FV3_GSD_SAR_v1 suite


gsketefian:
The name of the suite apparently changed from FV3_GSD_SAR_v1 to (maybe) FV3_GSD_noah.  Change suite name, template file names, etc to the correct names and ensure that this test runs successfully.

I will assign it to Linlin since he is the one who initially created this test.

jwolff-ncar:
@gsketefian I think this SDF has probably changed names at least one other time since this issue was added. Is it still necessary or can it be closed?

gsketefian:
The FV3_GSD_SAR_v1 suite has been removed from the workflow by PR #[287](https://github.com/NOAA-EMC/regional_workflow/pull/287), so this issue is no longer relevant.

PR #194, opened 2020-04-13
Add capability to run tasks outside of rocoto.

gsketefian:
Summary:
* Add capability to run tasks outside of a workflow manager like rocoto.
* Add capability to specify compute resources in the workflow configuration file (config.sh). 

Test results:
14 out of 15 workflow end-to-end (WE2E) tests were successful.  Test regional_013 was not successful but that is because the name of the physics suite that that configuration tests for has been changed (the configuration file and other template files must be fixed).

New issue to fix WE2E test regional_013 created here: https://github.com/NOAA-EMC/regional_workflow/issues/193

How to use:
The new script that can run the workflow tasks individually outside of rocoto is at ush/run_one_task.sh.  To see the help/usage, first generate an experiment (using generate_FV3SAR_wflow.sh).  In the top level of the experiment directory, there will be a link to this script (having the same name as this new script).  From that directory, type "run_task.sh --help" to get the help/usage.  It includes examples of how to call the script (slurm only thus far).


gsketefian:
Mike,

I manually merged several of the files to be sure the merge is done right.  Please review and let me know what you think.  Actually, please try running an experiment to see if it still works the way you intended.

Question:  What does the <nodesize> tag in the XML do?  I found only a brief description of it by Chris Harrop online (on github), but nothing in-depth.
 
I'm currently rerunning all my WE2E tests on hera, but it's taking a while because it looks like some nodes are down for maintenance.

Gerard

gsketefian:
Hi Laurie and Mike,

Could you also take a look at the script ush/run_one_task.sh in my PR to see if it makes sense?  Specifically, is it easily portable to say cheyenne?  I was thinking we can modify/add inputs for other schedulers, and then at the end where there is the if-statement that calls sbatch for slurm, have another elif stanza for pbspro and call the appropriate command there (qsub?).

Thanks,
Gerard

mkavulich:
Looks as if the merge was successful, I can still build and run just fine on Cheyenne. 

I'm starting to play around with the new script and I noticed that I am not able to see the usage instructions unless I'm in an experiment directory; the script fails before getting to the usage routine. Can you re-arrange this so that you can print usage instructions regardless of whether the experiment has been set up yet?

gsketefian:
Hi Mike,

Yes, that was a compromise I had to make in order to give the user more information in the usage message.  That more information is obtained by sourcing the var_defns.sh file, which is created only after generating an experiment directory.  I can instead source the config_defaults.sh file (which always exists in the repo), but the info in the usage message may be wrong in case, e.g., the user changes the task names (MAKE_GRID_TN, etc) in the user-specified configuration file (config.sh).  But I guess I can note that in the usage message (i.e. say that the values printed in the usage message are the default values).

Gerard

llpcarson:
Hi Gerard - I started looking at your script, but haven't tried to run it yet.  One overarching question, though.
- is there a reason you have added all of the various command-line-args, rather than relying on the var_defns.sh file that is generated from the experiment config file?  I can see the possible need for the UPP-fcsthr option, and, of course, the job name, but otherwise, I would think that a user would set up their configuration once in the var_defns.sh , and then run the jobs one-by-one with this script?
- this seems to add alot of complexity to the script, and likely isn't the right place to have a user change something like a queue name or wall-clock-time or even cycle-start-time?

Will continue reviewing and try a run...

llpcarson:
Another note - in order to use this, the user needs to first run the generate_FV3SAR_wflow.sh script, true?  But, that script requires that $MACHINE match exactly a named list.  So, would it make sense to add a "generic_linux" or "macos" or some other options that would generate the experiment directory on a generic platform?

gsketefian:
Hi Laurie,

I made some improvements to the run_one_task.sh script (and to other workflow scripts in places related to this issue).  I also added a new script called run_one_task_simplified.sh that has a reduced set of input arguments.  

Notes:

* The cycle-dependent tasks need to have the cycle date (cdate) specified in case there is more than one cycle in the experiment.  Currently, there is no option to run a task for all cycles (e.g. by specifying cdate="all" or something like that), and I don't know if doing so would make sense anyway since when DA is included, one cycle will depend on the tasks from a previous cycle - so I'd basically start rewriting rocoto.  Anyway, at least for now, cdate is a necessary input argument if running a cycle-dependent task.

* Similarly, for a given cdate, the post-processing task must be called one forecast hour at a time (i.e. there is no option like fhr="all" to run UPP for all forecast hours for a given cdate).  So fhr is a necessary argument for the run_post task.

* I added a new argument (exptdir) to specify the experiment directory.  If exptdir is set on the command line, then that's where the script will look for a var_defns.sh file.  If not, it assumes the current directory is the experiment directory.  This allows the script to be called from a location other than the top level of the experiment directory.  Thus, a symlink at the top level of the experiment directory to this script (which is something that is created by generate_FV3SAR_wflow.sh) is not necessary (but is still nice to have).  

* I included the remaining arguments (sched, acct, part, qos, nnodes, ppn, wtime) in the original script (run_one_task.sh) as a convenience for the user so he/she doesn't have to go to var_defns.sh and modify it (and we generally discourage modifying var_defns.sh; if one wants to change experiment parameters, it's safer to create a new config.sh in the ush directory and generate a new experiment).  In this case, changing var_defns.sh and rerunning run_one_task.sh will work I think (I will test that soon), but it seems more convenient to me to specify it on the command line so you don't have to open a file and edit it.  In any case, the new simplified script (run_one_task_simplified.sh) does not have these arguments - it gets them all from var_defns.sh.  Note that removing these arguments didn't simplify the script very much because local variables part, qos, nnodes, ppn, and wtime still have to be set depending on the task.  You can diff the two scripts to see what I mean.  We can eventually remove one of these scripts or keep both.

* The user can now set the workflow variable SCHED to "none" in config.sh (or, for run_one_task.sh, set sched=none on the command line).  This will cause run_one_task.sh or run_one_task_simplified.sh to run the j-job without a job scheduler.

Finally, yes, it would it make sense to add a "generic_linux" or "macos" for generic platforms.  However, I'm thinking we can do that as a separate PR.

The new changes pass 14 of the 15 WE2E tests (test 13 is bad and needs to be fixed).

I tried both run_one_task.sh and run_one_task_simplified.sh from the command line on all tasks, and they seem to work fine.

Please let me know what you think.  Thanks.
Gerard

llpcarson:
Hi Gerard - been in meetings all day, but thanks for considering and updating this PR.  I do wonder, though, if we need to think more simply.  I would recommend not having two scripts, one without a few options --- that's a maintenance nightmare!  

I will take a look at your other options tomorrow, but would recommend removing the "simplified" script (again, just my suggestion)

jobs/JREGIONAL_MAKE_ICS
```diff
@@ -134,7 +134,8 @@ case "$MACHINE" in
   ;;
 #
 "CHEYENNE")
-  APRUN="mpirun -np ${NPROCS}"
+  nprocs=$(( NNODES_MAKE_ICS*PPN_MAKE_ICS ))
+  APRUN="mpirun -np $nprocs"
```
christinaholtNOAA, line 138: Curious why this is sometimes set at the Jobs level, and sometimes at the ex-script level.

ush/load_modules_run_task.sh
```diff
@@ -255,61 +257,61 @@ task (task_name) does not exist:
...
 ules_dir) for the specified task (task_name) failed:
   task_name = \"${task_name}\"
   modulefile_name = \"${modulefile_name}\"
   modules_dir = \"${modules_dir}\""
-     ;;
-   #
-   esac
+    ;;
+#
+  esac
```
christinaholtNOAA, line 314: I know this isn't part of this PR, necessarily, but I noticed with my XML work that this file runs `exec` on user command line input that is completely unchecked. This opens a user developer to an extreme data-loss vulnerability, as well as any other malicious attacks. This code exists in a complicated scripting environment, and a single setting gone unnoticed could be quite problematic. Running the task script directly from the XML at least reduces the indirection in understanding whether a vulnerability exists.

gsketefian:
Hi Laurie,
Yes, having just one is better, I agree.  Let me know which one you think users would prefer and I will delete the other.
Gerard

jobs/JREGIONAL_MAKE_ICS
gsketefian: Mike asked this too a while back!

It is not clear in the WCOSS Implementation Standards document where this should go.  So back a few months or a year ago (when I was just learning how to use github), I left it like this in order for us to remember to fix this later on (I wasn't sophisticated enough with github back then to just make an issue).

So I just made an issue to fix this (https://github.com/NOAA-EMC/regional_workflow/issues/198).  Once we determine at which level these should go, we need to fix it one way or the other.

ush/load_modules_run_task.sh
gsketefian: We did not do things this way originally, but NCO apparently doesn't want module loads in the j-jobs or ex-scripts.  The method (of using exec) that we use now was devised by EMC and is apparently acceptable to NCO, so we just adopted it.  We can point out the vulnerabilities you mention (can you give an example?) to EMC/NCO and see what they say.  If you know of a more secure way to do this that is acceptable to NCO, we can go with that.

christinaholtNOAA: It's a general best practice to find a way other than "eval" or "exec" where possible, especially when there are no filters/checks in place for the ingested command to be executed. Here's a decent write up: https://medium.com/dot-debug/the-perils-of-bash-eval-cc5f9e309cae.

A comment was left in `FV3SAR_wflow.xml` that provides an alternative:

```<!-- 
    <command>{ . &LOAD_MODULES_RUN_TASK_FP; "&MAKE_GRID_TN;"; 
               &JOBSDIR;/JREGIONAL_MAKE_GRID;
             }</command>
-->
```
I understand that the `<command>` tag is essentially an eval statement for running the supplied command, and it has decently tight restrictions on what is called. 

Removing the somewhat "hidden" eval for users not running Rocoto might be a bit safer. You never know when a copy/paste gone wrong could lead to loss of data when you don't realize a command-line argument might get executed as-is.

gsketefian: Yeah, that was me who wrote that comment, and I totally forgot about it!  If you think that alternative is secure enough, let's make an issue to propose it to EMC/EIB (and then they have to ask NCO...).

PR #195, opened 2020-04-14
Remove Jet module load stanza in exregional_make_grid.sh

JeffBeck-NOAA:
Tested on Jet.


JeffBeck-NOAA:
Module loads are handled in a specific module file instead.

gsketefian:
And I don't think you need the "export" in front of APRUN any more either.

PR #204, opened 2020-05-04
Fix Cheyenne segfaults, add capability for tasks to fall back to default modulefile

mkavulich:
This PR adds the capability for tasks to fall back to default modulefile if task-specific modulefile does not exist. In doing so, it gets rid of the old, hacky way I got post to build on Cheyenne, and replaces it with a build and run process that utilizes a new cheyenne.default modulefile.

This should allow us to condense down builds on new machines to use fewer task-specific modulefiles.


ush/load_modules_run_task.sh
```diff
@@ -277,43 +287,50 @@ does not exist:
...
-
-fi #End if statement for tasks that load no modules
+       ;;
+     #
+     esac
+   else # using default modulefile
+     module load ${default_modulefile_name} || print_err_msg_exit "\
+Loading of default module file failed:
+  task_name = \"${task_name}\"
+  default_modulefile_name = \"${modulefile_name}\"
```
christinaholtNOAA: Should this be `default_modulefile_name`?

ush/load_modules_run_task.sh
```diff
@@ -229,24 +231,32 @@ modulefile_name="${task_name}"
...
+     if [ -f "${default_modules_dir}/${default_modulefile_name}" ]; then
+        #If the task-specific modulefile does not exist but a default one does, use it!
+        print_info_msg "$VERBOSE" "
+        Task-specific modulefile (${modulefile_path}) does not exist
+        Will attempt to use default modulefile (${default_modules_dir}/${default_modulefile_name}) ..."
+        modules_dir=$default_modules_dir
+        use_default_modulefile=1
 
-     if [ "${task_name}" = "${MAKE_OROG_TN}" ] || \
+     elif [ "${task_name}" = "${MAKE_OROG_TN}" ] || \
```
christinaholtNOAA, line 251: Will logic ever get to this block if you want to tell the user they should build the code first?

ush/load_modules_run_task.sh
mkavulich: Yes, fixed

ush/load_modules_run_task.sh
mkavulich: I don't think that's a problem. If the user has managed to get this far without building the code they have already done something horribly wrong. Also, even if they did manage to get this far without building the code, the subsequently called exscript should check for that binary and see that it is missing, and fail accordingly.

PR #205, opened 2020-05-05
Cheyenne improvements: include pre-build wgrib2, allow valid_param_vals to be machine-specific

mkavulich:
Two improvements to the user experience on Cheyenne:

# Include pre-built wgrib2

The original Cheyenne capability was limited by the lack of a wgrib2 module or linked binary. I did not realize this initially because I had a wgrib2 binary in my own personal path.

I went with the easy fix and created a directory for hosting cheyenne-specific binaries needed for the regional_workflow (/glade/p/ral/jntp/UFS_CAM/bin). This then included in the path when the various "modulefiles" (actually shell scripts) are referenced. This also includes a link to the pre-built rocoto binaries, which users may find useful.

Eventually we should consider taking the wgrib2 binary from the NCEPLIBS build automatically

# Allow valid_param_vals to be machine-specific

The file ush/valid_param_vals.sh contains arrays of strings that contain all the valid values for a given user-specified variable. These are checked using the "check_var_valid_value" function, which currently takes in 2 or 3 arguments: the variable name, the name of the array of valid values, and an optional error message.

I have modified this script to take an additional optional argument: the machine name. These are disambiguated using the prefix "msg=" for the error message, and/or "mach=" for the machine name.

The machine-specific arrays build on the previous convention of ush/valid_param_vals.sh: previously all the arrays of valid values were named "valid_vals_VAR", where VAR was the relevant variable name for comparison. Now we can *also* allow machine-specific arrays of valid values, using the naming convention "MACHINE_valid_vals_VAR". If the machine-specific array does not exist for a given machine, the script falls back to the default non-machine-specific valid_vals_ array.

This effort was undertaken because the valid queue values are different for different machines, so it is helpful for the user to make this machine-specific to avoid hard-to-solve runtime errors. I anticipate this capability will be helpful for other variables in the future as well.

# Tests run
This has been tested by running the workflow generation script for several different config.sh variations on both Cheyenne and Hera. As was anticipated/hoped, current behavior does not change on Hera. On Cheyenne, wgrib2 is now found even if the user does not have it in their PATH, and the generate_workflow step will fail if the user specifies a queue that is not supported on Cheyenne.


ush/bash_utils/check_var_valid_value.sh
```diff
@@ -97,14 +80,72 @@ where the arguments are defined as follows:
...
-  fi
+  machine=''
+
+  while [ "$#" -gt 2 ]; do
+    # Don't be scared of the scary bash variable manipulations! Quick tutorial:
+    # ${abc%:*}  Represents everything BEFORE the FIRST "=" character from the string variable abc
+    # ${def%%:*} Represents everything BEFORE the LAST ":" character from the string variable def
+    # ${ghij#*=} Represents everything AFTER the FIRST "=" character from the string variable ghij
+    # ${klm##*!} Represents everything AFTER the LAST "!" character from the string variable klm
+    if [ "${3%%=*}" == "msg" ]; then
```
christinaholtNOAA, line 94: Can the user only specify one or the other of these? Could one of them be ${4}?

christinaholtNOAA: TIL about `shift`! Mike addressed this one in person.

gsketefian: TIL what TIL means :)

gsketefian:
Hi Mike,

A simpler solution for this might be to have machine-specific valid-parameter-values files, e.g. valid_param_vals.cheyenne.sh.  Then, in setup.sh, once the machine name is checked to be valid, that machine-specific file gets sourced (as long as it exists).  That file can just overwrite the arrays in the default valid_param_vals.sh file, and/or it can define new ones.  So something like this in setup.sh:

```
MACHINE="${MACHINE^^}"
check_var_valid_value "MACHINE" "valid_vals_MACHINE"

machine="${MACHINE,,}"
valid_vals_fp="$USHDIR/valid_param_vals.$machine.sh"  # We can put this file elsewhere.
if [ -f "${valid_vals_fp}" ]; then
  . "${valid_vals_fp}"
fi
```

For cheyenne, valid_param_vals.cheyenne.sh can contain:

`valid_vals_QUEUE=("premium" "regular" "economy" "share")
`

This would overwrite the value of `valid_vals_QUEUE` in the default file.  What do you think?

Gerard

mkavulich:
@gsketefian My initial approach was along those lines of thinking, but I wanted to avoid further proliferation of machine-specific files, and avoid ending up with potentially conflicting variable definitions in different places, since that could end up giving the user unexpected behavior. In both cases we may potentially have to add/change a variable in two different places, but at least in this case those two places will be in the same file. It admittedly does make the check_var_valid_value.sh script more complicated, but I think that's a price worth paying to avoid cluttering the workflow with *even more* machine-specific files.

gsketefian:
Hi Mike,

Certainly the additions don't belong in check_var_valid_value.sh.  That's just a utility function that should remain as simple as possible without needing to know anything about the machine.  You can do the complicated machine-dependent logic in setup.sh; that's a more appropriate place.  Or write another script that setup.sh calls.

I am far more inclined to use machine-specific configuration files than have hard-coded machine-dependent code in the scripts.  We just have to document what those machine-specific files are for and what variables they need to contain.  To reduce the number of such files, we can try to combine them into one file using jinja, for example, just like Christina did for the various physics suites.  If we go down the road of changing the code to have more and more machine-dependent behavior, it will turn into a bunch of if/case statements that do different things for different machines.  Then we might as well just have a separate workflow for each machine.  The machine-specific config files are the lesser of the two evils, imo.

Gerard

gsketefian:
Hey Mike:
Btw, there's already a lot of machine-dependent code in the experiment generation scripts in ush.  See for example set_extrn_mdl_params.sh.  At the time I wrote that code, I hadn't thought about making machine-specific config files for these.  Since you don't like such files, we'll leave that code as is.  So you can make a new script that setup.sh calls that sets valid_vals_WHATEVER to an appropriate value for the machine.  And probably remove the setting of valid_vals_WHATEVER from valid_param_vals.sh (and maybe rename valid_param_vals.sh to something that indicates that it only contains machine-independent parameters, e.g. valid_param_vals.any_machine.sh).
G

christinaholtNOAA:
I agree that it would be a great approach to pull out all machine-specific settings in the scripts at all levels of the workflow. I believe nearly every script currently has at least one machine case statement, if not multiples.

Even before there is the ability to use YAML config files throughout, there's a straightforward way to handle this in bash. A single machine.sh file with a machine input argument (e.g., "hera", "cheyenne") could include a single case statement for all machine-specific variables. That script would be the oracle for "hardcoded" paths/settings/etc needed by all scripts.

gsketefian:
Yes Christina, I think what you suggest is a good idea.  And I think it is doable for most workflow variables.  I don't want to say all variables because there might be some dependencies between variables that I haven't thought of (e.g. a variable might depend not only on machine but also on external model, etc).  We'll have to try and see.

mkavulich:
Per discussion here, will re-work and re-open as a new PR.

PR #212, opened 2020-05-11
Improve the way fixed files are handled.

gsketefian:
This PR makes the following modifications:

* Improve the way fixed files are handled so that what is happening with the fixed files is clearer.
* Use the surface climatology files on the FV3SAR grid in input.nml (instead of the global files in the system directory FIXgsm).
* Reduce the number of symlinks needed in each cycle directory.
* Fix the EMC_CONUS_3km grid settings so that it has a valid BLOCKSIZE.
* Add new WE2E test for EMC_CONUS_3km grid.

On hera, this PR passes all WE2E tests except regional_013 (which tests a physics suite for which an XML does not exist).


gsketefian:
@mkavulich Can you test this on Cheyenne?
@JeffBeck-NOAA (or anyone else) Can you test on Jet?

Thx.

JeffBeck-NOAA:
Can confirm that it works on Jet.

ush/templates/FV3SAR_wflow.xml
```diff
@@ -3,39 +3,23 @@
 <!DOCTYPE workflow [
 
 <!--
-User-specific variables that are needed for batch job submission and
-workflow control. They are used within the XML and are not passed to
-the shell scripts 
+************************************************************************
+Entities that are directly modified by the workflow generation scripts.
+************************************************************************
```
christinaholtNOAA: My personal opinion is that it doesn't make sense to break it up like this after this template transitioned to a Jinja2 template. It's also misleading to imply that the file isn't changed below this section. Entities, by convention, are only declared at the beginning of an XML file, so we don't get much useful help with this comment. 

This organization also leads to duplication of information since the template can handle putting the script variables directly where they're used without the use of indirection with entities. See next comment.

I think we should really aim to make this file, in particular, as streamlined as possible, and declare only the ENTITIES Rocoto needs. We are already seeing issues with running Rocoto at more frequent intervals on the HPC machines, so we should pay close attention to coding efficiency here. Reducing the amount of work to fill in variables with every iteration may seem like a small part of that, but every little bit helps.

ush/templates/FV3SAR_wflow.xml
```diff
@@ -47,47 +31,100 @@ Rocoto task names.
...
 
-<!ENTITY PROC_POST                "2:ppn=24">
-<!ENTITY NPROCS_POST              "48">
+<!--
+Number of physical cores per node for the current machine.  This is used
+below in the <nodesize> tag, but that tag is not clearly documented.  This
+parameter may be unnecessary since each task now has its own variable that
+specifies the number of processes per node being used (the PPN_... entities).
+-->
+<!ENTITY NCORES_PER_NODE "{{ ncores_per_node }}">
```
christinaholtNOAA, line 40: I'm curious if this tag is still needed, and even potentially dangerous, with the inclusion of the `<nodes>` tags that are now also included for each task.

ush/templates/FV3SAR_wflow.xml
```diff
@@ -47,47 +31,100 @@ Rocoto task names.
...
 -->
-<!ENTITY LOAD_MODULES_RUN_TASK_FP "&USHDIR;/load_modules_run_task.sh">
+<!ENTITY PROC_MAKE_GRID           "&NNODES_MAKE_GRID;:ppn=&PPN_MAKE_GRID;">
+<!ENTITY PROC_MAKE_OROG           "&NNODES_MAKE_OROG;:ppn=&PPN_MAKE_OROG;">
+<!ENTITY PROC_MAKE_SFC_CLIMO      "&NNODES_MAKE_SFC_CLIMO;:ppn=&PPN_MAKE_SFC_CLIMO;">
+<!ENTITY PROC_GET_EXTRN_MDL_FILES "&NNODES_GET_EXTRN_MDL_FILES;:ppn=&PPN_GET_EXTRN_MDL_FILES;">
+<!ENTITY PROC_MAKE_ICS            "&NNODES_MAKE_ICS;:ppn=&PPN_MAKE_ICS;">
+<!ENTITY PROC_MAKE_LBCS           "&NNODES_MAKE_LBCS;:ppn=&PPN_MAKE_LBCS;">
+<!ENTITY PROC_RUN_FCST            "&NNODES_RUN_FCST;:ppn=&PPN_RUN_FCST;">
+<!ENTITY PROC_RUN_POST            "&NNODES_RUN_POST;:ppn=&PPN_RUN_POST;">
```
christinaholtNOAA: An example to follow my previous comment:

```
<!ENTITY PROC_RUN_POST            "{{ nnodes_run_post }}:ppn={{ ppn_run_post }}">
```

Which allows you to remove 16 additional lines above (36-43, and 49-56). The same applies for the the walltime tags in the next section.

gsketefian: Yes, that's a good idea.  I'll make those changes.

ush/templates/FV3SAR_wflow.xml
gsketefian: @mkavulich I think you included this when porting to cheyenne.  What exactly does <nodesize> do, and is it still needed for cheyenne?

mkavulich: Running on Cheyenne without the <nodesize> tag caused rocoto to spit out a number of warning messages every time you submitted a task. It also caused tasks to default to assuming 72 cores per node, which is incorrect and I assume would cause problems. It is *possible* this is no longer needed, as I was running with an older version of rocoto at the time, but I suspect that removing it will cause issues. I can run some tests if you like; I'm currently running through the tests of the current state of this branch.

gsketefian: @mkavulich I'd say finish the tests with the current state of the branch, and once those pass, we can think about this issue.  With the new version of rocoto, I wonder if it will still assume 72 cores per node (or whether that will matter at all) when you explicitly specify a ppn (just thinking out loud; you don't have to answer that!).

ush/templates/FV3SAR_wflow.xml
gsketefian: @christinaholtNOAA Yes, I agree it doesn't make sense now with what jinja2 can do.  

So do you think we should replace pretty much all ENTITY declarations with jinja2 placeholders (or whatever they're called; the things in double curly braces)?  E.g., replace all occurrences of `&ACCOUNT;` with `{{ account }}` and eliminate the ENTITY declaration for it?

christinaholtNOAA: I think striking a balance is probably a good option. For variables that are heavily re-used ENTITIES are probably the way to go. For example, if I wanted to manually change the `&ACCOUNT;` midway through a long retro, I'd only need to do it in a single place with the entity. If it's a value that's only used in one place, maybe it is better for Jinja to go ahead and fill it in up front.

gsketefian: Ok, I'll try to strike a reasonable balance.  Thx!

gsketefian:
@christinaholtNOAA Please take a look at the latest version of `ush/templates/FV3SAR_wflow.xml` and let me know what you think.  Thx!

christinaholtNOAA:
@gsketefian It looks much more streamlined, and efficient. Thanks!

gsketefian:
@christinaholtNOAA (and others):
I'm bringing in a new physics suite (FV3_GFS_2017_gfdlmp_regional) that EMC wants (basically including Ratko's PR #197 here).  I noticed that in FV3.input.yml, the variables are sorted alphabetically, which is nice because it makes it easier to find a variable.  I'm thinking of doing the same in input.nml.FV3 so its easier to find variables in there as well.  Any objections?

JeffBeck-NOAA:
No objections here, although, when we collaborate with other groups, they
may not be used to that organization when we share namelists.  Also, I hope
it doesn't take too long to do!

- Jeff

christinaholtNOAA:
I have no objections! It's so much easier to find everything when it's alphabetical. The set_namelist.py script may help. I added the ability to create a YML file that is the difference between two input namelists. You should be able to set two additional options in the command line arguments to have it do that. Use `-h` to check out all the options to that script.

If you manually edit set_namelist.py to have the `f90nml.write()` command use `order=True` it should do the work for you. Check out the f90nml options here: https://f90nml.readthedocs.io/en/latest/usage.html

gsketefian:
@JeffBeck-NOAA Yes, that's my concern, too, that people aren't used to it and will complain about the new order.  But let's try it and see what they say.
@christinaholtNOAA Thanks for adding the difference capability.  Very useful.  I had to do that manually yesterday, and it was a pain and error-prone.  And yes, I'll probably use order=True to make the final namelist ordered alphabetically.

gsketefian:
Hi all,

I tested this on hera but would like others to test as well, as follows:

@JeffBeck-NOAA : Please test on Jet.
@RatkoVasic-NOAA : Please test on Hera.
@mkavulich : Please test on Cheyenne.
@BenjaminBlake-NOAA : Please test on WCOSS Cray/Dell.

Also, @christinaholtNOAA : 
Can you please generate an experiment with this and take a look at the `fsmcl` variable in input.nml?  I'm getting `fsmcl = 99999, 99999, 99999` but I should be getting `fsmcl(2:4) = 99999, 99999, 99999`.  I must have changed something inadvertently, and you can probably find it a lot faster than me.

Thanks all!

BenjaminBlake-NOAA:
Hey Gerard, I should be able to test the changes on WCOSS tomorrow after they are finished with today's production switch.  Thanks!

gsketefian:
@BenjaminBlake-NOAA Ok, sounds good!

Also, if anyone needs help running the WE2E tests, let me know.

RatkoVasic-NOAA:
Hi Gerard,
I managed to run NCO version on Hera end-to-end. Just one small thing:

1) In file ./sorc/build_forecast.sh, line
#EMC  ./compile.sh "$FV3" "$target" "NCEP64LEV=Y HYDRO=N 32BIT=Y CCPP=Y STATIC=Y SUITES=FV3_GFS_2017_gfdlmp_regional" 1
please remove 1 at the end. This creates executable named fv3_1.exe, while workflow is looking for fv3.exe

2) Could you please include in README.md Users Guide link.
Also, step 3 is missing in the same file, and Github web page is putting steps 4 and 5 in the same line.

OK, those were two and half things.
Thanks!

gsketefian:
That's great Ratko!  I hope it was relatively straightforward.  

I made the change to the build_forecast.sh script and will push it as part of this PR.  Also, I created a new PR (#220) for editing the README.md file.  I made the fixes you suggest except that I am not sure what User's Guide to point to.  Let's discuss it in that other PR.

Also, maybe you already know, but just wanted to mention that the following two workflow test configurations (in tests/baseline_configs) are for the EMC CONUS grid with the FV3_GFS_2017_gfdlmp_regional suite:

config.nco_conus.sh
config.nco_conus_c96.sh

The first is the 3km grid and the other is with the much coarser (~23.4km) grid.  I usually run the coarse grid for debugging purposes, and once that passes, then I run the 3km one.

BenjaminBlake-NOAA:
I completed a successful test on the WCOSS Cray with my WCOSS modifications.  Next I will test on the WCOSS Dell and then we will be good to go on WCOSS.  For the WCOSS mods, should I add those to this PR once I complete my Dell testing this week or should I wait until this gets merged and then submit a new PR?

JeffBeck-NOAA:
Testing on Jet is on hold until we modify the external model file variable to be independent of location (HPSS vs. on disk). I will test the upcoming PR that will contain that change.

gsketefian:
@BenjaminBlake-NOAA,:
If most/all of the changes you're making are specific to the WCOSS Cray and Dell, then I suggest doing it as a separate PR since you can describe those changes a lot better than I can (and to keep this PR from getting too large).  It's looking like I need to make another branch/PR to get things working on Jet, and @JeffBeck-NOAA and I are going to iterate on that the next couple of days.

Once you get this PR working on both WCOSS machines, make a new branch/PR for your changes, and we'll get the two PRs in one right after the other.  Then we'll merge those changes into the jet PR and get that in as quickly as we can.  I haven't heard from @mkavulich about how things are going on Cheyenne.

If you have suggestions on how to make the testing (of any PR) easier on WCOSS Cray/Dell, let us know, e.g. maybe include some new WE2E tests for those two machines.

BenjaminBlake-NOAA:
@gsketefian That sounds good to me, I should be done with my WCOSS testing later today (I hope because there is an emergency production switch tomorrow morning).  Then I will work on creating a separate PR as soon as this one goes in, which I assume will be very soon?

BenjaminBlake-NOAA:
I am done with testing on WCOSS and can submit a separate PR for those changes once this one goes in.

BenjaminBlake-NOAA:
Can I go ahead and merge this PR if we are labeling it as Hera-specific only?  I want to create my WCOSS PR using the changes from this one.  Thanks!

gsketefian:
@BenjaminBlake-NOAA 
I was hoping to hear from Mike about whether this works on Cheyenne since I don't want to break things for the workflow focus team testers (Ben, Rocky, etc).  

@mkavulich 
Any updates?  Is it ok if we merge it now?

Thanks.

gsketefian:
@mkavulich tested this on cheyenne, and it passed, so I will merge.

PR #213, opened 2020-05-13
Feature/nssl stampede

ywangwof:
These commits add support for Stampede2.tacc.utexas.edu and fix support for Odin (a NSSL Cray machine).


sorc/build_forecast.sh
```diff
@@ -52,7 +55,9 @@ CCPP=${CCPP:-"false"}
 cd tests/
 if [ $CCPP  = true ] || [ $CCPP = TRUE ] ; then
 #EMC  ./compile.sh "$FV3" "$target" "NCEP64LEV=Y HYDRO=N 32BIT=Y CCPP=Y STATIC=Y SUITES=FV3_GFS_2017_gfdlmp_regional" 1
-  ./compile.sh "$FV3" "$target" "CCPP=Y STATIC=N 32BIT=Y REPRO=Y"
+  #./compile.sh "$FV3" "$target" "CCPP=Y STATIC=N 32BIT=Y REPRO=Y"
+  ./compile.sh ${FV3} "$target" "CCPP=Y HYBRID=N STATIC=Y SUITES=FV3_GFS_v15_thompson_mynn,FV3_GSD_v0,FV3_GSD_noah 32BIT=Y REPRO=Y INTEL18=Y" 32bit_static YES NO
```
JeffBeck-NOAA: @ywangwof @gsketefian Changing to a static build is a good idea; however, you'll need to add the FV3_SAR_GSD for our current RRFS runs and FV3_GFS_2017_gfdlmp_regional for EMC.

ywangwof: Sure. I will add those CCPP suites mentioned. @gsketefian, Do you have the config.sh files for your end-to-end tests on Jet? If so, I can just run my branch with your configuration files to make sure they are still works on Jet. Otherwise, direct me what configuration will make one "end-to-end tests" Jeff mentioned. Thanks.

modulefiles/tasks/odin/get_extrn_lbcs
```diff
@@ -0,0 +1,8 @@
+#%Module#####################################################
+## Module file for get_extrn_lbcs task.
+#############################################################
+
+#module purge
+
+#module load hpss
+
```
christinaholtNOAA, line 8: This file doesn't actually do anything. Is it needed?

JeffBeck-NOAA:
Yunheng, Gerard,

I'm also wondering if we should include the other options listed in the
ush/templates directory for the CCPP build.  If a user tries one of those
suites, it will crash unless it's included in the static CCPP/FV3 build.
That's starting to be a lot of physics suites, though.  Gerard, what are
your thoughts?  The more we limit the number of schemes that are built, the
faster the executable, but the less flexibility the user has to choose a
scheme.  We have to do this anyway, though, since the next version of CCPP
requires static builds.

- Jeff

jobs/JREGIONAL_MAKE_ICS
```diff
@@ -131,12 +131,18 @@ case "$MACHINE" in
...
+  APRUN="srun"
   ;;
 #
 "CHEYENNE")
   APRUN="mpirun -np ${NPROCS}"
   ;;
 #
+"STAMPEDE")
+  APRUN="ibrun"
+  module load python3
```
christinaholtNOAA: Modules are typically not loaded here.

ywangwof:
> Yunheng, Gerard, I'm also wondering if we should include the other options listed in the ush/templates directory for the CCPP build. If a user tries one of those suites, it will crash unless it's included in the static CCPP/FV3 build. That's starting to be a lot of physics suites, though. Gerard, what are your thoughts? The more we limit the number of schemes that are built, the faster the executable, but the less flexibility the user has to choose a scheme. We have to do this anyway, though, since the next version of CCPP requires static builds. - Jeff
> […](#)

Jeff and Gerard,

For me, I would like to append the suite name to the built executable. For example,

```
suites=(FV3_GFS_v15_thompson_mynn FV3_GSD_v0 FV3_GSD_noah)

for suite in ${suites[@]}; do
    compile.sh ${FV3} "$target" "CCPP=Y HYBRID=N STATIC=Y SUITES=${suite} 32BIT=Y REPRO=Y INTEL18=N" ${suite##FV3_} YES NO
done
```

After a successful build, we will get,

```
fv3_GFS_v15_thompson_mynn.exe
fv3_GSD_v0.exe
fv3_GSD_noah.exe
```

And the workflow can look for the specific executable based on the CCPP suite user specified in “_config.sh_”. If not exist, an exception will be raised.

Yunheng.

jobs/JREGIONAL_MAKE_ICS
ywangwof: Thanks for pointing it out. Actually, the module load command does not play any role there. I am wondering how to force the user to load module python3 before running any script from the workflow. Any idea?

Actually, there are indeed module load commands there. For example, the section for THEIA. I know "theia" is not supported any more. But I am following the existing coding style.

JeffBeck-NOAA:
> > Yunheng, Gerard, I'm also wondering if we should include the other options listed in the ush/templates directory for the CCPP build. If a user tries one of those suites, it will crash unless it's included in the static CCPP/FV3 build. That's starting to be a lot of physics suites, though. Gerard, what are your thoughts? The more we limit the number of schemes that are built, the faster the executable, but the less flexibility the user has to choose a scheme. We have to do this anyway, though, since the next version of CCPP requires static builds. - Jeff
> > […](#)
> 
> Jeff and Gerard,
> 
> For me, I would like to append the suite name to the built executable. For example,
> 
> ```
> suites=(FV3_GFS_v15_thompson_mynn FV3_GSD_v0 FV3_GSD_noah)
> 
> for suite in ${suites[@]}; do
>     compile.sh ${FV3} "$target" "CCPP=Y HYBRID=N STATIC=Y SUITES=${suite} 32BIT=Y REPRO=Y INTEL18=N" ${suite##FV3_} YES NO
> done
> ```
> 
> After a successful build, we will get,
> 
> ```
> fv3_GFS_v15_thompson_mynn.exe
> fv3_GSD_v0.exe
> fv3_GSD_noah.exe
> ```
> 
> And the workflow can look for the specific executable based on the CCPP suite user specified in “_config.sh_”. If not exist, an exception will be raised.
> 
> Yunheng.

I had no idea this was possible (multiple executables for each static CCPP suite.  I think this may be the best solution to the static problem.  It might take a bit longer to build (I'm not sure), but at least it will speed up the iteration of the model. @gsketefian , what do you think?

jobs/JREGIONAL_MAKE_ICS
JeffBeck-NOAA: @ywangwof We're handling the requirement of loading python3 outside of the workflow, with the onus on the user to do it.  There will be information in the documentation about that.  Regarding the Theia stanza, that was before we moved all module loads out of the j-job/ex-script layer to make it standardized for NCO.  We should completely remove any Theia information that's still left in the workflow.

gsketefian:
@ywangwof @JeffBeck-NOAA:
I think Yunheng's approach is fine.  During the experiment generation step, the script looks in the $HOMErrfs/exec directory for the presence of the specific executable corresponding to the suite.  If it's not there, it issues an error message and quits.

I'm assuming here that the build script will copy all executables from $HOMErrfs/sorc/ufs-weather-model/tests to $HOMErrfs/exec at the end of the build step.

Another question is how to specify which physics suites to build executables for.  There should be a default in the build script that includes all the ones we're going to support for the release and others for which we're running tests.  In addition, there should be a way to override that default on the command line so that if somebody is interested in say only one suite, they don't have to spend time building all the others.

gsketefian:
@ywangwof Are there module files for stampede in the UFS_UTILS repo?  If so, as @christinaholtNOAA mentioned, we can use those for the most part to be sure we're using the same software versions as the build.  If not, we can make them in the NCAR fork of UFS_UTILS since we have control over that --  and will eventually go into the NOAA-EMC authoritative repo with all the other changes we need.  It would be nice not to keep copies of module files in this repo that we'd need to update manually.

ywangwof:
> @ywangwof Are there module files for stampede in the UFS_UTILS repo? If so, as @christinaholtNOAA mentioned, we can use those for the most part to be sure we're using the same software versions as the build. If not, we can make them in the NCAR fork of UFS_UTILS since we have control over that -- and will eventually go into the NOAA-EMC authoritative repo with all the other changes we need. It would be nice not to keep copies of module files in this repo that we'd need to update manually.

No. The module files for UFS_UTILS are not in the repo. Should I tar the files and then pass them to you for adding to the NCAR repo?

JeffBeck-NOAA:
Yunheng,

Yes.  Please do.  We'll incorporate them into the NCAR fork of UFS_UTILS.

Gerard, I figured the list of suites would be contained within the
build_forecast.sh script in the /sorc directory.  Users could go in there
and remove/add what they want.  If we want to, it's also possible to create
a "suites list text file" that would sit in /sorc that would have a long
list of possible suite files that users could add/comment out. This would
be similar to what Yunheng has already suggested:

suites=(FV3_GFS_v15_thompson_mynn FV3_GSD_v0 FV3_GSD_noah ... )

The suites list text file with this "suites" variable could be sourced by
the build_forecast.sh script and used in the compile command line step.

- Jeff

gsketefian:
Jeff and Yuneng,

As I thought about it a bit more, I don't know if EMC is going to go along with putting modulefiles in the NOAA-EMC/UFS_UTILS authoritative repo for machines other than the few that EMC uses.  They may want to keep the repo from sprawling out of control.  This will be a good test.  If they don't accept new modulefiles, I guess we'll have to keep them in the NCAR fork.

For specifying the suites to build, I'd say having a configuration file containing a list of suites is a good idea - as long as that file is not part of the repo because then it would get changed all the time.  I'd say have the short script Yunheng wrote optionally read in the name of this config file.  If no file name is specified on the command line, the script uses a default set of suites.  That's what I do now in the script that runs the workflow end-to-end tests -- the user optionally passes to the script the name of a text file containing a list of test names.  If such a file is not specified, all the tests are run.

llpcarson:
I haven't been following this PR closely, but wanted to comment on the suites to compile.  I would strongly discourage building in some different protocols than what is already being done in the ufs-weather-model with respect to which suites get compiled.  Right now, one can specify one-or-many SDFs at compile time (arg or envar), or one can specify no SDFs, and the build will compile all of the SDFs it finds in the FV3/ccpp/suites directory.  Recompiling the model for many-suites with one suite per executable is an extreme overhead, unless there is a substantial reason to require this?  For operations, for example, just one suite would be compiled, of course.

gsketefian:
@llpcarson If more than one suite is specified at compile time, is the result just one executable that can run any one of those specified suites, or are multiple executables generated (i.e. one per suite)?

llpcarson:
It is one executable that can run any of the suites that were compiled (listed in the arg/env or in the directory).  If this list is very long, the compile can take awhile, but for less than 10 or so, it's negligible.  The "compile all in the suites directory" is a recent change, so the version of ufs-weather-model used by regional_workflow may be far enough behind to not have that capability.

JeffBeck-NOAA:
My concern moving forward is that CCPP will have a static-only option.  So
we can obviously list all the executables that we support in the community
workflow, however, the model will run faster if the user only compiles the
CCPP suite/suites they're interested in.  Maybe we can just provide a
variable for users to list the suites they want to compile, then use that
in the compile.sh command?

- Jeff

JulieSchramm:
The basic premise that I have been using is building the ufs-weather-model
using build.sh with

export CCPP_SUITES="
FV3_GFS_2017_gfdlmp,FV3_GSD_v0,FV3_GSD_SAR,FV3_CPT_v0,FV3_GFS_v15p2,FV3_GFS_v16beta
"


which are defined in regional_workflow/ush/valid_param_vals.sh


valid_vals_CCPP_PHYS_SUITE=( \

"FV3_GFS_2017_gfdlmp" \

"FV3_GSD_v0" \

"FV3_GSD_SAR" \

"FV3_CPT_v0" \

"FV3_GFS_v15p2" \

"FV3_GFS_v16beta" \

"FV3_GSD_SAR_v1" )

I had to exclude FV3_GSD_SAR_v1, since it wasn't in the latest
ufs-weather-model list of suites.

llpcarson:
@JeffBeck-NOAA No difference in RUN time, but in compile time.  But, it's a much faster compile to build in 10 suites into one executable, than to build 10 executables!  RUN-time is no different.  The static build runs faster than the old dynamically linked version.

sorc/build_forecast.sh
ywangwof: Jeff and all,

The latest community_develop branch (168c7dcf on 2020-05-20) did not work on Jet. It is not because of my code changes, but caused by the newly add python script "create_xml.py". The error message is 
`KeyError: 'range does not exist in user-supplied settings!'`
"range" should be a builtin function in Python.  I do not know how to set this variable in "user-supplied settings". Furthermore, there is no default setting in "config_defaults.sh". Any idea? Thanks.

JeffBeck-NOAA:
Hi Yunheng,

Please try running "module purge", "module load contrib miniconda3", then
"conda activate regional_workflow".  This is required to run the community
workflow on Hera and Jet with Python.   Make sure you run these commands
after you've compiled the code.

The user must generate their own config.sh script.  They can use one of the
templates that are provided, but it's up to the user to create this
script.  The config_defaults.sh script shouldn't be touched, and is used
when the config.sh file created by the user doesn't contain certain
necessary variables.

Hope that helps!

- Jeff

christinaholtNOAA:
@JeffBeck-NOAA,

I wasn't entirely sure what your last comment was in response to. I will mention that the miniconda3 module, and the supported conda environments are ONLY available on Jet and Hera. If Yunheung is working on Stampede. I'm fairly certain, those won't work. 

It will be up to users on new platforms to generate an environment that has the appropriate version of Python and the required packages. I've provided a way to create such an environment with conda using the environment.yml file.

ywangwof:
Thanks Jeff, 

It is not working for "generate_FV3SAR_wflow.sh". But it cannot submit job "make_grid" because of this error, "Invalid qos specification" and the job script likes this,

```
#! /bin/sh
#SBATCH -t 00:10:00
#SBATCH --qos=xjet
#SBATCH --account=hpc-wof1
#SBATCH --nodes=1-1
#SBATCH --tasks-per-node=24
#SBATCH --job-name=make_grid
#SBATCH -o /mnt/lfs4/NAGAPE/hpc-wof1/ywang/regional_fv3/run_dirs/GDAS05303km/log/make_grid.log
#SBATCH --comment=67dba5bf21ca84f90140ff6b1794ae74
export GLOBAL_VAR_DEFNS_FP='/mnt/lfs4/NAGAPE/hpc-wof1/ywang/regional_fv3/run_dirs/GDAS05303km/var_defns.sh'
export PDY='20190530'
/mnt/lfs4/NAGAPE/hpc-wof1/ywang/regional_fv3/regional_workflow/ush/load_modules_run_task.sh "make_grid" "/mnt/lfs4/NAGAPE/hpc-wof1/ywang/regional_fv3/regional_workflow/jobs/JREGIONAL_MAKE_GRID"
```

I do not know why rokoto adds the "--qos" option?

christinaholtNOAA:
Yunheng,

The `-qos` flag is pretty standard for submitting jobs to the Slurm batch system. Does Stampede use Slurm or PBS? Or maybe something different?

ywangwof:
Christina, Thanks for the reply. It is actually on Jet. It seems that "--qos" option takes "QUEUE_DEFAULT" variables from "config.sh". I do not know how to specify those parameters on Jet. Any experience?

dmwright526:
Hi Yunheng,

I have been running on Jet and have QUEUE_DEFAULT="batch" in my config.sh.

Hope this helps!

JeffBeck-NOAA:
Yunheng,

As David said, you need to define QUEUE_DEFAULT in your config.sh file.
Hopefully that fixes your problem.

- Jeff

JeffBeck-NOAA:
Thanks, Christina.  Yunheng was having problems running on Jet, so my
instructions were only for that platform.  Is there documentation on how to
create an environment for Python and the required packages that we can
include in the users' guide?

Thanks!

- Jeff

christinaholtNOAA:
This whole discussion makes so much more sense if Yunheng is running on Jet! 

Information for creating conda environments is available here: https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html

ywangwof:
Hi Jeff and all,

Thanks for your help. My test with 3 km run is successful on Jet with this branch. So it is ready to be checked in.

JeffBeck-NOAA:
@ywangwof Hi Yunheng, I just took a look at your PR again, and it looks like there are still some changes that need to take place since we've moved to using the ufs_srweather_app umbrella repository.  All the build_*.sh scripts are in that umbrella repo now, so you'll need to remove those from here and open a PR in that repo.  Also, aside from the global_equiv_resol, mosaic_file, and regional_grid utilities, the modulefiles are now copied over from their respective repositories instead of using hardcoded files.  Ideally the hardcoded files you provided in this PR should be added to the NCAR fork of UFS_UTILS (in both the develop and feature/chgres_grib2 branches).  Are you comfortable making those changes there?  Finally, we want to put all module loads that are specific to the run-time environment in the .local files that we store in the modulefiles directory in regional_workflow, instead of invoking them in the ex-scripts.  We also want to move all export commands to setenv("var_name", "var_value") in those .local files.  You can contact @gsketefian for more details on that.

We definitely want Stampede and Odin support for the SRW App, so while these changes might take some time to implement, it will keep all our platform-specific settings organized in a similar manner across files.  Please let me know if you'd like to discuss these changes, and we can arrange a meeting with everyone.  Thank you!

- Jeff

PR #215, opened 2020-05-18
Add Python plotting script for FV3SAR grib2 output

BenjaminBlake-NOAA:
This Python script creates plots of the following variables with FV3SAR grib2 output:

Sea level pressure
2-m temperature
2-m dew point temperature
10-m wind speed
Surface-based CAPE/CIN
500-mb heights/winds/vorticity
250-mb wind speed
Total precipitation
Composite reflectivity
Max/Min Hourly 2-5 km Updraft Helicity

It was originally written in Python 2 but was converted to Python 3 and tested on Hera.

### DIFF
--- a/ush/Python/plot_allvars.py
+++ b/ush/Python/plot_allvars.py
@@ -1,0 +1,773 @@
+################################################################################
+####  Python Script Documentation Block
+#                      
+# Script name:       	plot_allvars.py
+# Script description:  	Generates plots from FV3SAR post processed grib2 output
+#			over the CONUS
+#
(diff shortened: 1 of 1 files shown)


BenjaminBlake-NOAA:
Jeff I didn't realize you were already selected so that is why I accidentally removed you.

JeffBeck-NOAA:
@mkavulich @JulieSchramm The instructions at the top of the Python script seem straight-forward to me.  There's an example at the bottom of those instructions.  Do either of you have additional concerns?

PR #221, opened 2020-05-29
Allow names of external model files to be different on disk vs. from HPSS

gsketefian:
This needs to be merged after PR #212. 

Summary of Major Modifications:
------------------------------
* Allow the names of the external model files copied from or linked to on disk to be different than the ones extracted from an archive file in HPSS.  This is the case on jet.
* In each of the get_extrn_ics and get_extrn_lbcs workflow tasks, after obtaining the external model files (either from disk or from HPSS), create a new file and save in it certain parameters that depend on this process (e.g. the names of the external model files that were obtained, since now these names depend on whether they are obtained from disk or HPSS).  Note that one file is created for the get_extrn_ics task and another one for the get_extrn_lbcs task.  These two files are actually shell scripts containing variable definitions that get sourced later by the make_ics and make_lbcs tasks, respectively.  This change removes the need to call the get_extrn_mdl_file_dir_info again as part of the make_ics and make_lbcs tasks.
* Modify the way the temporary fix for the suite_FV3_GFS_2017_gfdlmp_regional physics suite works so that the suite definition file (SDF) is not changed in the ufs_weather_model directory where the clone of the ufs-weather-model repo is located (so that no changes to tracked files are reported in the clone of that repo) and instead, the SDF is first copied to the experiment directory and then changed there.

Tests Conducted:
---------------
* All WE2E tests pass except regional_013.  regional_013 doesn't pass because the physics suite used by that test (FV3_GSD_SAR_v1) doesn't actually exist in the ufs-weather-model repo.
* The tests nco_conus_c96 and regional_009 (both are in NCO mode) cannot be run simultaneously.  The user must wait for one of these tests to finish before running the other.  This is because in NCO mode, directories generated by the workflow for a given experiment are not unique to that experiment and may be written to by another experiment running in NCO mode.  This must be fixed (see To-dos section below).  

To-dos:
------
* Fix the way the directories generated by the workflow for a given experiment are set so that they are unique to the experiment.  This is to prevent two NCO-mode experiments running simultaneously from interfering with (e.g. writing to) each other's directories.  There is now a new PR for this: #222 

Modifications common to more than one file (used below in listing of file-by-file modifications):
------------------------------------------------------------------------------------------------
(A) Rename local variables to be lower case.
(B) Fix/add/delete comments and/or informational and/or error messages.
(C) Remove theia stanza in machine-dependent case-statement.
[...]

PR #222, opened 2020-06-01
Fix directory overwrite bug in NCO mode.

gsketefian:
This needs to be merged after PR #221.

Summary of modifications:
------------------------
* Modify scripts to prevent simultaneously running experiments in NCO mode from writing to the same set of directories.

Tests conducted:
---------------
* All WE2E tests pass except regional_013. regional_013 doesn't pass because the physics suite used by that test (FV3_GSD_SAR_v1) doesn't actually exist in the ufs-weather-model repo.

Modifications common to more than one file (used below in listing of file-by-file modifications):
------------------------------------------------------------------------------------------------
(A) Fix/add/delete comments and/or informational and/or error messages.
(B) Remove unused code and/or fix code indentation.
(C) Set the RUN and envir variables to the name of the WE2E test name contained in EXPT_SUBDIR.  Note that even though EXPT_SUBDIR is set to a null string near the beginning of this file, the script that runs the WE2E tests inserts the name of the test into the definition of EXPT_SUBDIR (i.e. it changes the line 'EXPT_SUBDIR=""' into 'EXPT_SUBDIR="test_name"') before calling the experiment generation script.  Thus, EXPT_SUBDIR will be properly set to the name of the test when the experiment generation script is called.
(D) Bug fix - change variable name from LBC_UPDATE_INTVL_HRS to LBC_SPEC_INTVL_HRS.

File-by-file description of modifications:
-----------------------------------------

jobs/JREGIONAL_RUN_POST:
* In the definition of COMOUT, use the new workflow variable COMOUT_BASEDIR that specifies the base directory under which each cycle's UPP output directories will be placed.

tests/baseline_configs/config.nco_conus.sh: (A), (C)

tests/baseline_configs/config.nco_conus_c96.sh: (A), (C)

tests/baseline_configs/config.regional_009.sh: (A), (C)
* Change the time interval for specifying external model LBCs from 6 hrs to 3 hrs (to match what is done in all the other WE2E tests that use FV3GFS as the external model for LBCs).

tests/run_experiments.sh: (A), (B)

ush/config.community.sh: (D)

ush/config.nco.sh: (D)

ush/config_defaults.sh: (A)

ush/setup.sh: (A)
* For simplicity, combine two blocks of code that create different directories into a single block (that creates both sets of directories).
* In NCO mode, change defintion of CYCLE_BASEDIR from "$STMP/tmpnwprd/${EMC_GRID_NAME}" to "$STMP/tmpnwprd/$RUN".  This is to make it easier to associate this directory with an experiment (wich can be done by setting RUN to the experiment name in the experiment's configuration file).  That way, two experiments running simultaneously (but having different names) will not end up writing to the same CYCLE_BASEDIR (which currently can happen if the two experiments are using the same grid, i.e. the same value of EMC_GRID_NAME).
* Add COMOUT_BASEDIR as a new secondary workflow variable.  In NCO mode, set this to the base directory under which the output from the RUN_POST_TN task will be placed (i.e. the cycle-independent portion of the RUN_POST_TN task's output directory) and then check whether this directory exists (and deal with it as specified by the variable PREEXISTING_DIR_METHOD).  The goal here is to be able to uniquely associate this directory to an experiment so that simultaneously running experiments do not write their UPP outputs into the same directory (and possibly overwrite each other's results).  This directory is given by COMOUT_BASEDIR="$COMROOT/$NET/$envir", so we can uniquely associate it with an experiment by setting envir to the name of the experiment (which can be done in the experiment's configuration file).  This is why in the WE2E configuration files listed above (config.nco_conus.sh, config.nco_conus_c96.sh, and config.regional_009.sh), we set envir to the name of the experiment.  In community mode, the new variable COMOUT_BASEDIR is not used, so in that case we set it to an empty string.


BenjaminBlake-NOAA:
@gsketefian Should I test this on WCOSS with my latest mods or will they not work?  They are based off of your feature/emc_fixes_with_python branch.

gsketefian:
@BenjaminBlake-NOAA Yes, please go ahead and test.  The problem that this fixes occurred when I was running two (or more) WE2E tests in NCO mode on the same grid.  If you run one experiment at a time, you may not have seen the problem at all.  This PR builds on top of #221, which in turn builds on top of #212.  

I'm planning on sending an email to @mkavulich to see if we can get #212, #221 (which @JeffBeck-NOAA successfully tested on jet and will hopefully approve soon), this PR, and yours for WCOSS merged first, and he can do his tests on cheyenne on the cumulative result instead of having to do it for each PR.

BenjaminBlake-NOAA:
Sounds good!  I will get started on this first thing tomorrow and can hopefully finish the testing relatively quickly.

BenjaminBlake-NOAA:
I completed a successful test on the WCOSS Cray and Dell.  I will approve this PR and once it is merged I will submit a PR for the WCOSS port changes.

mkavulich:
I'm running tests on Cheyenne now, but I'm curious: why is the variable rename from LBC_UPDATE_INTVL_HRS  to LBC_SPEC_INTVL_HRS listed as a "bug fix"?

gsketefian:
@mkavulich I had made the variable name change from LBC_UPDATE_INTVL_HRS to LBC_SPEC_INTVL_HRS in a previous PR (#221) just for clarity (because this variable represents the time interval in hours between the times at which the LBCs are specified (SPEC); nothing is being updated (UPDATE) at these times).  Then @JeffBeck-NOAA pointed out a couple of places where I forgot to make the name change, so that's what I'm doing in this PR and calling it a bug fix.

gsketefian:
@mkavulich Thanks for testing on Cheyenne.  Glad to hear it works!  

But now I'm having some difficulty with the PRs since we moved to the squash-and-merge approach.  That's why you currently see a long list of conflicts for this PR, i.e. because the history for PR #212 (and maybe #221) which this PR builds on no longer exists -- because several commits from PR #212 were removed and replaced by a single new (squash) commit.  So commits that should be recognized as common to both PRs are not, and we get lots of conflicts that have to be handled manually.

Do you have any suggestions on how to get around this?  Thx.

BenjaminBlake-NOAA:
@gsketefian  As we discussed earlier today, I will wait until this PR gets merged into community develop before submitting a PR for the WCOSS port changes.

gsketefian:
It seems (from reading on this issue online) that the way to deal with the problem of getting a whole bunch of new unexpected conflicts that a squash-and-merge causes is to do a rebase of a second PR that builds off an initial PR that was squashed and merged.  I haven't tried it yet.  Will do so soon and report back.

mkavulich:
@gsketefian This is true, for PRs that have a common commit history that is *not* in the main (community_develop) branch, the later PR will have to be rebased to the community_develop before merging. It looks like this branch also has the commit history of #221, so I think that should be merged first.

 If you have some time later in the day I can walk you through the process over hangouts or something, it's not difficult but it should be done with extreme caution to avoid losing work.

PR #226, opened 2020-06-05
Fix for contrib module handling.

christinaholtNOAA:
## DESCRIPTION OF CHANGES:.
All the module files contained in this repository have been updated for the change on Jet/Hera to remove the contrib module, and instead use the `module use -a` method for locating necessary contrib package modulefiles. This will be the ONLY way to use these modules after June 11, 2020, and is already in place to work now. As such, this is a **time sensitive** PR.

## TESTS CONDUCTED:.
Loaded a few different modules on front end node. This included modules that need to be loaded, and those "modules" that are actually shell scripts and must be sourced. Both methods worked as expected.

I did not run any workflow tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,8 @@ This step will also copy the executables to the `exec` directory and link the fi
 On some platforms this environment is already available with a few commands;For example, on Jet/Hera:
 ```
 cd ush
-module load contrib miniconda3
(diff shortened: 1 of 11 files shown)


mkavulich:
Given this is a time-sensitive PR, I'm running the Hera test suite now and will merge the PR when it completes.

PR #227, opened 2020-06-05
WCOSS port

BenjaminBlake-NOAA:
Here are the changes required for running the community workflow on WCOSS.  This PR is based on PR #222.  A few notes:

1. For the modulefiles, I created hardcoded versions of the make_ics, make_lbcs, run_fcst, and run_post tasks.  The modulefiles from the source code directories don't work.  Eventually we can be more elegant and have .local files like the ones for Hera and other systems but the .hardcoded versions work for now.

2. I had to make changes to exregional_make_ics.sh and exregional_make_lbcs.sh to run with the latest UFS_UTILS code but we decided that I should hold off on committing that since the UFS_UTILS code in the community workflow utilizes the version with Larissa's chgres/grib2 capabilities.

3. On the Cray, one needs to add a <shared> tag for jobs that use the dev_transfer queue.  On the Dell, one needs to add a <memory> tag for jobs that use the dev_transfer queue.  I didn't know how to set that up but perhaps an issue can be created and we can discuss how to best handle that in the future.  That can wait until after the merge is complete though.


gsketefian:
Hi Ben,
Should we close this PR?
Gerard

BenjaminBlake-NOAA:
Yes since you merged it as part of PR 228.  I'll close this one.

PR #251, opened 2020-07-10
Make the location of the make_orog fix files more flexible

mkavulich:
## DESCRIPTION OF CHANGES: 
Previously, in exregional_make_orog.sh, a local variable "topo_dir" was used as the location of the fix files for the make_orog task, and this variable was hard-coded as a path relative to the global "FIXgsm" variable. Now topo_dir is set as a global variable in setup.sh, just like FIXgsm and SFC_CLIMO_INPUT_DIR, to allow more flexibility for the location of these fixed input files.

In addition, all of these paths can now be directly specified by the user in config.sh, rather than needing to modify setup.sh every time you want to point to a new fix file location.

## TESTS CONDUCTED: 
Ran a standard end-to-end case on Cheyenne (GSD_HRRR25km, 2019090118), and also ran the same test where I pointed to fixed data that I had copied to different directories via the config.sh file. These tests performed as expected.

Also successfully ran test suites on Hera (GSD_RAP13km new_JPgrid regional_002 regional_005 regional_006 regional_007 regional_008 regional_015, the rest have current known issues) and Jet (). These were successful as expected: unless the user modified their config.sh file to include one of these new paths, the workflow's behavior should not change.

### DIFF
--- a/ush/config_defaults.sh
+++ b/ush/config_defaults.sh
@@ -900,15 +900,16 @@ SFC_CLIMO_FIELDS=( \
 # Set parameters associated with the fixed (i.e. static) files.  Definitions:
 #
 # FIXgsm:
-# The location on disk of the majority of the static input files.
+# System directory in which the majority of fixed (i.e. time-independent) 
+# files that are needed to run the FV3SAR model are located
 #
-# topo_dir:
-# The location on disk of the static input files used by the make_orog task.
-# Can be the same as FIXgsm.
+# TOPO_DIR:
+# The location on disk of the static input files used by the make_orog
+# task (orog.x and shave.x). Can be the same as FIXgsm.
 #
 # SFC_CLIMO_INPUT_DIR:
-# The location on disk of the static surface climatology input fields. These
-# files are only used if RUN_TASK_MAKE_SFC_CLIMO=FALSE
+# The location on disk of the static surface climatology input fields, used by 
+# sfc_climo_gen. These files are only used if RUN_TASK_MAKE_SFC_CLIMO=TRUE
 #
 # FNGLAC, ..., FNMSKH:
 # Names of (some of the) global data files that are assumed to exist in 
@@ -961,7 +962,7 @@ SFC_CLIMO_FIELDS=( \
 # to a null string which will then be overwritten in setup.sh unless the
 # user has specified a different value in config.sh
 FIXgsm=""
-topo_dir=""
+TOPO_DIR=""
 SFC_CLIMO_INPUT_DIR=""
 
 FNGLAC="global_glacier.2x2.grb"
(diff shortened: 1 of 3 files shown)


mkavulich:
Still waiting for a second review in order to merge this PR...

ush/config_defaults.sh
```diff
@@ -899,6 +899,17 @@ SFC_CLIMO_FIELDS=( \
 #
 # Set parameters associated with the fixed (i.e. static) files.  Definitions:
 #
+# FIXgsm:
+# The location on disk of the majority of the static input files.
+#
+# topo_dir:
```
gsketefian: Can you make topo_dir uppercase since that's the convention we're using for workflow (global) variables?

ush/config_defaults.sh
```diff
@@ -899,6 +899,17 @@ SFC_CLIMO_FIELDS=( \
 #
 # Set parameters associated with the fixed (i.e. static) files.  Definitions:
 #
+# FIXgsm:
+# The location on disk of the majority of the static input files.
```
gsketefian: Are these all global files? (I'd guess so because of the "g" in FIXgsm, but just wondering...)

ush/config_defaults.sh
```diff
@@ -899,6 +899,17 @@ SFC_CLIMO_FIELDS=( \
...
+#
+# topo_dir:
+# The location on disk of the static input files used by the make_orog task.
+# Can be the same as FIXgsm.
+#
+# SFC_CLIMO_INPUT_DIR:
+# The location on disk of the static surface climatology input fields. These
+# files are only used if RUN_TASK_MAKE_SFC_CLIMO=FALSE
```
gsketefian: You mean this is used when RUN_TASK_MAKE_SFC_CLIMO is set to TRUE.  This directory contains the files that are the input to the sfc_climo_gen code that regrids that input to the native FV3SAR grid.

ush/setup.sh
```diff
@@ -608,6 +608,10 @@ HH_FIRST_CYCL=${CYCL_HRS[0]}
 # System directory in which the fixed (i.e. time-independent) files that
 # are needed to run the FV3SAR model are located.
 #
+# topo_dir:
```
gsketefian: I'd say define topo_dir (TOPO_DIR) and SFC_CLIMO_INPUT_DIR should be defined just once in config_defaults.sh, so move these definitions to there.  Just state here that you're defining those system directories that have been left unspecified by the user.

ush/config_defaults.sh
mkavulich: I do not know, I assume the static input files are all global since they shouldn't be domain-specific, and most of the filenames start with "global". But I am not certain.

ush/config_defaults.sh
mkavulich: Thanks, I misunderstood exactly what those input files were used for. I updated the comment accordingly

ush/setup.sh
mkavulich: Done

ush/config_defaults.sh
mkavulich: Done

PR #256, opened 2020-07-24
Move creation of model_config file from the experiment generation step to the forecast task.

gsketefian:
## Description of changes:

In PR #245, the creation of model_config files was moved out of the forecast task (in scripts/exregional_run_fcst.sh) and into the experiment generation step (in ush/generate_FV3SAR_wflow.sh), and a model_config file was created only in the cycle directory (i.e. not in each run directory) assuming that the parameters in model_config will not change for forecasts under a given cycle (e.g. for various ensemble members under a given cycle).  Although this is true for certain cases (e.g. cold starts), as pointed out by @christinaholtNOAA, this assumption does not hold in certain use cases (e.g. real-time runs) in which there may be multiple forecasts for a given cycle that need different values of the parameters in model_config (e.g. different forecast lengths, different computational resources, etc).  For this reason, a model_config file must be created for each specific forecast and placed in the run directory for that forecast.  Thus, in this PR, we move the creation of the model_config files back into the ush/exregional_run_fcst.sh script.

Note that this PR is needed before PR #253 can be merged into the develop branch.

## Tests conducted:
Ran WE2E tests regional_002 and community_ensemble_2mems to make sure changes do not break the non-ensemble (regional_002) and ensemble cases (community_ensemble_2mems).  These were successful.  Currently running all WE2E tests.

## File-by-file description of modifications:

jobs/JREGIONAL_RUN_FCST:
* Set the new cdate argument of the exregional_run_fcst() funciton to CDATE (which is set as an environment variable in the rocoto XML).

scripts/exregional_run_fcst.sh:
* Add new input argument cdate to this function.  This is needed in creating the model_config file for this forecast.
* Call the new function create_model_config_file() to create a model_config file in the current run directory.
* For ensemble forecasts, remove creation of link to a cycle-specific model_config file.  This file is now created on a per-forecast basis.
* Edit comments.

tests/baselines_list.txt:
* Order WE2E test names lexicographically.

ush/create_model_config_file.sh:
* New function to set the parameters in the model_config file for a given cdate and forecast run directory.

ush/create_model_config_files.sh:
* Remove this file since the work this used to do for all cycles is now done by the new function ush/create_model_config_file.sh for one cycle.  [Actually, this file got renamed to ush/create_model_config_file.sh (drop the "s" after file).]

ush/generate_FV3SAR_wflow.sh:
* Remove creation of model_config files for all cycles.  This is now done on a per-forecast basis in scripts/exregional_run_fcst.sh.


gsketefian:
@christinaholtNOAA Ok, just finished full suite of tests, and they all pass except regional_003, regional_004, and regional_010 (as expected).  If this looks good to you, I can merge it.

PR #260, opened 2020-08-06
Update RRFS domain parms to match ESRL realtime runs

TrevorAlcott-NOAA:
-Changed CONUS/AK domain names to remove references to any laboratory names
-Updated GFDLgrid parameters to match those used in ESRL's CONUS and Alaska real-time runs


christinaholtNOAA:
@gsketefian  Quick question: do any of the tests in the E2E suite need to be updated with this change?

ush/set_predef_grid_params.sh
```diff
@@ -497,9 +485,9 @@ predefined domain:
 
     DT_ATMOS="40"
 
-    LAYOUT_X="18"
-    LAYOUT_Y="12"
-    BLOCKSIZE="46"
+    LAYOUT_X="34"
```
gsketefian: @christinaholtNOAA Are these changes ok with you?

ush/set_predef_grid_params.sh
```diff
@@ -808,35 +796,28 @@ predefined domain:
...
-  
-    num_margin_cells_T6_bottom=249
-    GFDLgrid_JSTART_OF_RGNL_DOM_ON_T6G=$(( num_margin_cells_T6_bottom + 1 ))
-  
-    num_margin_cells_T6_top=249
-    GFDLgrid_JEND_OF_RGNL_DOM_ON_T6G=$(( GFDLgrid_RES - num_margin_cells_T6_top ))
+    GFDLgrid_STRETCH_FAC="1.161"
+    GFDLgrid_REFINE_RATIO="4"
+
+    GFDLgrid_ISTART_OF_RGNL_DOM_ON_T6G=205
```
gsketefian: For the GFDLgrid type of grids, I thought it was easier to visualize what's happening if the parameters we're changing here are the widths (in number of cells) of the left, right, bottom, and top margins; if one wants the grid to be symmetric about the center of the tile (and there's no reason not to make it symmetric), this approach makes it convenient to specify the same width for both margins.  Same goes for the RRFS_HRRR3km grid above.  But if you like just the final numbers, that's fine too.

ush/set_predef_grid_params.sh
```diff
@@ -808,35 +796,28 @@ predefined domain:
...
 
     GFDLgrid_USE_GFDLgrid_RES_IN_FILENAMES="FALSE"
 
-    DT_ATMOS="18"
+    DT_ATMOS="40"
 
-    LAYOUT_X="24"
-    LAYOUT_Y="24"
-    BLOCKSIZE="15"
+    LAYOUT_X="18"
```
gsketefian, line 833: Have these been at least somewhat optimized?

ush/set_predef_grid_params.sh
TrevorAlcott-NOAA: I was just trying to keep it simple.  Doesn't matter to me

ush/set_predef_grid_params.sh
TrevorAlcott-NOAA: @christinaholtNOAA these were chosen based on Sam T's testing, correct?  I just grabbed them from the current real-time runs, which we know are working well

ush/valid_param_vals.sh
```diff
@@ -12,12 +12,12 @@ valid_vals_PREDEF_GRID_NAME=( \
 "GSD_HAFSV0.A3km" \
 "GSD_HAFSV0.A13km" \
 "GSD_HAFSV0.A25km" \
-"GSD_RRFSAK_3km" \
-"GSD_HRRR_AK_50km" \
-"GSD_HRRR3km" \
-"GSD_HRRR13km" \
-"GSD_HRRR25km" \
-"GSD_RAP13km" )
+"RRFS_HRRRAK3km" \
```
gsketefian, line 15: Some of these changes need to also be made in the WE2E configuration files in `regional_workflow/tests/baseline_configs`.  Possibly also elsewhere, I don't know if you grepped for these to see all the places they appear.  For example, this is what I do from the `regional_workflow` directory for "GSD_HRRR_AK_50km":
`grep -n -i -r "GSD_HRRR_AK_50km"
`To skip searching certain file patterns and directories that I know aren't relevant, I add these options:
`grep -n -i -r "GSD_HRRR_AK_50km" --exclude=\*{old,orig}\* --exclude-dir=\*{sorc}\* *`

gsketefian:
@christinaholtNOAA Yes, thanks for pointing that out.  I put in a comment for @TrevorAlcott-NOAA for this.

ush/set_predef_grid_params.sh
christinaholtNOAA: @TrevorAlcott-NOAA I thought these original numbers (18, 12, 46) are the ones currently running in real time. I don't know that we want to change them. I did not change this file in the real time runs. I applied these changes only into the real-time config files, not the ush. I made a PR here to update these numbers before I ever changed real time runs.

ush/valid_param_vals.sh
TrevorAlcott-NOAA: OK I can apply these changes to the "tests" directory too.  It's starting to look like we're carrying around a lot of testing baggage in the repo.  What is the purpose of these directories, and do we need to keep the references to them? They obviously cannot be referenced when building on Jet.  /scratch2/BMC/det/FV3SAR_pregen/grid/GSD_HRRR25km  /scratch2/BMC/det/Gerard.Ketefian/UFS_CAM/test_latest_20191002/expt_dirs/test_sheleg_GSD_HRRR3km_01 /scratch2/BMC/det/Gerard.Ketefian/UFS_CAM/FV3SAR_demo_PASPBL_20190725/expt_dirs/test_GSD_HRRR25km

ush/set_predef_grid_params.sh
christinaholtNOAA: I feel that it's nice to see these rules explicitly here.

ush/set_predef_grid_params.sh
christinaholtNOAA: That's right. These numbers are consistent with what is currently running.

ush/valid_param_vals.sh
gsketefian: I think you're just looking at my NCL plotting scripts in regional_workflow/ush/NCL where I've tried plotting output in various locations.  I will be removing those NCL scripts at some point before the release, so don't worry about them.  Filter that directory out in your grep like this:
`grep -n -i -r "GSD_HRRR_AK_50km" --exclude=\*{old,orig}\* --exclude-dir=\*{sorc,NCL}\* *
`There may be more directories you'll have to filter out.  Let me know what you find and I can see if they're relevant or not.

TrevorAlcott-NOAA: Thanks @gsketefian , using your syntax the only affected files are the tests/config_xxx scripts.  I have updated those files and added them to this PR

ush/set_predef_grid_params.sh
TrevorAlcott-NOAA: This is in the JPgrid section.  We aren't running a JPgrid in real time, so I'm not sure we even know what these values should be.

TrevorAlcott-NOAA: This is probably some version-itis with the real-time scripts because we never paid attention to this section.  I can revert, but chances are we're going to have to change the values again anyway

TrevorAlcott-NOAA: Latest push reverts to 18/12/46

ush/set_predef_grid_params.sh
TrevorAlcott-NOAA: Added these back for HRRR3km and HRRRAK3km grids

gsketefian:
@TrevorAlcott-NOAA Couple of other (easy) requests:

1) Can you rename the file `tests/baseline_configs/config.GSD_RAP13km.sh` to ...RRFS...?
2) Can you change the test name "GSD_RAP13km" in tests/baselines_list.txt to "RRFS..."?

A couple of more involved requests:
1) It would be nice to run the full suite of WE2E tests to make sure nothing broke.  It's not too complicated, and it's something you'll have to do sooner or later!  I'm happy to help you with it.  Once you launch them (it would be just on hera for now), the testing script will set up cron jobs for you, so you just wait 1-2 hours and then check on them.
2) None of the WE2E tests run on the 3km grids that you changed because those take a lot longer to complete.  I'm leaning towards running a couple of short (say 6-hour forecast length) tests with these two domains just to make sure they're still working.  We can also add these tests to the set of WE2E tests, although we wouldn't require people to run them for every PR.  Creating these new tests is just a matter of taking the existing WE2E tests on the coarser grids and changing the setting of the predefined grid to use.

What do you think?
Gerard

christinaholtNOAA:
Pro tip: Use Hera for WE2E tests. The changes for optimization run the forecasts faster.

gsketefian:
@christinaholtNOAA We can add the 3km grids in the WE2E tests if they're a lot faster now.  Also, I forgot to mention that we actually run the EMC 3km grid as part of the tests, and it doesn't take too long.

TrevorAlcott-NOAA:
The original goal here was to get the current AK domain parameters into the repo, and that has snowballed into a lot of requested changes.  I know it's important to get this right, but I also don't have time to test 16 different model configurations.  I don't even know what WE2E is.  Maybe someone more familiar with this whole situation can take over this PR?  I'm clearly out of my depth.

gsketefian:
Ok, I'll continue with this PR and combine it with the PR that I was going to create to (1) change the JPgrid_NX value of the GSD_HRRR3km grid from 1734 to 1728 (so that it's divisible by LAYOUT_X="18" and (2) enable the GSD_HRRR3km grid to be used in NCO mode.  (The GSD_HRRR3km grid will also be renamed RRFS_HRRR3km.)

JeffBeck-NOAA:
@gsketefian, do we now have these AK domains covered and can close this PR?

Issue #264, opened 2020-08-14
Bug fix for background map plotting in python script


dmwright526:
Python plotting script does not allow for contours and background map to be display accurately at the same time. Updating script to allow for both to be displayed.

jwolff-ncar:
Has this issue been sufficiently addressed at this point? Can this be closed?

JeffBeck-NOAA:
This specific issue has been resolved, but there are follow-up problems related to contours and NaN values within the write component output.  @dmwright526 is looking into that currently.  I plan to coordinate with @gsketefian to fix the 25-km CONUS write component output domain, which is to narrow and contains NaNs.

PR #265, opened 2020-08-14
Issue #264 Bug fix for background map and removal of yaml in script

dmwright526:
## DESCRIPTION OF CHANGES: 
- Change from background_img() to imshow() to remove need for a .json file to show a background map image.
- Remove the color palette conversion to allow for background image and contours to display.
- Remove yaml mentions as it is not in use.
- Comment out QPF plotting since it is not available the files I am testing with. Others can comment back in if this is a local error.
- Added information in header to point users to the guide for downloading files needed for Cartopy.

## TESTS CONDUCTED: 
Script tested on Jet using older simulations. Will need testing on other platforms with more recent runs.

## ISSUE (optional): 
Fixes issue mentioned in #264

### DIFF
--- a/ush/Python/plot_allvars.py
+++ b/ush/Python/plot_allvars.py
@@ -53,7 +53,6 @@ from scipy import ndimage
 from netCDF4 import Dataset
 import pyproj
 import argparse
-import yaml
 import cartopy
 
 #--------------Define some functions ------------------#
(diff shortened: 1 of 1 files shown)


JeffBeck-NOAA:
Hi @dmwright526, I still see the env.yaml file in this PR.  Was that supposed be removed?

PR #270, opened 2020-08-17
add FV3_RRFS_vebeta test

siwei-noaa:
## DESCRIPTION OF CHANGES:  
The SDF FV3_RRFS_v1beta has been merged into the ufs-weather-model. The regional workflow does not have capability/configurations for running this new SDF. Such capability was added and tested in this request. 

It should be noted that two topographic files are required to run FV3_RRFS_v1beta as "drag_suite" scheme is used. So far (Aug 17, 2020), no scripts for automatic creating the required files, and they are copied from a temporary directory. 

## TESTS CONDUCTED: 
1. The capability for running SDF FV3_RRFS_v1beta has been added in the regional workflow.
2. The FV3_RRFS_v1beta test was performed on NOAA Hera with GSD_HRRR 25km pre-defined CONUS domain.

## CONTRIBUTORS: 
Dr. Michael Toy (NOAA/GSL) provided the required topographic files for using "drag_suite" scheme.


mkavulich:
@gsketefian Would it be appropriate to add this suite to one of the existing end-to-end test suites so that it gets regularly tested?

JulieSchramm:
This suite or the FV3_GFS_v16beta needs to be the out-of-the-box case.

ush/set_predef_grid_params.sh
```diff
@@ -345,7 +345,7 @@ predefined domain:
 
     JPgrid_WIDE_HALO_WIDTH=6
 
-    DT_ATMOS="300"
+    DT_ATMOS="40"
```
mkavulich: We should not change the default timestep, this will change results for people unexpectedly. If you need a different timestep you can overwrite it in config.sh.

scripts/exregional_make_orog.sh
```diff
@@ -342,6 +342,9 @@ raw_orog_fn="${raw_orog_fn_prefix}.${fn_suffix_with_halo}"
 raw_orog_fp="${raw_dir}/${raw_orog_fn}"
 mv_vrfy "${raw_orog_fp_orig}" "${raw_orog_fp}"
 
+# copy two files for drag_suite to the orog directory (temporary)
```
mkavulich, line 345: If they are necessary for running the case, these files should be added to the standard static/fix files on each official platform. We should not point to personal directories that could change unexpectedly.

ush/set_predef_grid_params.sh
JeffBeck-NOAA: @mkavulich I suggested Siwei commit this change.  The default timestep is too long, and will fail for this pre-defined domain every time.  A dt_atmos of 40 is what we should be using as the default for all pre-defined domains, since they all use the same vertical coordinate structure, and that is currently the limiting factor for stability.

JeffBeck-NOAA:
> The file config.sh should not be committed to the repository; please remove it from version control in your branch.

@siwei-noaa When you issue a "git add" for files in your local branch, be careful of "git add -A", as that will add everything that has changed, and may not be what you want to commit.  The config.sh file is one example, since it's a user-defined file used for specific experiment configuration.

ush/set_predef_grid_params.sh
mkavulich: If that's the case then the change should definitely be noted in the PR message and broadcast more widely than just in this PR; it will change results for people who do not specify DT_ATMOS in their config.sh

scripts/exregional_make_orog.sh
siwei-noaa: Thank you. We are waiting on Mike's script to replace those files, and will update this when it is ready.

siwei-noaa:
> The file config.sh should not be committed to the repository; please remove it from version control in your branch.

The file has been removed.

ush/set_predef_grid_params.sh
siwei-noaa: Thank you. I revoked the change on DT_ATMOS, and will be on a separate PR.

gsketefian:
> @gsketefian Would it be appropriate to add this suite to one of the existing end-to-end test suites so that it gets regularly tested?

Yes, whenever a new suite is added, a new WE2E test configuration should be added to test it regularly, especially an important one like this one  

@siwei-noaa Please create a new WE2E test for this scheme as follows.  Start with the file regional_workflow/tests/config.regional_003.sh
Copy it to a new file named regional_workflow/tests/config.FV3_RRFS_v1beta.sh and then in that new file, change
CCPP_PHYS_SUITE="FV3_GSD_v0"
to
CCPP_PHYS_SUITE="FV3_RRFS_v1"
Then test it (on hera) as follows:
a) cd regional_workflow/tests
b) echo "FV3_RRFS_v1beta" > my_expts.txt
c) Make sure your python environment is loaded:
   module use -a /contrib/miniconda3/modulefiles
   module load miniconda3
   conda activate regional_workflow
d) Run this new WE2E test:  
   ./run_experiments.sh expts_file="my_expts.txt" machine="hera" account="gsd-fv3" 
   Here, I'm assuming you usually run under the gsd-fv3 account; if not, change it to your usual account.
This should generate a new experiment under expt_dirs (in a subdirectory named FV3_RRFS_v1beta) and automatically relaunch the workflow until it succeeds or fails.

I'm happy to talk on the phone if these instructions are not clear.

scripts/exregional_run_fcst.sh
```diff
@@ -252,6 +252,26 @@ Cannot create symlink because target does not exist:
   target = \"$target\""
 fi
 
+# two files for drag_suite scheme
```
JeffBeck-NOAA, line 255: Hi @siwei-noaa, @gsketefian and I looked at your PR a bit more this morning, and we noticed that this block of code needs to be wrapped in an if statement for the RRFS_v1beta.  We don't want to do these linking steps unless that SDF is used.  So an if [ "${CCPP_PHYS_SUITE}" = "FV3_RRFS_v1beta" ] block should be added for these lines.

ush/link_fix.sh
```diff
@@ -244,6 +244,8 @@ Creating links in the FIXsar directory to the grid files..."
     fns=( \
 "C*${DOT_OR_USCORE}oro_data.tile${TILE_RGNL}.halo${NH0}.nc" \
```
JeffBeck-NOAA: @siwei-noaa Same comment as above for these lines.  An "if [ "${CCPP_PHYS_SUITE}" = "FV3_RRFS_v1beta" ]" statement should be added for the oro_data_??.tile* files in this block.  After the initial "if" statement, an "else" clause would have the original oro_data.tile* files, so that those would be defined for anything other than FV3_RRFS_v1beta.

scripts/exregional_run_fcst.sh
siwei-noaa: In testing, will commit if it works.

ush/link_fix.sh
siwei-noaa: In testing, will commit if it works.

scripts/exregional_run_fcst.sh
```diff
@@ -252,6 +252,28 @@ Cannot create symlink because target does not exist:
...
+if [ "${CCPP_PHYS_SUITE}" = "FV3_RRFS_v1beta" ]; then
```
JeffBeck-NOAA, line 256: @siwei-noaa, one last bug in this block of code.  All ${FIXsar} variables are now ${FIXLAM} in the community workflow.  Once that change is in, I think we'll be ready to merge.

PR #278, opened 2020-08-24
RRFS grid updates

gsketefian:
This PR is meant to replace PR #260.

## DESCRIPTION OF CHANGES: 
* Change predefined grid names of interest from "GSD_..." to "RRFS_...".
* Rename WE2E test configuration files so that they correspond to the changes in the predefined grid names.
* Update the parameters for the RRFS_HRRR3km grid of GFDLgrid type and the RRFS_HRRRAK3km grid of GFDLgrid type to those specified by Trevor's PR #260 (which this PR will replace).
* Add WE2E test configurations for the RRFS_HRRR3km grid of GFDLgrid type and the RRFS_HRRRAK3km grid of GFDLgrid type.  Tests are named RRFS_HRRR3km_GFDLgrid and RRFS_HRRRAK3km_GFDLgrid.

### On Hera
* Ran the regional_002 and regional_003 tests to ensure that the change in grid names did not cause a bug.  Both passed.
* Ran the two new WE2E tests (RRFS_HRRR3km_GFDLgrid and RRFS_HRRRAK3km_GFDLgrid).  Both failed in the run_fcst task.  The error is that during the adi (alternating-direction implicit?) procedure to solve for delz's, NaN values are returned for delz.
  * Ran the RRFS_HRRRAK3km_GFDLgrid test with both DT_ATMOS = 40 sec and DT_ATMOS = 18 sec to be sure that the problem is not a CFL instability; both cases failed, so it's probably not a CFL issue.
  * Ran the RRFS_HRRR3km_GFDLgrid test only with DT_ATMOS = 40 sec (which failed).
* Have not yet run the remaining (original) WE2E tests.

## CONTRIBUTORS (optional): 
@TrevorAlcott-NOAA


gsketefian:
@TrevorAlcott-NOAA @chunhuazhou @christinaholtNOAA 
1) Can you check to make sure that the grid parameter values are the ones you want?  I got them from @TrevorAlcott-NOAA 's PR #260.
2) The run_fcst task for both grids is currently failing.  Do any of you have forecasts with these grids (maybe with slightly different grid parameters) that are successful?  I realize we are eventually going to move away from these GFDLgrid type grids, but as long as we allow them in the workflow, we should have tests for them that work.

tests/baseline_configs/config.nco_RRFS_HRRR3km_HRRRX_RAPX.sh
```diff
@@ -26,7 +26,7 @@ VERBOSE="TRUE"
 RUN_ENVIR="nco"
 PREEXISTING_DIR_METHOD="rename"
 
-EMC_GRID_NAME="GSD_HRRR3km"
+EMC_GRID_NAME="RRFS_HRRR3km"
```
christinaholtNOAA, line 29: Why does this variable have a Center name in it? Will it not work to use the same variable for all grid names?

tests/baselines_list.txt
```diff
@@ -1,8 +1,8 @@
-GSD_RAP13km
+RRFS_RAP13km
```
christinaholtNOAA, line 1: Is it still important to test this configuration?

tests/baseline_configs/config.nco_RRFS_HRRR3km_HRRRX_RAPX.sh
TrevorAlcott-NOAA: I'm wondering the same thing.  It seems like the use of the "EMC_GRID_NAME" variable restricts us to a smaller subset of domains with nco mode.  This was a problem when trying to use nco mode with the GSD Alaska domain (soon to be RRFS_HRRRAK3km).  Is it possible to remove that limitation, either by ending use of "EMC_GRID_NAME" or some other means?

gsketefian: Yes, the variable EMC_GRID_NAME needs to be removed, and instead the variable PREDEF_GRID_NAME should be used.  I reluctantly added EMC_GRID_NAME a while ago because EMC was using a different grid naming convention than we were (with some grids being the same as ours but with a different name), and I wanted to use a different variable to store their grid names.  

I'd like to talk to them before making this change to see if they're ok with having a single set of grid names.  We (GSL and EMC) have to come to an agreement on a grid naming convention.  For example, they have a 3km CONUS grid named just "conus" while ours is called "RRFS_HRRR3km".  Do we just keep both?  Or do they move to using ours, which is presumably the official one since it's got "RRFS" in its name?  Etc...  If there are any EMC people on the 11am workflow telecon today, I will bring it up.

Btw, how did you (the GSL real-time group) decide on the grid names (RRFS_HRRR3km, RRFS_HRRRAK3km, etc)?  Was there any consultation with EMC?

tests/baselines_list.txt
gsketefian: I have not heard otherwise, but I'm not always up-to-date.  It is a nice test of the application of an ESG grid to a large domain, which is when using an ESG grid really matters.

JeffBeck-NOAA:
@gsketefian, Can we close this PR?

PR #280, opened 2020-08-27
WE2E test configuration for FV3_RRFS_v1beta

siwei-noaa:
## DESCRIPTION OF CHANGES:  
Add WE2E test configuration for FV3_RRFS_v1beta.

## TESTS CONDUCTED: 
The test has been successfully conducted on Hera.


scripts/exregional_make_orog.sh
```diff
@@ -342,6 +342,9 @@ raw_orog_fn="${raw_orog_fn_prefix}.${fn_suffix_with_halo}"
 raw_orog_fp="${raw_dir}/${raw_orog_fn}"
 mv_vrfy "${raw_orog_fp_orig}" "${raw_orog_fp}"
 
+# copy two files for drag_suite to the orog directory (temporary)
```
gsketefian, line 345: Need to wrap this inside an if-statement to prevent failure with other physics suites and machines.  Like this:

```
#
#
# Copy the two orography files needed for the drag suite in the FV3_RRFS_v1beta
# physics suite.
# 
# Note that the following is a temporary fix.  We need a long-term solution 
# that calls a script or program to generates the necessary files (instead 
# of copying them).
#
#
if [ "${CCPP_PHYS_SUITE}" = "FV3_RRFS_v1beta" ]; then
  if [ "$MACHINE" = "HERA" ]; then
    cp_vrfy /scratch2/BMC/gsd-fv3-test/she/orog/* ${OROG_DIR}
  else
    print_err_msg_exit "\
The location from which to copy the files needed by the gravity wave drag
parameterization in the current physics suite (CCPP_PHYS_SUITE) has not
been specified for this machine (MACHINE):
  CCPP_PHYS_SUITE = \"${CCPP_PHYS_SUITE}\"
  MACHINE = \"${MACHINE}\""
  fi
fi

```

scripts/exregional_run_fcst.sh
```diff
@@ -252,6 +252,28 @@ Cannot create symlink because target does not exist:
   target = \"$target\""
 fi
 
+# two files for drag_suite scheme
```
gsketefian, line 255: For clarity, please say in the comment that this is only for the drag parameterization in the FV3_RRFS_v1beta physics suite.

ush/templates/model_configure.FV3_RRFS_v1beta
```diff
@@ -0,0 +1,23 @@
+total_member:            1
```
gsketefian: I think in a previous PR we forgot to remove all the model_configure.${CCPP_PHYS_SUITE} files in the regional_workflow/ush/templates directory.  I don't think we need them; we just need the template file named model_configure (the workflow creates an actual model_configure file for an experiment using this).  Can you try removing this file and rerunning?  If things are ok, we'll also have to do a PR to remove the other model_configure.${CCPP_PHYS_SUITE} files.

gsketefian:
As a temporary measure to stop the workflow generation if the GSD_HRRR3km grid is not specified with the FV3_RRFS_v1beta suite or if it's not being run on hera, I'd put this bit of code somewhere in generate_FV3LAM_wflow.sh:

```
if [ "${CCPP_PHYS_SUITE}" = "FV3_RRFS_v1beta" ]; then
  if [ "${PREDEF_GRID_NAME}" != "GSD_HRRR3km" ] || [ "$MACHINE" != "HERA" ]; then
    print_err_msg_exit "\
The FV3_RRFS_v1beta currently cannot be used with this machine (MACHINE) and 
predefined grid (PREDEF_GRID_NAME) combination:
  CCPP_PHYS_SUITE = \"${CCPP_PHYS_SUITE}\"
  MACHINE = \"${MACHINE}\""
  PREDEF_GRID_NAME = \"${PREDEF_GRID_NAME}\""
  fi
fi

```
If we want to run this suite on other machines, we can copy Mike's files over to those and take out the machine dependency in this if-statement (and the one in exregional_make_orog.sh).

scripts/exregional_make_orog.sh
gsketefian: To avoid the hard-coded directory in the above snippet, we can add this temporary section to the end of config_defaults.sh:
```
#
#
# GWD_RRFS_v1beta_DIR:
# Temporary workflow variable specifies the directory in which to look 
# for certain fixed orography files needed only by the gravity wave drag 
# parameterization in the FV3_RRFS_v1beta physics suite.  This variable
# is added in order to avoid including hard-coded paths in the workflow
# scripts.  Currently, the workflow simply copies the necessary files 
# from this directory to the orography directory (OROG_DIR) under the 
# experiment directory.  
#
# Note that this variable is only used when using the FV3_RRFS_v1beta 
# physics suite.  Also, as of 20200828, versions of these orography files 
# available only for the GSD_HRRR3km grid.  This variable should be 
# removed from the workflow once there is a script that generates these 
# files for any grid.
#
#
GWD_RRFS_v1beta_DIR=""

```
Then change the above to use this new variable:
```
#
#
# Copy the two orography files needed for the drag suite in the FV3_RRFS_v1beta
# physics suite.
# 
# Note that the following is a temporary fix.  It works only for the 
# GSD_HRRR3km predefined grid and only on hera.  We need a long-term 
# solution that calls a script or program to generates the necessary 
# "ls" and "ss" orography files (instead of copying them).
#
#
if [ "${CCPP_PHYS_SUITE}" = "FV3_RRFS_v1beta" ]; then
  if [ "${PREDEF_GRID_NAME}" = "GSD_HRRR3km" ] && [ "$MACHINE" = "HERA" ]; then
    cp_vrfy ${GWD_RRFS_v1beta_DIR}/* ${OROG_DIR}
  else
    print_err_msg_exit "\
The location from which to copy the files needed by the gravity wave drag
parameterization in the current physics suite (CCPP_PHYS_SUITE) has not
been specified for this machine (MACHINE) and predefined grid (PREDEF_GRID_NAME) 
combination:
  CCPP_PHYS_SUITE = \"${CCPP_PHYS_SUITE}\"
  MACHINE = \"${MACHINE}\""
  PREDEF_GRID_NAME = \"${PREDEF_GRID_NAME}\""
  fi
fi
```
Finally, set this in your config.sh:
`GWD_RRFS_v1beta_DIR="/scratch2/BMC/gsd-fv3-test/she/orog"
`
This way we can avoid hard-coded directories but still allow GSL to test this physic suite on the GSD_HRRR3km grid.

gsketefian:
@siwei-noaa Given the new info about which orog files are available from Mike, this is what I would do:

Add this to the end of config_defaults.sh:
```
#
#
# GWD_RRFS_v1beta_BASEDIR:
# Temporary workflow variable specifies the base directory in which to 
# look for certain fixed orography files needed only by the gravity wave 
# drag parameterization in the FV3_RRFS_v1beta physics suite.  This variable
# is added in order to avoid including hard-coded paths in the workflow
# scripts.  Currently, the workflow simply copies the necessary files 
# from a subdirectory under this directory (named according to the specified
# predefined grid) to the orography directory (OROG_DIR) under the 
# experiment directory.  
#
# Note that this variable is only used when using the FV3_RRFS_v1beta 
# physics suite.  This variable should be removed from the workflow once 
# there is a script that generates these files for any grid.
#
#
GWD_RRFS_v1beta_BASEDIR=""
```

Add this to setup.sh, say around line 913, right before where EXPT_BASEDIR is set:
```
#
#
# If using the FV3_RRFS_v1beta physics suite, make sure that the directory
# from which certain fixed orography files will be copied to the experiment 
# directory actually exists.  Note that this is temporary code.  It should
# be removed once there is a script that will create these orography files
# for any grid.
#
#
GWD_RRFS_v1beta_DIR="${GWD_RRFS_v1beta_BASEDIR}/${PREDEF_GRID_NAME}"
if [ "${CCPP_PHYS_SUITE}" = "FV3_RRFS_v1beta" ]; then

  if [ -z "${PREDEF_GRID_NAME} ]; then

    print_err_msg_exit "\
A predefined grid name (PREDEF_GRID_NAME) must be specified when using 
the FV3_RRFS_v1beta physic suite:
  CCPP_PHYS_SUITE = \"${CCPP_PHYS_SUITE}\"
  PREDEF_GRID_NAME = \"${PREDEF_GRID_NAME}\""

  else

    if [ ! -d "${GWD_RRFS_v1beta_DIR}" ]; 
      print_err_msg_exit "\
The directory (GWD_RRFS_v1beta_DIR) that should contain the gravity wave 
drag-related orography files for the FV3_RRFS_v1beta does not exist:
  GWD_RRFS_v1beta_DIR = \"${GWD_RRFS_v1beta_DIR}\""
    elif [ ! "$( ls -A ${GWD_RRFS_v1beta_DIR} )" ]; then
      print_err_msg_exit "\
The directory (GWD_RRFS_v1beta_DIR) that should contain the gravity wave 
drag related orography files for the FV3_RRFS_v1beta is empty:
  GWD_RRFS_v1beta_DIR = \"${GWD_RRFS_v1beta_DIR}\""
    fi

  fi

fi
```
The above is just a bunch of checks to make GWD_RRFS_v1beta_DIR is properly set up.  Also, add this to setup.sh
`GWD_RRFS_v1beta_DIR="${GWD_RRFS_v1beta_DIR}"
`
right after the line 
`SFC_CLIMO_DIR="${SFC_CLIMO_DIR}"
`
This writes the value of GWD_RRFS_v1beta_DIR to the variable definitions file (var_defns.sh) so it is accessable by other scripts.
Finally, change the current line in exregional_make_orog.sh where the ls and ss files are copied to this:
```
#
#
# Copy the two orography files needed for the drag suite in the FV3_RRFS_v1beta
# physics suite.
# 
# Note that the following is a temporary fix.  We need a long-term solution
# that calls a script or program to generates the necessary "ls" and "ss" 
# orography files (instead of copying them) for any grid.
#
#
if [ "${CCPP_PHYS_SUITE}" = "FV3_RRFS_v1beta" ]; then
  cp_vrfy ${GWD_RRFS_v1beta_DIR}/* ${OROG_DIR}
fi

```
I haven't actually run this code, so it probably contains bugs, but you get the idea.  I removed all checks for machines so that it should run on any machine as long as the files are copied onto that machine.  Of course, the user must specify GWD_RRFS_v1beta_BASEDIR in config.sh, and the files must be located in ${GWD_RRFS_v1beta_BASEDIR}/${PREDEF_GRID_NAME}.

scripts/exregional_run_fcst.sh
```diff
@@ -251,7 +251,31 @@ else
 Cannot create symlink because target does not exist:
   target = \"$target\""
 fi
-
+#
+# two files for drag_suite scheme
+# this is only for the drag parameterization in the FV3_RRFS_v1beta physics suite
+#
+if [ "${CCPP_PHYS_SUITE}" = "FV3_RRFS_v1beta" ]; then
+    # Symlink to orographic statistics fields file with "${CRES}_" and "halo0" stripped from name.
```
gsketefian: Can we use 2 spaces for indent since that's what's used everywhere else in the scripts?  Thx.

scripts/exregional_run_fcst.sh
```diff
@@ -251,7 +251,31 @@ else
...
+    target="${FIXLAM}/${CRES}${DOT_OR_USCORE}oro_data_ls.tile${TILE_RGNL}.halo${NH0}.nc"
+    symlink="oro_data_ls.nc"
+    if [ -f "${target}" ]; then
+      ln_vrfy -sf ${relative_or_null} $target $symlink
+    else
+      print_err_msg_exit "\
+    Cannot create symlink because target does not exist:
```
gsketefian: Lines 266 and 267: No need to indent because the indentation spaces will be printed out in the error message as well (unless that's what you want).  Same for lines 275-276 below.

gsketefian:
@siwei-noaa The conflicts are showing up I suspect because you started out this work with your previous branch for PR#[270](https://github.com/NOAA-EMC/regional_workflow/pull/270) (feature/rrfs_v1beta_test).  When that PR was merged, all the commits in that branch were squashed into a single new commit into develop (squash-and-merge), the commits in that branch are not in the develop branch (and vice versa).  That's why when git compares the develop and the current branch, it doesn't see the same commit hash history and thinks they're completely different sets of changes (to the same files).  That results in conflicts.  To avoid this situation, start each PR by first updating your develop branch (by pulling from the NOAA-EMC/regional_workflow repo) and then checking out from that a new feature branch in which you'll do your work.  (There is an alternative way around this problem that involves rebasing commits, but it's complicated and may be dangerous, so we'll avoid it for now.)  

In this case, what you can do is update your develop branch, then try to merge it into your feature/rrfs_v1beta_test branch.  You'll get conflicts which you'll have to resolve manually.  I use "git mergetool" with xxdiff to do that (on hera).  If you haven't done that before and would like help, let me know.

JeffBeck-NOAA (approved):

PR #292, opened 2020-09-11
Add halo blending option.

JeffBeck-NOAA:
DESCRIPTION OF CHANGES:
Adds halo blending, including a user-defined HALO_BLEND variable in config_defaults, "nrows_blend" value in input.nml.FV3, and applies the HALO_BLEND user-defined value during the generate step.

TESTS CONDUCTED:
Tested on Hera.


ush/config_defaults.sh
```diff
@@ -1219,6 +1219,16 @@ SKEB_VDOF="10"
 USE_ZMTNBLCK="false"
 #
 #-----------------------------------------------------------------------
+#
```
gsketefian, line 1222: What to do if you don't want blending?  Set to HALO_BLEND to 0?

JeffBeck-NOAA:
Yes, that will effectively shut it off.  We should probably add a comment
to that effect in the config_defaults.sh file.

- Jeff

JeffBeck-NOAA:
Just pushed a commit adding a comment about shutting off halo blending by setting HALO_BLEND to zero.

ush/config_defaults.sh
JeffBeck-NOAA: See comments below.

gsketefian:
@JeffBeck-NOAA Can you update your PR#[17](https://github.com/ufs-community/ufs-srweather-app/pull/17) into ufs-srweather-app so that it uses the release/public-v2 branch?  The current ufs-weather-model does not have the RRFS_v1beta suite defined yet (SDF is missing).

gsketefian:
> @JeffBeck-NOAA Can you update your PR#[17](https://github.com/ufs-community/ufs-srweather-app/pull/17) into ufs-srweather-app so that it uses the release/public-v2 branch? The current ufs-weather-model does not have the RRFS_v1beta suite defined yet (SDF is missing).

Oops, this is in the wrong PR!!

PR #303, opened 2020-10-01
Wrapped global_equiv_resol exec_fp with APRUN just as other commands …

ywangwof:
Wrapped global_equiv_resol exec_fp with APRUN just as other commands do and inserted CASE statements for Stampede.

- in scripts/exregional_make_grid.sh, added "$APRUN" before "${exec_fp}" for "global_equiv_resol". This is to make sure the command "global_equiv_resol" will be executed on compute nodes just as other commands do. I am not sure why "$APRUN" is missed for "global_equiv_resol" only. Otherwise, the command will fail on Odin.

- Added platform support for "STAMPEDE" in some CASE statements in _ush/get_extrn_mdl_file_dir_info.sh_ & _ush/set_extrn_mdl_params.sh_.

### DIFF
--- a/scripts/exregional_make_grid.sh
+++ b/scripts/exregional_make_grid.sh
@@ -429,7 +429,7 @@ cubed-sphere grid equivalent resolution does not exist:
 Please ensure that you've built this executable."
 fi
 
-${exec_fp} "${grid_fp}" || \
+$APRUN ${exec_fp} "${grid_fp}" || \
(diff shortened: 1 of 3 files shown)


gsketefian (approved):
Looks good.  Unless anyone else has objections, feel free to squash-and-merge.  Thanks!

JeffBeck-NOAA (approved):

Issue #281, opened 2020-08-28
Remove model_config.${CCPP_PHYS_SUITE} files from ush/templates


gsketefian:
The physics-suite-specific files model_config.${CCPP_PHYS_SUTE} in ush/templates are not needed because the workflow generates a model_config file for a given experiment from the model_config template file in ush/templates (which should NOT be deleted!).

Currently, the model_config template file is not a jinja template.  Modify this file (along with ush/create_model_config_file.sh) to use jinja.

christinaholtNOAA:
Should this also move to a Jinja template at the same time?

gsketefian:
@christinaholtNOAA Let me know if the new issue title/description needs any more mods.  Thx.

jwolff-ncar:
Status update? Is this required prior to UFS SRW v1 release?

gsketefian:
This was resolved by PR #321.

PR #321, opened 2020-10-16
Remove redundant model_configure.${CCPP_PHYS_SUITE} template files; use Jinja2 to create model_configure

gsketefian:
## DESCRIPTION OF CHANGES:
* Remove model_configure template files whose names depend on the physics suite, i.e. files with names of the form model_configure.${CCPP_PHYS_SUITE}.  Only a single template file is needed because the contents of the model_configure file are not suite dependent.  This leaves just one template file (named model_configure).
* Change the function create_model_configure_file.sh and the template file model_configure so they use jinja2 instead of sed to replace placeholder values.
* Absorb the contents of the write-component template files wrtcmp_lambert_conformal, wrtcmp_regional_latlon, and wrtcmp_rotated_latlon into the new jinja2-compliant model_configure file.  We can do this because Jinja2 allows use of if-statements in the template file.
* In the new model_configure jinja2 template file, include comments to explain the various write-component parameters.

## TESTS CONDUCTED:
On Hera, ran the two WE2E tests new_ESGgrid and new_GFDLgrid.  The first uses a "lambert_conformal" type of write-component grid, and the second uses a "rotated_latlon" type of write-component grid.  (The write-component also allows "regional_latlon" type grids, which is just the usual earth-relative latlon coordinate system, but we do not have any cases that use that.)  Both tests succeeded.

## ISSUE (optional): 
This PR resolves issue #281.


dmwright526 (approved):
Just did a quick check on Jet. Tests new_ESGgrid and new_GFDLgrid worked for me as well.

gsketefian:
@dmwright526 Thanks for testing!  I'm going to also ask @christinaholtNOAA to take a look to make sure my jinja coding is in good order since she's the jinja guru!

@christinaholtNOAA Can you please take a look at this PR?  I am not able to officially add you as a reviewer to PRs since the repo was accidentally deleted.  @mkavulich Can you add Christina back to the list of reviewers (I think you did that last time).  Thx.

ush/create_model_configure_file.sh
```diff
@@ -106,65 +99,119 @@ run directory (run_dir):
...
-      set_file_param "${model_config_fp}" "nx" "${WRTCMP_nx}"
-      set_file_param "${model_config_fp}" "ny" "${WRTCMP_ny}"
-      set_file_param "${model_config_fp}" "dx" "${WRTCMP_dx}"
-      set_file_param "${model_config_fp}" "dy" "${WRTCMP_dy}"
     elif [ "${WRTCMP_output_grid}" = "regional_latlon" ]; then
-      set_file_param "${model_config_fp}" "lon2" "${WRTCMP_lon_upr_rght}"
-      set_file_param "${model_config_fp}" "lat2" "${WRTCMP_lat_upr_rght}"
-      set_file_param "${model_config_fp}" "dlon" "${WRTCMP_dlon}"
-      set_file_param "${model_config_fp}" "dlat" "${WRTCMP_dlat}"
+
```
christinaholtNOAA: I'm not seeing how these two clauses differ. Maybe I'm just missing it? The same applies for the if statements in the XML template.

christinaholtNOAA (commented):
I think the templating looks fine with just a quick review. 

I do think that a Jinja template for this file limits you very greatly, though. You cannot change any field that is not already templated. The file is a yaml file and can be treated as such, building it the same way you'd build a namelist, but with yaml output instead of a Fortran namelist.

ush/create_model_configure_file.sh
gsketefian: No, good catch.  They used to be different but eventually converged to the same.  I will collapse them.  In the XML, the comments defining the grid parameters are different for regional_latlon vs. rotated_latlon.  I'll rearrange so the comments are done before the code.

gsketefian:
@christinaholtNOAA Ok, just pushed a commit to collapse the if-statment stanzas into one.  I reran the two WE2E tests and they both passed.

> I do think that a Jinja template for this file limits you very greatly, though. You cannot change any field that is not already templated. The file is a yaml file and can be treated as such, building it the same way you'd build a namelist, but with yaml output instead of a Fortran namelist.

Thanks for the suggestion.  Since this will take me some time to do and is not essential for the release, I will create an issue and get to it later.

JeffBeck-NOAA (approved):

gsketefian:
@christinaholtNOAA Ok, the issue to remind us to change from jinja to yaml is #331.  I will go ahead and merge this now.  Thanks for reviewing it!

PR #309, opened 2020-10-09
Port workflow to Orion

gsketefian:
IMPORTANT NOTE:
On Orion, in order to load the regional_workflow environment needed for generating an experiment, the user must first issue the following commands:
```
module use -a /apps/contrib/miniconda3-noaa-gsl/modulefiles
module load miniconda3
conda activate regional_workflow
```

## DESCRIPTION OF CHANGES:
* Add stanzas for Orion where necessary.
* Add new module files for Orion.
* On Orion, both the slurm partition and the slurm QOS need to be specified in the rocoto XML in order to be able to have wall times longer than 30 mins (the partition needs to be specified because it is by default "debug", which has a limit of 30 mins).  Thus, introduce modifications to more easily specify slurm partitions:
    * Remove the workflow variables QUEUE_DEFAULT_TAG, QUEUE_HPSS_TAG, and QUEUE_FCST_TAG that are currently used to determine whether QUEUE_DEFAULT, QUEUE_HPSS, and QUEUE_FCST specify the names of queue/QOS's or slurm partitions.
    * Add the workflow variables PARTITION_DEFAULT_TAG, PARTITION_HPSS_TAG, and PARTITION_FCST_TAG.  These will be used to specify slurm partitions only, and the variables QUEUE_DEFAULT, QUEUE_HPSS, and QUEUE_FCST will be used to specify queues/QOS's only.

## TESTS CONDUCTED:
Ran 11 WE2E tests on Orion, Hera, and Cheyenne.

Results on Orion:
```
community_ensemble_2mems          SUCCESS
DOT_OR_USCORE                     SUCCESS
grid_GSD_HRRR_AK_50km             FAILURE - In the run_fcst task.
  * Error message:
  !!! (1) Error in subr radiation_aerosols: unrealistic surface pressure =
           1                     NaN
new_ESGgrid                       SUCCESS
new_GFDLgrid                      SUCCESS
regional_001                      SUCCESS
regional_002                      SUCCESS
suite_FV3_GFS_v15p2               SUCCESS
suite_FV3_GFS_v16beta             SUCCESS
suite_FV3_GSD_SAR                 SUCCESS
suite_FV3_GSD_v0                  SUCCESS
```
Results on Hera:
```
community_ensemble_2mems          SUCCESS
DOT_OR_USCORE                     SUCCESS
grid_GSD_HRRR_AK_50km             SUCCESS
new_ESGgrid                       SUCCESS
new_GFDLgrid                      SUCCESS
regional_001                      SUCCESS
[...]

PR #343, opened 2020-10-28
Allow running from crontab on Cheyenne, as well as fix workflow generation on Cheyenne

### LINKED ISSUE #326: Load new Cheyenne rocoto module in launch_FV3LAM_wflow.sh

mkavulich:
## DESCRIPTION OF CHANGES: 
Changes in #339 broke the jinja templating script for creating the workflow xml on platforms that do not have "partition_default", "partition_hpss", or "partition_fcst" settings. Basically it boils down to platforms whose job scheduler does not support the rocoto "partition" tag. Cheyenne is one of these platforms.

Additionally, this PR specifies the location of the new rocoto module on cheyenne, so tests can now be launched from a cron job on Cheyenne.

## TESTS CONDUCTED: 
Ran a small set of end-to-end tests on Cheyenne (intel) successfully with no failures: DOT_OR_USCORE, grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2, suite_FV3_GFS_v15p2, and suite_FV3_RRFS_v1beta (test location: /glade/scratch/kavulich/UFS_CAM/testing/rocoto_on_cheyenne/expt_dirs)

Ran a full suite of tests on Hera to ensure the jinja changes did not negatively impact tests on that platform. All tests expected to pass passed.

## ISSUE (optional): 
Resolves #326

### DIFF
--- a/ush/fill_jinja_template.py
+++ b/ush/fill_jinja_template.py
@@ -259,7 +259,7 @@ def main(cla):
     tvars = {}
     for var in template_vars:
 
-        if cfg.get(var) is None:
+        if cfg.get(var, "NULL") == "NULL":
             raise KeyError(f'{var} does not exist in user-supplied settings!')
 
         if not cla.quiet:
--- a/ush/templates/FV3LAM_wflow.xml
+++ b/ush/templates/FV3LAM_wflow.xml
@@ -64,17 +64,17 @@ the "HPSS" type is used for the GET_EXTRN_ICS_TN and GET_EXTRN_LBCS_TN
 tasks; and the "FCST" type is used for the RUN_FCST_TN task.
 -->
 
-{%- if (partition_default != "")  %}
+{%- if partition_default is not none %}
 <!ENTITY RSRV_DEFAULT "<account>&ACCOUNT;</account><queue>&QUEUE_DEFAULT;</queue><partition>{{ partition_default }}</partition>">
 {%- else %}
 <!ENTITY RSRV_DEFAULT "<account>&ACCOUNT;</account><queue>&QUEUE_DEFAULT;</queue>">
 {%- endif %}
-{%- if (partition_hpss != "")  %}
+{%- if partition_hpss is not none %}
 <!ENTITY RSRV_HPSS    "<account>&ACCOUNT;</account><queue>&QUEUE_HPSS;</queue><partition>{{ partition_hpss }}</partition>">
 {%- else %}
 <!ENTITY RSRV_HPSS    "<account>&ACCOUNT;</account><queue>&QUEUE_HPSS;</queue>">
 {%- endif %}
-{%- if (partition_fcst != "")  %}
+{%- if partition_fcst is not none %}
 <!ENTITY RSRV_FCST    "<account>&ACCOUNT;</account><queue>&QUEUE_FCST;</queue><partition>{{ partition_fcst }}</partition>">
 {%- else %}
 <!ENTITY RSRV_FCST    "<account>&ACCOUNT;</account><queue>&QUEUE_FCST;</queue>">
--- a/ush/launch_FV3LAM_wflow.sh
+++ b/ush/launch_FV3LAM_wflow.sh
@@ -91,13 +91,14 @@ expt_name="${EXPT_SUBDIR}"
 #
 #-----------------------------------------------------------------------
 #
-if [ "$MACHINE" != "CHEYENNE" ]; then
-  if [ "$MACHINE" = "ORION" ]; then
-    module load contrib rocoto
-  else
-    module purge
-    module load rocoto
-  fi
+if [ "$MACHINE" = "CHEYENNE" ]; then
+  module use -a /glade/p/ral/jntp/UFS_SRW_app/modules/
+  module load rocoto
+elif [ "$MACHINE" = "ORION" ]; then
+  module load contrib rocoto
+else
+  module purge
+  module load rocoto
 fi
 #
 #-----------------------------------------------------------------------


gsketefian:
@mkavulich I tested PR #339 (really #309) on several platforms, including Cheyenne.  See #309 for the list of tests run on Cheyenne and the results.  Summary is that 10 of 11 tests passed.  The one that failed was an Alaska domain (which also failed on Orion but succeeded on Hera; indicates a bug in the weather model to me).   #309 ensures that the \<partition\> tag is not passed to the rocoto XML on those platforms for which PARTITION_DEFAULT, PARTITION_HPSS, and PARTITION_FCST are set to empty strings.  The default values (in config_defaults.sh) of these three parameters are empty strings, and in the case statement around line 400 of setup.sh, they do not get reset for Cheyenne, so the \<partition\> tag should not appear in the XML on Cheyenne.  If that's not the case, then there is a bug somewhere.

mkavulich:
@gsketefian I made a dumb mistake and forgot to push the actual fix to github. The PR has been updated with the real fix to the xml/python logic.

> @mkavulich I tested PR #339 (really #309) on several platforms, including Cheyenne. See #309 for the list of tests run on Cheyenne and the results. Summary is that 10 of 11 tests passed. The one that failed was an Alaska domain (which also failed on Orion but succeeded on Hera; indicates a bug in the weather model to me). #309 ensures that the <partition> tag is not passed to the rocoto XML on those platforms for which PARTITION_DEFAULT, PARTITION_HPSS, and PARTITION_FCST are set to empty strings. The default values (in config_defaults.sh) of these three parameters are empty strings, and in the case statement around line 400 of setup.sh, they do not get reset for Cheyenne, so the <partition> tag should not appear in the XML on Cheyenne. If that's not the case, then there is a bug somewhere.

I also ran tests on Cheyenne and they passed, but you made a change after these tests were run that resulted in the final result failing on platforms where the partition tag is not supported. This change I believe: https://github.com/NOAA-EMC/regional_workflow/pull/309/commits/26b4089c906398a9a1560886d9eafe4e31cf90a3 

I can add some more details on the specifics of the problem and the fix if you like. This fix will need to be applied to `develop` as well.

mkavulich:
Testing update: Once I pulled in the latest changes, all tests I ran were successful on Hera:

 - DOT_OR_USCORE  
 - grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16beta
 - get_extrn_files_from_hpss_FV3GFS_FV3GFS
 - grid_RRFS_CONUS_25km_ics_HRRRX_lbcs_RAPX_suite_GSD_SAR
 - get_extrn_files_from_hpss_GSMGFS_GSMGFS
 - grid_RRFS_CONUS_25km_ics_HRRRX_lbcs_RAPX_suite_RRFS_v1beta
 - get_extrn_files_from_hpss_HRRRX_RAPX
 - get_extrn_files_from_hpss_RAPX_RAPX
 - grid_RRFS_CONUS_3km_ics_HRRRX_lbcs_RAPX_suite_RRFS_v1beta
 - grid_CONUS_25km_GFDLgrid_FV3GFS_FV3GFS
 - new_GFDLgrid__GFDLgrid_USE_GFDLgrid_RES_IN_FILENAMES_eq_FALSE
 - new_GFDLgrid__GFDLgrid_USE_GFDLgrid_RES_IN_FILENAMES_eq_TRUE
 - grid_RRFS_CONUS_13km_ics_HRRRX_lbcs_RAPX_suite_GSD_SAR 
 - suite_FV3_GFS_v15p2
 - grid_RRFS_CONUS_13km_ics_HRRRX_lbcs_RAPX_suite_RRFS_v1beta
 - suite_FV3_RRFS_v1beta

ush/templates/FV3LAM_wflow.xml
```diff
@@ -64,17 +64,17 @@ the "HPSS" type is used for the GET_EXTRN_ICS_TN and GET_EXTRN_LBCS_TN
 tasks; and the "FCST" type is used for the RUN_FCST_TN task.
 -->
 
-{%- if (partition_default != "")  %}
+{%- if partition_default is not none %}
```
gsketefian, line 67: I had to search quite a bit online to find documentation on how to this comparison in Jinja2.  Apparently the "!=" method doesn't work on the version of Jinja2 on Cheyenne.  If you know of good documentation (preferably with examples!), let me know.  Thx.

gsketefian:
@mkavulich Great!  Approving.

gsketefian (approved):

ush/templates/FV3LAM_wflow.xml
mkavulich: Unfortunately I don't know too much about jinja still; this fix came from @christinaholtNOAA, so I guess I should list her as a co-author!

christinaholtNOAA: This is needed since the empty-string bash variable comes in through YAML and that layer converts to a NoneType value in Python. Once it gets here in the Jinja template it is no longer an empty string and can't be compared reliably as such.

ush/fill_jinja_template.py
```diff
@@ -259,7 +259,7 @@ def main(cla):
     tvars = {}
     for var in template_vars:
 
-        if cfg.get(var) is None:
+        if cfg.get(var, "NULL") == "NULL":
```
christinaholtNOAA, line 262: It's worth mentioning here that Python's behavior returns a NoneType value when a dictionary key does not exist. The old logic will not differentiate between a variable that is set to None (as is the case for empty string bash variables) and one that does not exist. This fix changes that behavior from the default.

PR #344, opened 2020-10-28
Incorporate Graduate Student Test (GST) into config.community.sh

gsketefian:
## DESCRIPTION OF CHANGES: 
* Modify config.community.sh to be very similar to the one in the release branch, which is being used for the Graduate Student Test (GST).
* Add the GST as another WE2E test.  It is called GST_release_public_v1.
* Modify the WE2E testing script (run_experiments.sh) so that it can handle both nemsio and grib2 format files from FV3GFS and GSMGFS.  This is needed for the GST test because that uses grib2 files.
* Add new file (testlist.release_public_v1.txt) containing a minimal list of WE2E tests that should be run for PRs to make sure that release features aren't broken.

## TESTS CONDUCTED: 
On Hera, ran the new WE2E test **GST_release_public_v1** as well as **grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2**.  Both succeeded.

## CONTRIBUTORS (optional): 
@jwolff-ncar


ush/config.community.sh
```diff
@@ -10,17 +10,28 @@ PREEXISTING_DIR_METHOD="rename"
 PREDEF_GRID_NAME="RRFS_CONUS_25km"
 GRID_GEN_METHOD="ESGgrid"
 QUILTING="TRUE"
-CCPP_PHYS_SUITE="FV3_GFS_2017_gfdlmp"
-FCST_LEN_HRS="6"
+
+CCPP_PHYS_SUITE="FV3_GFS_v15p2"
+FCST_LEN_HRS="48"
```
jwolff-ncar, line 15: If you don't want to run the entire 48 hours you do not have to. For this type of test just doing 6 hours is fine.

jwolff-ncar (commented):
I am good with this. See one comment in config.community.sh.

ush/config.community.sh
gsketefian: @jwolff-ncar I think I'll leave it at 48 hr to make it as similar to the actual GST as possible.  There is another WE2E test that is pretty similar to this but only goes out to 6 hr.  Also, the whole thing is very fast (less than 5 mins) on hera.  We can change it back to 6 hr if it's a problem for others.

Can you give me an approve so I can merge?  Thx.

jwolff-ncar (approved):

PR #351, opened 2020-11-04
Develop branch: Update module files to source READMEs in ufs-srweather-app and change HALO_BLEND default value back to 10.

JeffBeck-NOAA:
## DESCRIPTION OF CHANGES: 
1. Modified source of module files from the individual repositories to ufs-srweather-app.  The system-specific Intel-based README file is now copied over into regional_workflow/modulefiles/tasks.  Note that we should eventually create a user-defined ${COMPILER} variable to allow for copying over either intel or gnu README files, based on how the user compiles their code.  

Each individual task has an identical module file, since they all come from the same system-specific ufs-srweather-app README file.  I hesitated creating one single module file, but we can do that in the future if we're confident things will stay this way.

Each module file (which is really just a bash script), is sourced, followed by a "module load" of the .local file, if it exists.

Note that the modifications in `generate_FV3LAM_wflow.sh` and `load_modules_run_task.sh` preserve old code in comments just in case a decision is made in the future to revert any portion of these changes.
  
2.  Change the default HALO_BLEND value in config_defaults.sh back to 10.

## TESTS CONDUCTED: 
Tested an end-to-end 25-km CONUS run on Hera, which ran to completion.  This PR *should* run fine on other platforms, but a test on Jet and other systems would be nice to confirm.

## CONTRIBUTORS (optional): 
@mkavulich, @gsketefian


ush/generate_FV3LAM_wflow.sh
```diff
@@ -330,30 +330,30 @@ done
...
-cp_vrfy -f "${UFS_UTILS_DIR}/modulefiles/build.$machine" "${MAKE_OROG_TN}"
-cp_vrfy -f "${UFS_UTILS_DIR}/modulefiles/build.$machine" "${MAKE_SFC_CLIMO_TN}"
-cp_vrfy -f "${UFS_UTILS_DIR}/modulefiles/build.$machine" "${MAKE_ICS_TN}"
-cp_vrfy -f "${UFS_UTILS_DIR}/modulefiles/build.$machine" "${MAKE_LBCS_TN}"
-if [ $MACHINE = "WCOSS_CRAY" -o $MACHINE = "WCOSS_DELL_P3" ] ; then
-  cp_vrfy -f "${UFS_WTHR_MDL_DIR}/modulefiles/$machine/fv3" "${RUN_FCST_TN}"
-else
-  cp_vrfy -f "${UFS_WTHR_MDL_DIR}/modulefiles/$machine.intel/fv3" "${RUN_FCST_TN}"
-fi
+cp_vrfy -f "${SR_WX_APP_TOP_DIR}/docs/README_${machine}_intel.txt" "${MAKE_GRID_TN}"
```
gsketefian: Since this file(s) (which I think we should refer to as platform- and compiler-specific environment files) is just going to be sourced as-is (it's not going to be combined with the local files to create a new file), we don't need to do a copy here.  In load_modules_run_task.sh, we should just source this file.  So I think you can just comment out lines 347 to 379 (but make some changes to load_modules_run_task.sh; see below).

ush/load_modules_run_task.sh
```diff
@@ -278,13 +278,36 @@ Call to \"module use\" command failed."
      module use -a "${modules_dir}" || print_err_msg_exit "\
 Call to \"module use\" command failed."
     
-     module load "${modulefile_name}" || print_err_msg_exit "\
-Loading of module file (modulefile_name; in directory specified by mod-
+     #module load "${modulefile_name}" || print_err_msg_exit "\
```
gsketefian: Suggestions:
1) Define a COMPILER workflow variable in config_defaults.sh (as you suggested) and set it to "intel" for now.  This variable along with valid_vals_CCPP_PHYS_SUITE (and possibly others later) must eventually be obtained from the build system, but we'll worry about that later.
2) Move this section up to line 147 before we deal with any module files at all.  Maybe do something like this:
```
  env_fn="README_${machine}_${COMPILER}.txt"
  env_fp="${SR_WX_APP_TOP_DIR}/docs/${env_fn}"
  source "${env_fp}" || print_err_msg_exit "\
Sourcing of platform- and compiler-specific environment file (env_fp) for the 
workflow task specified by task_name failed:
  task_name = \"${task_name}\"
  env_fp = \"${env_fp}\""
```

Also, need to remove the "module purge" on line 268.

PR #356, opened 2020-11-04
Release branch: Update module files to source READMEs in ufs-srweather-app and change HALO_BLEND default value back to 10.

JeffBeck-NOAA:
DESCRIPTION OF CHANGES:

1. Modified source of module files from the individual repositories to ufs-srweather-app. The system-specific Intel-based README file is now copied over into regional_workflow/modulefiles/tasks. Note that we should eventually create a user-defined ${COMPILER} variable to allow for copying over either intel or gnu README files, based on how the user compiles their code.

Each individual task has an identical module file, since they all come from the same system-specific ufs-srweather-app README file. I hesitated creating one single module file, but we can do that in the future if we're confident things will stay this way.

Each module file (which is really just a bash script), is sourced, followed by a "module load" of the .local file, if it exists.

Note that the modifications in `generate_FV3LAM_wflow.sh` and `load_modules_run_task.sh` preserve old code in comments just in case a decision is made in the future to revert any portion of these changes.

2. Change the default HALO_BLEND value in config_defaults.sh back to 10.

TESTS CONDUCTED:

Tested an end-to-end 25-km CONUS run on Jet and Hera, which ran to completion. This PR should run fine on other platforms, but a test on Jet and other systems would be nice to confirm.

CONTRIBUTORS (optional):

@mkavulich, @gsketefian


ush/load_modules_run_task.sh
```diff
@@ -147,6 +147,22 @@ jjob_fp="$2"
 #
 #-----------------------------------------------------------------------
 #
+# Sourcing ufs-srweather-app README file (in directory specified by mod-
+# ules_dir) for the specified task
+#
+#-----------------------------------------------------------------------
+#
+machine=${MACHINE,,}
+  env_fn="README_${machine}_${COMPILER}.txt"
```
gsketefian: Lines 156, 157, and 158: Remove indentation.

ush/generate_FV3LAM_wflow.sh
```diff
@@ -330,54 +330,55 @@ done
...
+#cp_vrfy -f "${SR_WX_APP_TOP_DIR}/docs/README_${machine}_intel.txt" "${MAKE_ICS_TN}"
+#cp_vrfy -f "${SR_WX_APP_TOP_DIR}/docs/README_${machine}_intel.txt" "${MAKE_LBCS_TN}"
+#if [ $MACHINE = "WCOSS_CRAY" -o $MACHINE = "WCOSS_DELL_P3" ] ; then
+#  cp_vrfy -f "${UFS_WTHR_MDL_DIR}/modulefiles/${machine}/fv3" "${RUN_FCST_TN}"
+#else
+#  cp_vrfy -f "${UFS_WTHR_MDL_DIR}/modulefiles/${machine}.intel/fv3" "${RUN_FCST_TN}"
+#fi
+#cp_vrfy -f "${SR_WX_APP_TOP_DIR}/docs/README_${machine}_intel.txt" "${RUN_FCST_TN}"
 
 task_names=( "${MAKE_GRID_TN}" "${MAKE_OROG_TN}" "${MAKE_SFC_CLIMO_TN}" "${MAKE_ICS_TN}" "${MAKE_LBCS_TN}" "${RUN_FCST_TN}" )
```
gsketefian: I think line 358 can also be commented out (task_names=...).

ush/generate_FV3LAM_wflow.sh
```diff
@@ -330,54 +330,55 @@ done
 #
```
gsketefian, line 335: Can you put a comment here with lots of ####s saying the following section (lines ??? to ???) can be removed once we know the new way of dealing with environments/modulefiles works.  Also, maybe make an issue to remind us to remove this later on.

ush/load_modules_run_task.sh
```diff
@@ -191,113 +206,116 @@ use_default_modulefile=0
 #
```
gsketefian, line 206: Same thing here.  Put in an obvious comment saying this section can be removed once we know the new approach works.

gsketefian (approved):

PR #361, opened 2020-11-11
Add the RRFS_v1alpha suite

gsketefian:
This PR should be merged at the same time as PR #[51](https://github.com/ufs-community/ufs-srweather-app/pull/51) into ufs-srweather-app:release/public-v1 branch.

## DESCRIPTION OF CHANGES:
* Add the RRFS_v1alpha suite to the set of valid suites that can be used in the workflow.  This entails:
  * Adding this suite to the list of valid values for CCPP_PHYS_SUITE in ush/valid_param_vals.sh.
  * Adding this suite to the various if-statements in the ex-scripts for the make_ics and make_lbcs tasks.
  * Adding a stanza for this suite in ush/templates/FV3.input.yml.
  * Adding a diag_table for this suite.
  * Adding a field_table for this suite.

  Note that the RRFS_v1alpha suite should be identical to RRFS_v1beta except that the MYNN surface layer parameterization is replaced with the GFS one.  However, currently in the ufs-weather-model, RRFS_v1alpha differs from RRFS_v1beta not only in the surface parameterization but also in the use of the Noah MP land surface model (instead of the Noah LSM).  Eventually, a PR should go into ufs-weather-model (and fv3atm) to set the LSM in RRFS_v1alpha to Noah (without MP).
* Add WE2E tests for the RRFS_v1alpha suite.
* Change the list of WE2E tests that must be run for the release branch (for testing PRs) to use the RRFS_v1alpha suite instead of RRFS_v1beta.
* Remove the RRFS_v1beta suite from the list of valid values for CCPP_PHYS_SUITE in ush/valid_param_vals.sh.  This disables the use of RRFS_v1beta in the release branch.  This must be done because the RRFS_V1beta suite has been removed from the release/public-v2 branch of ufs-weather-model.
* Remove from the yaml configuration file for the FV3 namelist (FV3.input.yml) those suites that are neither in the develop branch nor the release/public-v2 branch of ufs-weather-model (or, really of its fv3atm submodule).  These suites are FV3_GSD_SAR_v1 and FV3_RRFS_v0.
* Modify the yaml configuration file for the FV3 namelist (FV3.input.yml) so that those suites that use Thompson micorphysics set the variable effr_in to False.  Note that we make this change not only for the suites that can be used with the release/public-v1 branch (of regional_workflow) but also the ones that can be sued with the develop branch in order to keep differences between these two branches to a minimum.

## TESTS CONDUCTED: 
The following WE2E tests were run on Hera:
* GST_release_public_v1
* grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha
* grid_RRFS_CONUS_13km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha
* grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
* grid_RRFS_CONUS_3km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha

All were successful.  Note that the tests involving RRFS_v1alpha are using the Noah MP LSM.  This will be changed in the future (in the fv3atm repo), at which point the tests should be rerun.

## CONTRIBUTORS (optional): 
@climbfuji @JacobCarley-NOAA


JeffBeck-NOAA (approved):
Looks good!

ush/templates/FV3.input.yml
```diff
@@ -42,19 +42,31 @@ FV3_GSD_SAR_v1:
...
+  gfs_physics_nml:
+    <<: *gsd_sar_phys
     do_deep: False
+    do_mynnsfclay: False
+    imfdeepcnv: -1
+    imfshalcnv: -1
+    lsm: 2
+    lsoil_lsm: 4
+  atmos_model_nml:
+    fdiag: 3
```
JeffBeck-NOAA, line 51: @gsketefian, Dom's [PR](https://github.com/NOAA-EMC/fv3atm/pull/207#event-4065342034) to add the RRFS_v1alpha SDF, which will be merged soon, includes a requirement for "effr_in" to be defined in the namelist.  Therefore, can you add "effr_in: False" to the list of settings here for FV3_RRFS_v1alpha?  Thanks.

JeffBeck-NOAA (changes requested):

ush/templates/FV3.input.yml
climbfuji: Will that be set `.true.` for Thompson MP? That is a requirement. GFS v15p2 I believe uses `.false.`, GFS v16 `.true.` (but need to double check that one)

JeffBeck-NOAA: My mistake, it should have been effr_in: True.  Thanks, @climbfuji!

gsketefian: @climbfuji I first mistakenly set effr_in to False for the RRFS_v1alpha suite.  The forecast model log file had the (correct) message:
`Thompson MP requires effr_in to be set to .true. - job aborted
`
but the forecast job didn't actually abort; it keeps on going until it runs out of wallclock.  Can we fix this so we don't waste core-hours?  Thx.

climbfuji: I don't think that's on my end. The "stop" is the same as for any other exception in `GFS_typedefs.F90`. I remember vaguely that @tanyasmirnova had issues with slurm not aborting a job correctly when slurm was introduced, and that the sysadmins gave her a solution/workaround.

gsketefian:
@climbfuji @JeffBeck-NOAA I just noticed in our yaml input namelist configuration file that effr_in for the following suites is set to True:

FV3_CPT_v0
FV3_GFS_2017_gfdlmp
FV3_GFS_2017_gfdlmp_regional
FV3_GFS_v16beta

I believe none of these use Thompson mp.  Does that mean this is an error and these should be reset to False, or is there another reason to set these to True?

climbfuji:
It doesn't mean it is an error. gfdlmp also has effr_in, it used to be .false. in gfs v15p2, but I think it is true for gfs v16 - please check against the operational namelist!

gsketefian:
@climbfuji Ok, thanks.  I am checking with Eric Aligo.

JeffBeck-NOAA (approved):

PR #366, opened 2020-11-13
Develop branch: Point exregional_make_ics.sh script to the new location for geogrid files (fix_am directory)

JeffBeck-NOAA:
## DESCRIPTION OF CHANGES: 
Point exregional_make_ics.sh script to the new location for geogrid files (fix_am directory).
Set the config.community.sh external model staging user-defined variable to "false."

## TESTS CONDUCTED: 
Tested on Jet.

## CONTRIBUTORS (optional): 
@KateFriedman-NOAA


scripts/exregional_make_ics.sh
```diff
@@ -421,12 +421,18 @@ case "${EXTRN_MDL_NAME_ICS}" in
     nsoill_out="9"
   fi
 #
-# These geogrid files need to be moved to more permanent locations.
+# Paths to the geogrid file for the HRRRX.
 #
   if [ "${MACHINE}" = "HERA" ]; then
-    geogrid_file_input_grid="/scratch2/BMC/det/beck/FV3-LAM/geo_em.d01.nc_HRRRX"
+    geogrid_file_input_grid="/scratch1/NCEPDEV/global/glopara/fix/fix_am/geo_em.d01.nc_HRRRX"
```
gsketefian: Except for Orion, these paths are in the variable FIXgsm that gets set in setup.sh (line 666).  So you can just remove all the if-statements and instead use:
`geogrid_file_input_grid="$FIXgsm/geo_em.d01.nc_HRRRX"
`
For Orion, the value set in setup.sh is different that the one here, but I wonder if we can change that.  I got the one in setup.sh a while ago from Ben.  If the two directories contain the same files, then let's just change the one in setup.sh to the one here so we don't need any if-statements.

scripts/exregional_make_ics.sh
```diff
@@ -463,14 +469,18 @@ case "${EXTRN_MDL_NAME_ICS}" in
     nsoill_out="9"
   fi
 #
-# These geogrid files need to be moved to more permanent locations.
+# Paths to the geogrid file for the RAPX.
 #
   if [ "${MACHINE}" = "HERA" ]; then
-    geogrid_file_input_grid="/scratch2/BMC/det/beck/FV3-LAM/geo_em.d01.nc_RAPX"
+    geogrid_file_input_grid="/scratch1/NCEPDEV/global/glopara/fix/fix_am/geo_em.d01.nc_RAPX"
```
gsketefian: Same comment here as above for HRRRX (i.e. just use FIXgsm).

ush/config.community.sh
```diff
@@ -1,15 +1,15 @@
...
-EXPT_SUBDIR="test_community"
+EXPT_SUBDIR="test_CONUS_25km_GFSv15p2"
 
 VERBOSE="TRUE"
 
 RUN_ENVIR="community"
 PREEXISTING_DIR_METHOD="rename"
 
 PREDEF_GRID_NAME="RRFS_CONUS_25km"
+GRID_GEN_METHOD="ESGgrid"
```
gsketefian: This should no longer be specified since it is part of the definition of a predefined grid.

ush/config.community.sh
```diff
@@ -28,7 +28,7 @@ WTIME_RUN_FCST="01:00:00"
 
 # The following is specifically for Hera.  It will have to be modified
 # if on another platform, using other dates, other external models, etc.
-USE_USER_STAGED_EXTRN_FILES="TRUE"
+USE_USER_STAGED_EXTRN_FILES="FALSE"
```
gsketefian: The default value of USE_USER_STAGED_EXTRN_FILES is already FALSE.  The point of this section is to demonstrate use of user-staged external model files.  Maybe better to comment out this line but with a comment explaining how to use the feature.  Overall, maybe something like this:
```
#
# Uncomment the following line in order to use user-staged external model 
# files with locations and names as specified by EXTRN_MDL_SOURCE_BASEDIR_ICS/
# LBCS and EXTRN_MDL_FILES_ICS/LBCS.
#
#USE_USER_STAGED_EXTRN_FILES="TRUE"
#
# The following is specifically for FV3GFS on Hera.  It will have to be
# modified if using another external model, another platform, etc.
#
EXTRN_MDL_SOURCE_BASEDIR_ICS="/scratch2/BMC/det/UFS_SRW_app/v1p0/model_data/FV3GFS"
EXTRN_MDL_FILES_ICS=( "gfs.pgrb2.0p25.f000" )
EXTRN_MDL_SOURCE_BASEDIR_LBCS="/scratch2/BMC/det/UFS_SRW_app/v1p0/model_data/FV3GFS"
EXTRN_MDL_FILES_LBCS=( "gfs.pgrb2.0p25.f006" "gfs.pgrb2.0p25.f012" "gfs.pgrb2.0p25.f018" "gfs.pgrb2.0p25.f024" \
                       "gfs.pgrb2.0p25.f030" "gfs.pgrb2.0p25.f036" "gfs.pgrb2.0p25.f042" "gfs.pgrb2.0p25.f048" )
```

scripts/exregional_make_ics.sh
JeffBeck-NOAA: @gsketefian, thanks for the review!  The exregional_make_ics.sh file now points to a single location for $FIXgsm/geo_em.d01.nc_HRRRX or $FIXgsm/geo_em.d01.nc_RAPX, regardless of platform.  I also changed the setup.sh script to reference the standard "glopara" path for "fix_am" on Orion, instead of the "cam" path. Also, for Orion, Jet, and Hera, I changed the SFC_CLIMO_INPUT_DIR paths to point to "glopara" instead of separate (or personal) directories.

scripts/exregional_make_ics.sh
JeffBeck-NOAA: Fixed!

ush/config.community.sh
JeffBeck-NOAA: Removed.

ush/config.community.sh
JeffBeck-NOAA: Updated!

ush/setup.sh
```diff
@@ -680,19 +680,19 @@ case $MACHINE in
   "HERA")
     FIXgsm=${FIXgsm:-"/scratch1/NCEPDEV/global/glopara/fix/fix_am"}
     TOPO_DIR=${TOPO_DIR:-"/scratch1/NCEPDEV/global/glopara/fix/fix_orog"}
-    SFC_CLIMO_INPUT_DIR=${SFC_CLIMO_INPUT_DIR:-"/scratch1/NCEPDEV/da/George.Gayno/ufs_utils.git/climo_fields_netcdf"}
+    SFC_CLIMO_INPUT_DIR=${SFC_CLIMO_INPUT_DIR:-"/scratch1/NCEPDEV/global/glopara/fix/fix_sfc_climo"}
```
gsketefian, line 683: I approved the PR, but I just want to make sure these changes in paths don't affect results.  Did you check that the same files (with same sizes) are in the new directories as the old?  Same question for Orion and Jet.

gsketefian (approved):

PR #371, opened 2020-12-03
Maximize RRFS_CONUS_[25,13,3]km grids so they span as much of HRRR CONUS domain as possible

gsketefian:
## DESCRIPTION OF CHANGES: 
* Change grid parameters for the RRFS_COUNS_25km, RRFS_CONUS_13km, and RRFS_CONUS_3km grids so that they span as much of the HRRR CONUS domain as possible.  Note that this is done without regard to whether nx and ny are evenly divisible by layout_x and layout_y because these restrictions have recently been removed from the forecast model.
* Remove several checks on the grid parameters.  These are:
  (1) The check on whether NX and NY are evenly divisible by LAYOUT_X and LAYOUT_Y, respectively.  This restriction has been removed from the forecast model.
  (2) The check on BLOCKSIZE, i.e. whether the quantity (NX/LAYOUT_X)*(NY/LAYOUT_Y) is evenly divisible by BLOCKSIZE.  This already depends on restriction (1) being satisfied, so it cannot be satisfied if (1) is not satisfied.
  (3) The check on whether NY is evenly divisible by the number of write-component tasks per write-component group.  Tests where run with parameters that violated this test, and the model did not crash, so we assume this is no longer a restriction.
* Remove section of code that copies the FV3 executable from its location in the ufs-weather-model directory to the bin directory for all other executables.  This is no longer necessary since the new build system already places this executable in the bin directory.
* Change the name of FV3 executable to NEMS.exe, which is the file that the new build system creates and places in the bin directory.
* Modify NCL scripts to make them work with the latest version of the workflow.
* Remove commented-out code (e.g. in setup.sh).
* Edit comments.

## TESTS CONDUCTED:
Note:  The following tests were run with hash 6891c8a7 of UFS_UTILS, but the latest is 879dc76.

Ran the following WE2E tests on hera (with the new grid parameters):
* GST_release_public_v1
* grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2

All were successful except **grid_RRFS_CONUS_3km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta**, which crashed after hour 4 (see further discussion below).  Since this test is with the FV3_RRFS_v1beta suite, and this suite is going to be replaced with the RRFS_v1alpha suite very soon (because the MYNN surface layer parameterization seems unstable), we will not dwell on this failure and retest again in a new PR once RRFS_v1beta is replaced with RRFS_v1alpha.  Those tests will be with the latest hash of UFS_UTILS (879dc76).


ush/generate_FV3LAM_wflow.sh
```diff
@@ -526,68 +469,6 @@ cp_vrfy "${CCPP_PHYS_SUITE_IN_CCPP_FP}" "${CCPP_PHYS_SUITE_FP}"
...
-
-if [ ! -f "${exec_fp}" ]; then
-  print_err_msg_exit "\
-The executable (exec_fp) for running the forecast model does not exist:
-  exec_fp = \"${exec_fp}\"
-Please ensure that you've built this executable."
-fi
-#
-# Make a copy of the executable in the executables directory only if a
-# copy doens't already exist or if a copy does exist but is older than
```
JeffBeck-NOAA, line 576: Typo here: "doesn't"

JeffBeck-NOAA (approved):

PR #382, opened 2020-12-24
Add the FV3_HRRR suite to the workflow

gsketefian:
## NOTE:
PR #[95](https://github.com/ufs-community/ufs-srweather-app/pull/95) into ufs-srweather-app should be merged right after this PR.

### Main changes:
* Add the FV3_HRRR suite to the workflow.  This consists of:
  * Adding "FV3_HRRR" to the list of valid values of CCPP_PHYS_SUITE in valid_param_vals.sh.
  * Adding the diag_table.FV3_HRRR and field_table.FV3_HRRR files to the ush/templates directory. 
  * Adding the FV3_HRRR suite to the case-statements in exregional_make_ics.sh and exregional_make_lbcs.sh.
  * Adding a stanza to FV3.input.yml to customize the FV3 namelist file for the FV3_HRRR suite.
* Add the following two WE2E tests for the FV3_HRRR suite on the RRFS_CONUS_25km grid.  The first is initialized for the 2020 derecho case (2020081000) and updates the LBCs every hour while the second is initialized on 2019052000 and updates the LBCs every 6 hours:
  * grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_RAP_suite_HRRR
  * suite_FV3_HRRR
* Add an additional three WE2E tests for the FV3_HRRR suite on the 25km, 13km, and 3km RRFS_CONUS grids.  All three are initialized for the 2020 derecho case (2020081000).  They are named as follows:
  * grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_HRRR
  * grid_RRFS_CONUS_13km_ics_HRRR_lbcs_RAP_suite_HRRR
  * grid_RRFS_CONUS_3km_ics_HRRR_lbcs_RAP_suite_HRRR
* Modify the following three WE2E tests that use the FV3_GSD_SAR suite so that they are initialized for the 2020 derecho case (2020081000) instead of 2020080100 (so that they are consistent with the 3 new WE2E tests added above):
  * grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_GSD_SAR
  * grid_RRFS_CONUS_13km_ics_HRRR_lbcs_RAP_suite_GSD_SAR
  * grid_RRFS_CONUS_3km_ics_HRRR_lbcs_RAP_suite_GSD_SAR
* Enable the workflow to find and copy to the experiment directory the pregenerated (for a given grid) orography statistics files `${CRES}_oro_data_ls.tile7.halo0.nc` and `${CRES}_oro_data_ss.tile7.halo0.nc` needed by the gravity wave drag parameterization in FV3_HRRR.  (Note that this is a temporary fix.  Eventually, a program will be made available that the make_orog task can call to generate these files as part of the workflow for any grid.)  Modifications for this purpose consist of:
  * Adding to config_defaults.sh the new workflow variable GWD_HRRRsuite_BASEDIR that specifies the base directory in which the aformentioned "ls" and "ss" orography statistics files are located for the predefined grid being used.  This workflow variable will be removed once a program is available to generate the "ls" and "ss" files on-the-fly for any grid.
  * Adding code to setup.sh that sets the new derived/secondary workflow variable GWD_HRRRsuite_DIR that specifies the directory in which the "ls" and "ss" files are located (as opposed to just the base directory, which is what GWD_HRRRsuite_BASEDIR represents).  This is set using GWD_HRRRsuite_BASEDIR and PREDEF_GRID_NAME.  Like GWD_HRRRsuite_BASEDIR, this variable will be removed once a program is available to generate the "ls" and "ss" files on-the-fly for any grid.
  * Adding code to link_fix.sh that creates symlinks in the fv3_lam directory that point to the "ls" and "ss" files in OROG_DIR. 
  * Adding code to exregional_make_orog.sh that copies the "ls" and "ss" files from the directory specified by GWD_HRRRsuite_DIR to the orography directory (OROG_DIR).  This code will evenutally be replaced by a call to a program that will generate these files and place them in OROG_DIR (instead of copying them).
  * Adding code to exregional_run_fcst.sh that creates links in the run directory to the"ls" and "ss" orography files that the FV3 model looks for.
  * Adding code to run_experiments.sh that sets GWD_HRRRsuite_BASEDIR in the workflow configuration file (config.sh) to the base directory in which the pregenerated "ls" and "ss" orography files are located.  Note that this setting is machine-dependent.

### Improvements:
* Edit error messages in exregional_make_ics.sh and exregional_make_lbcs.sh to make them clearer.
* Correct code indentation and remove blank lines.
* Alphabetize FV3_GSD_SAR stanza in FV3.input.yml.
* In tests/run_experiments.sh, change the location FIXLAM_NCO_BASEDIR on Jet to be a more "official" directory (instead of a personal directory).
* Add the WE2E test grid_RRFS_CONUS_25km_ics_HRRR_lbcs_HRRR_suite_GSD_SAR that uses the HRRRX external model for both the ICs and the LBCs (and uses the FV3_GSD_SAR suite).
* Clean up comments.

## TESTS CONDUCTED:
Ran the following WE2E tests on Hera:

* grid_RRFS_CONUS_13km_ics_HRRR_lbcs_RAP_suite_GSD_SAR
* grid_RRFS_CONUS_13km_ics_HRRR_lbcs_RAP_suite_GSD_SAR_2020020800
* grid_RRFS_CONUS_13km_ics_HRRR_lbcs_RAP_suite_HRRR
* grid_RRFS_CONUS_13km_ics_HRRR_lbcs_RAP_suite_HRRR_2020020800
* grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GSD_SAR
* grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_RAP_suite_HRRR
* grid_RRFS_CONUS_25km_ics_HRRR_lbcs_HRRR_suite_GSD_SAR
* grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_GSD_SAR
* grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_GSD_SAR_2020020800
* grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_HRRR
* grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_HRRR_2020020800
* grid_RRFS_CONUS_25km_modify_DT_ATMOS_LAYOUT_XY_BLOCKSIZE
* grid_RRFS_CONUS_3km_ics_HRRR_lbcs_RAP_suite_GSD_SAR
* grid_RRFS_CONUS_3km_ics_HRRR_lbcs_RAP_suite_GSD_SAR_2020020800
* grid_RRFS_CONUS_3km_ics_HRRR_lbcs_RAP_suite_HRRR
* grid_RRFS_CONUS_3km_ics_HRRR_lbcs_RAP_suite_HRRR_2020020800
* nco_CONUS_25km_GFDLgrid
* nco_RRFS_CONUS_25km_HRRR_RAP
* suite_FV3_GSD_v0
* suite_FV3_HRRR

All tests were successful. Note that the six tests whose names end with "_2020020800" are not in the workflow.  They are generated by taking the respective WE2E tests in the workflow with names without the trailing "_2020020800" and simply changing DATE_FIRST_CYCL and DATE_LAST_CYCL from 20200801 to 20200208 and LBC_SPEC_INTVL_HRS from 3 to 1. The change in date was to try out a winter case.

## CONTRIBUTORS (optional): 
@mdtoy


scripts/exregional_make_ics.sh
```diff
@@ -497,8 +494,16 @@ case "${EXTRN_MDL_NAME_ICS}" in
   elif [ "${CCPP_PHYS_SUITE}" = "FV3_GSD_v0" ] || \
        [ "${CCPP_PHYS_SUITE}" = "FV3_GSD_SAR" ] || \
        [ "${CCPP_PHYS_SUITE}" = "FV3_RRFS_v1alpha" ] || \
-       [ "${CCPP_PHYS_SUITE}" = "FV3_RRFS_v1beta" ]; then
+       [ "${CCPP_PHYS_SUITE}" = "FV3_RRFS_v1beta" ] || \
+       [ "${CCPP_PHYS_SUITE}" = "FV3_HRRR" ]; then
     thomp_mp_climo_file="${FIXam}/Thompson_MP_MONTHLY_CLIMO.nc"
+  else
+    print_err_msg_exit "\
```
JeffBeck-NOAA: @gsketefian, thanks for adding these print statements for undefined IC/CCPP combinations.  In this stanza, it would probably be better to say "A determination of whether the variable "thomp_mp_climo_file" should be defined for the chosen IC/CCPP combination has not yet been made" or something similar to that.

scripts/exregional_make_lbcs.sh
```diff
@@ -297,20 +294,20 @@ case "${EXTRN_MDL_NAME_LBCS}" in
   elif [ "${CCPP_PHYS_SUITE}" = "FV3_GSD_v0" ] || \
        [ "${CCPP_PHYS_SUITE}" = "FV3_GSD_SAR" ] || \
        [ "${CCPP_PHYS_SUITE}" = "FV3_RRFS_v1alpha" ] || \
-       [ "${CCPP_PHYS_SUITE}" = "FV3_RRFS_v1beta" ]; then
+       [ "${CCPP_PHYS_SUITE}" = "FV3_RRFS_v1beta" ] || \
+       [ "${CCPP_PHYS_SUITE}" = "FV3_HRRR" ]; then
      thomp_mp_climo_file="${FIXam}/Thompson_MP_MONTHLY_CLIMO.nc"
   else
     print_err_msg_exit "\
```
JeffBeck-NOAA: Same comment as for undefined IC/CCPP combinations.

scripts/exregional_make_lbcs.sh
```diff
@@ -323,52 +320,41 @@ case "${EXTRN_MDL_NAME_LBCS}" in
...
```
JeffBeck-NOAA: Same comment as undefined IC/CCPP combinations.

scripts/exregional_make_lbcs.sh
```diff
@@ -383,24 +369,26 @@ case "${EXTRN_MDL_NAME_LBCS}" in
...
```
JeffBeck-NOAA: Same comment here too.

JeffBeck-NOAA (changes requested):

JeffBeck-NOAA (approved):

scripts/exregional_make_ics.sh
```diff
@@ -417,34 +410,36 @@ case "${EXTRN_MDL_NAME_ICS}" in
      [ "${CCPP_PHYS_SUITE}" = "FV3_GFS_v15p2" ] || \
      [ "${CCPP_PHYS_SUITE}" = "FV3_GFS_v16beta" ] || \
      [ "${CCPP_PHYS_SUITE}" = "FV3_RRFS_v1alpha" ] || \
-     [ "${CCPP_PHYS_SUITE}" = "FV3_RRFS_v1beta" ]; then
+     [ "${CCPP_PHYS_SUITE}" = "FV3_RRFS_v1beta" ] || \
+     [ "${CCPP_PHYS_SUITE}" = "FV3_HRRR" ]; then
```
christinaholtNOAA: We learned in our recent RRFS_dev4 tests that FV3_HRRR should have nsoill_out=9 for both RAP and HRRR ICS and LBCS.

scripts/exregional_make_ics.sh
```diff
@@ -456,34 +451,36 @@ case "${EXTRN_MDL_NAME_ICS}" in
...
```
christinaholtNOAA: FV3_HRRR should be in elif.

tests/baselines_list.txt
```diff
@@ -17,14 +17,17 @@ grid_RRFS_AK_3km_RAP_RAP
...
 grid_RRFS_CONUS_13km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
 grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
 grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16beta
 grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_GSD_SAR
+grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_HRRR
 grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
 grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
 grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16beta
 grid_RRFS_CONUS_3km_ics_HRRR_lbcs_RAP_suite_GSD_SAR
+grid_RRFS_CONUS_3km_ics_HRRR_lbcs_RAP_suite_HRRR
```
christinaholtNOAA, line 44: Are these tests expected to pass? Does develop have an appropriate tag of the model to run them successfully?

ush/templates/FV3.input.yml
```diff
@@ -62,6 +62,19 @@ FV3_RRFS_v1beta:
...
+    nrows_blend: 10
+    regional_bcs_from_gsi: False
+    write_restart_with_bcs: False
+  gfs_physics_nml:
+    <<: *gsd_sar_phys
+    do_mynnsfclay: True
+    imfdeepcnv: -1
+    imfshalcnv: -1
+    gwd_opt: 3
+
```
christinaholtNOAA: The settings you've added here are very different from those we use in AVID.

scripts/exregional_make_ics.sh
gsketefian: Ok, will change.

scripts/exregional_make_ics.sh
gsketefian: Ok, will change.

tests/baselines_list.txt
gsketefian: Once I make the FV3 namelist mods and the nsoill_out = 9 fix in the chgres_cube namelist, I will rerun these tests (and others) with the HEADS of both the ufs-community and NOAA-GSL repos.  If the tests pass, then we're good and can merge.  If one or both codes fail, then I may merge this PR anyway but with the FV3_HRRR suite turned off (by removing it from valid_param_vals.sh) just so we can get most of the FV3_HRRR suite script infrastructure in and the work doesn't get wasted.

ush/templates/FV3.input.yml
gsketefian: I will change these as you specified in your email and rerun the tests.  Thanks.

scripts/exregional_make_ics.sh
christinaholtNOAA: This did not change with the most recent pushes.

scripts/exregional_make_ics.sh
christinaholtNOAA: Same here.

ush/generate_FV3LAM_wflow.sh
```diff
@@ -533,13 +543,64 @@ Please change one or more of these parameters or provide a value for lsoil
...
+Unexpected value returned for RUC_LSM_name_or_null:
+  RUC_LSM_name_or_null = \"${RUC_LSM_name_or_null}\"
+This variable should be set to either \"${RUC_LSM_name}\" or an empty
+string."
+fi
+
+kice=""
+if [ "${RUC_LSM_is_used}" = "TRUE" ]; then
+  kice="9"
+fi
```
christinaholtNOAA: Setting kice really should be a part of FV3.input.yml for the suites that use RUC LSM, no?

ush/link_fix.sh
```diff
@@ -245,6 +245,12 @@ Creating links in the FIXLAM directory to the grid files..."
     "C*${DOT_OR_USCORE}oro_data.tile${TILE_RGNL}.halo${NH0}.nc" \
     "C*${DOT_OR_USCORE}oro_data.tile${TILE_RGNL}.halo${NH4}.nc" \
         )
+    if [ "${CCPP_PHYS_SUITE}" = "FV3_HRRR" ]; then
+      fns+=( \
+      "C*${DOT_OR_USCORE}oro_data_ss.tile${TILE_RGNL}.halo${NH0}.nc" \
+      "C*${DOT_OR_USCORE}oro_data_ls.tile${TILE_RGNL}.halo${NH0}.nc" \
+           )
+    fi
```
christinaholtNOAA, line 253: Why is this information repeated here? Wasn't it handled in an ex-script?

ush/templates/FV3.input.yml
```diff
@@ -59,6 +65,65 @@ FV3_RRFS_v1beta:
...
+# to the LSM that is used in the SDF (suite definition file).  And lsoil 
+# should just not be set at all (by setting it to !!python/none in 
+# input.nml.FV3 and not setting it in this file at all, probably for all 
+# suites).
+    lsoil: 9
+# As of 20210108, kice should not be set if using the ufs-community
+# version of ufs-weather-model and should be set to 9 if using the 
+# NOAA-GSL version.  Once the changes in NOAA-GSL are merged into 
+# ufs-community, then kice should be set to 9 when using any suite that 
+# includes RUC-LSM.
```
christinaholtNOAA: Is FV3_HRRR available in the community develop branch at all?

christinaholtNOAA (approved):
This looks functional, but I did leave a few comments related to duplication of information and the need to over-engineer the namelist setting for kice.

gsketefian:
@christinaholtNOAA This is actually not the final version of this PR.  I was working on Jet and have changes there that still need to be incorporated when Jet is back up tomorrow.  While working on Hera yesterday, I merged the latest develop into this PR and made a couple of other changes .  After I get the Jet changes in, I'll have to retest.

ush/link_fix.sh
gsketefian: The new code snippet in exregional_make_orog.sh copies the ls and ss files from the location specified in GWD_HRRRsuite_DIR to the "orog" subdirectory under the experiment directory (OROG_DIR); those copy commands will eventually be replaced by a call to Mike Toy's code to generate the ls and ss files for any grid (instead of copying).  On the other hand, this snippet is adding the ls and ss files to the list of files (in this case in OROG_DIR) to which symlinks will be created in the FIXLAM directory.  So it's not a repeat of what's in exregional_make_orog.sh.

I've been thinking about simplifying the treatment of the preprocessing directories by making GRID_DIR, OROG_DIR, and SFC_CLIMO_DIR all the same, e.g set them all to FIXLAM.  That's already what happens in NCO mode, where these three variables are all set to ${FIXLAM_NCO_BASEDIR}/${PREDEF_GRID_NAME}.  But that would have to be a separate PR.

ush/templates/FV3.input.yml
gsketefian: Not currently.  This PR is to add it.

ush/generate_FV3LAM_wflow.sh
gsketefian: Yes, it currently is in this PR, at least for the suites that I know use RUC_LSM.  I added kice=9 to the GSD_SAR and FV3_HRRR suites in FV3.input.yml (see below).  But I think I'm going to change that and set it in this file (as is done in the "settings" variable below) because setting parameters on a suite basis is becoming cumbersome and error prone.  I'm leaning towards doing things on a parameterization basis, i.e. have the workflow check if a certain parameterization is called in the SDF and set variables accordingly.  That way we don't have to keep track of which suite uses which parameterization; the workflow will figure that out for you and set variables accordingly.  

In this case, this code snippet checks whether the SDF is calling the RUC LSM and sets kice accordingly, so we don't have to remember all the suites that use RUC LSM and then make the appropriate changes in FV3.input.yml.  (For example, I can't say I'm sure that GSD_SAR and FV3_HRRR are the only suites that use RUC LSM; I have to go check the SDFs.)  For other variables like thompson_mp_climo_file and nsoill_out in exregional_make_ics.sh and exregional_make_lbcs.sh, this parameterization-based approach will eliminate the need to have the long if-statements in those scripts (repeated several times in each file, no less).  This way, any changes to these parameterization-dependent variables can be made in one place instead of a bunch of places in several scripts.

Actually, there is already a file in the workflow that does this -- set_thompson_mp_fix_files.sh.  So the idea is to do something analogous for other parameterizations, as needed.

scripts/exregional_make_ics.sh
gsketefian: As above, I'm seeing the correct version.  Try reloading the page.  Github has marked this comment as "outdated".

ush/generate_FV3LAM_wflow.sh
```diff
@@ -531,6 +531,7 @@ fi
 #
 # NOTE:
 # May want to remove lsoil from FV3.input.yml (and maybe input.nml.FV3).
+# Also, may want to set lsm here as well depending on SDF_USES_RUC_LSM.
 #
```
JeffBeck-NOAA, line 535: Yep, I'd agree that defining lsoil and lsm values could be more easily handled based on functions such as "SDF_USES_RUC_LSM."  It would remove the potential for errors in the YAML file or input.nml template.

JeffBeck-NOAA:
@gsketefian, thoroughly reviewed this PR again. Looks good to me.

Issue #384, opened 2021-01-02
release/public-v1: Failure of running the workflow tasks on WCOSS Dell and Cray


chan-hoo:
- The workflow tasks do not run correctly on WCOSS Dell and Cray.
- The '--relative' flag of the symlink command in the scripts does not work on WCOSS Cray.
- The module list for each task in 'modulefiles/tasks/wocss_[dell_p3/cray]' needs to be updated.
- A couple of command lines need to be added to the template file 'templates/FV3LAM_wflow.xml' only for WCOSS Dell and Cray.

PR #385, opened 2021-01-02
release/public-v1: Fix to run workflow tasks on WCOSS Dell and Cray

chan-hoo:
## DESCRIPTION OF CHANGES: 
- Removed the '--relative' flag for the symlink command in multiple scripts only for WCOSS Cray.
- Updated the module file list for each task in 'modulefiles/tasks/wcoss_[dell_p3/cray]'.
- Added some command lines to the workflow template xml file ('FV3LAM_wflow_wcoss.xml') for WCOSS Dell or Cray.
- Modified the 'ush/launch_FV3LAM_wflow.sh' script for WCOSS Dell and Cray.

## TESTS CONDUCTED: 
- The build-environment file ('README_wcoss_[dell_p3/cray].txt') were added to SRW App (https://github.com/ufs-community/ufs-srweather-app/issues/79).
- It was tested on WCOSS Dell/Cray as well as Hera with the same sample case described in 'config.community.sh'.
- It worked without any issues on WCOSS Dell/Cray and Hera. 

## ISSUE: 
Fixes issue mentioned in #384
Related to https://github.com/ufs-community/ufs-srweather-app/issues/79

## CONTRIBUTORS: 
@BenjaminBlake-NOAA


scripts/exregional_run_fcst.sh
```diff
@@ -174,7 +179,7 @@ the grid and (filtered) orography files ..."
 cd_vrfy ${run_dir}/INPUT
 
 relative_or_null=""
-if [ "${RUN_TASK_MAKE_GRID}" = "TRUE" ]; then
+if [ "${RUN_TASK_MAKE_GRID}" = "TRUE" ] && [ "${MACHINE}" != "WCOSS_CRAY" ]; then
```
gsketefian, line 182: On WCOSS_CRAY, what is the equivalent of the "--relative" flag to the ln command?  There must be one...  We can use that instead of leaving relative_or_null set to a null string.  That way, the behavior of the workflow will be the same on all machines.

chan-hoo: @gsketefian, since WCOSS_CRAY is running an old version of SuSE Linux (v11.3 or v11.4) which sits on an old version of the Linux kernel, it does not support the "--relative" flag of the symlink command.

BenjaminBlake-NOAA: @gsketefian  According to @RatkoVasic-NOAA , the Cray uses the 2011 version of the ln command which does not have the "--relative" capability.  I'm not sure what the equivalent would be - Ratko do you know?

ush/templates/FV3LAM_wflow_wcoss_cray.xml
```diff
@@ -0,0 +1,428 @@
+{#
```
gsketefian: Instead of creating new template XML files for WCOSS_CRAY and WCOSS_DELL_P3, I think you can use a jinja2 if-statement in FV3LAM_wflow.xml to add the lines you need, something like the below (code not tested though!):
```
{% if machine in ["WCOSS_CRAY"]  %}
    <shared/>
{% elif machine in ["WCOSS_DELL_P3"]  %}
    <memory>2048M</memory>
    <native>-R affinity[core]</native>
{% endif %}
```
Also see line 356 in FV3LAM_wflow.xml.  There is more info on jinja2 templates [here](https://jinja.palletsprojects.com/en/2.11.x/templates/).

scripts/exregional_run_fcst.sh
```diff
@@ -92,7 +92,12 @@ case $MACHINE in
   "WCOSS_CRAY")
     ulimit -s unlimited
     ulimit -a
-    APRUN="aprun -b -j1 -n${PE_MEMBER01} -N24 -d1 -cc depth"
+
```
gsketefian, line 95: Instead of making this change here, it might be better to change PE_MEMBER01 in setup.sh, around line 1864.  That way the correct value for PE_MEMBER01 (i.e. the value actually used in the aprun command) will be recorded in var_defns.sh.  So on line 1864 of setup.sh, add something like this (code untested!):
```
if [ "$MACHINE" -eq "WCOSS_CRAY" ] && [ ${PE_MEMBER01} -lt 24 ]; then
  PE_MEMBER01=24
fi
```

scripts/exregional_run_fcst.sh
gsketefian: @chan-hoo @BenjaminBlake-NOAA @RatkoVasic-NOAA If there is a way to make relative links on WCOSS_CRAY, it would be nice to use it so that the symlinks are to locations within the experiment directory (so they're independent of the root directory of the system).  Otherwise, no big deal.

tests/baseline_configs/config.community_wcoss_dell_p3.sh
```diff
@@ -0,0 +1,34 @@
+RUN_ENVIR="community"
+PREEXISTING_DIR_METHOD="rename"
+
+PREDEF_GRID_NAME="RRFS_CONUS_25km"
+GRID_GEN_METHOD="ESGgrid"
```
gsketefian: Don't need this line for GRID_GEN_METHOD because now, that is determined by PREDEF_GRID_NAME (i.e. it is part of the definition of the predefined grid).  You only need to set GRID_GEN_METHOD if you're specifying a new grid in config.sh.

tests/baseline_configs/config.community_wcoss_dell_p3.sh
```diff
@@ -0,0 +1,34 @@
...
+DATE_LAST_CYCL="20201215"
+CYCL_HRS=( "00" )
+
+EXTRN_MDL_NAME_ICS="FV3GFS"
+EXTRN_MDL_NAME_LBCS="FV3GFS"
+
+FV3GFS_FILE_FMT_ICS="grib2"
+FV3GFS_FILE_FMT_LBCS="grib2"
+
+WFLOW_XML_FN="FV3LAM_wflow_wcoss_dell_p3.xml"
```
gsketefian: You won't need this line if my suggestion about using a jinja2 if-statement in FV3LAM_wflow.xml works (instead of creating new XML templates).

ush/templates/FV3LAM_wflow_wcoss_cray.xml
chan-hoo: @gsketefian, I'll modify the scripts (including others below), and then test the feature branch again.

tests/baseline_configs/config.community_wcoss_dell_p3.sh
```diff
@@ -0,0 +1,34 @@
...
+PPN_MAKE_GRID="1"
+PPN_MAKE_OROG="1"
+
+WTIME_RUN_FCST="01:00:00"
+
+USE_USER_STAGED_EXTRN_FILES="FALSE"
```
gsketefian: If you have this set to "FALSE", then the variables below (e.g. EXTRN_MDL_SOURCE_BASEDIR_ICS) will not be used.  Do you want to use staged files for this test?

The way I handle user-staged external files in the WE2E tests is to set USE_USER_STAGED_EXTRN_FILES to "TRUE" here in the configuration file and then have the WE2E testing script run_experiment.sh check for the value of USE_USER_STAGED_EXTRN_FILES (starting around line 671).  If it is set to "TRUE" (which it will be), then run_experiments.sh will add to the configuration file appropriate (i.e. machine-dependent) values for the remaining variables such as EXTRN_MDL_SOURCE_BASEDIR_ICS.  See, for example, the file 
`regional_workflow/tests/config.grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2.sh
`
Note that all the WE2E tests are meant to be run using the run_experiments.sh script.  That script will take your config.some_test_name.sh file and create from it a final config.sh file that contains all the necessary variables.  Once you call run_experiments.sh and the experiment directory for your WE2E test is created, a copy of that final config.sh can be found in that experiment directory.

So, in summary, if you want to use user-staged external files for this test, you can set USE_USER_STAGED_EXTRN_FILES to "TRUE" here and then set the remaining variables in run_experiments.sh instead of below (most of what you need is already set up for you in run_experiment.sh; you just need to add a case for "WCOSS_DELL_P3" in the appropriate if-statement(s)).

tests/baseline_configs/config.community_wcoss_cray.sh
```diff
@@ -0,0 +1,35 @@
+RUN_ENVIR="community"
```
gsketefian: See my comments for config.community_wcoss_dell_p3.sh.  Same comments apply to this file.

tests/baseline_configs/config.community_wcoss_dell_p3.sh
chan-hoo: @gsketefian, I removed two cases for wcoss here. Since wcoss does not use separate templates, I think we don't need sample cases for wcoss any more.

tests/baseline_configs/config.community_wcoss_dell_p3.sh
chan-hoo: removed.

tests/baseline_configs/config.community_wcoss_dell_p3.sh
chan-hoo: removed

tests/baseline_configs/config.community_wcoss_cray.sh
chan-hoo: removed

tests/baselines_list.txt
```diff
@@ -2,7 +2,7 @@ DOT_OR_USCORE
 GST_release_public_v1
 community_ensemble_008mems
 community_ensemble_2mems
-get_extrn_files_from_hpss_FV3GFS_FV3GFS
+et_extrn_files_from_hpss_FV3GFS_FV3GFS
```
gsketefian: Typo (missing "g")?

chan-hoo: Oops! Sorry. Corrected.

JeffBeck-NOAA (changes requested):
Hi @chan-hoo, we don't have platform-specific config templates in the /ush directory.  Is there a reason the current config.community.sh doesn't work for you?

chan-hoo:
@JeffBeck-NOAA, that script is just for a test on wcoss. Since I made some changes in several scripts, I need to test this feature branch on the dell and cray again. Once the test is complete, I'll remove this script for wcoss.

JeffBeck-NOAA:
> @JeffBeck-NOAA, that script is just for a test on wcoss. Since I made some changes in several scripts, I need to test this feature branch on the dell and cray again. Once the test is complete, I'll remove this script for wcoss.

Thanks @chan-hoo!  I saw it was added to the PR, so I was confused.  Thanks for clarifying!

scripts/exregional_run_fcst.sh
chan-hoo: @gsketefian, I couldn't find any other options for this.

ush/templates/FV3LAM_wflow_wcoss_cray.xml
chan-hoo: Done.

scripts/exregional_run_fcst.sh
chan-hoo: @gsketefian, It didn't work for the Cray. It caused other errors. I think this is the best option for the Cray.

chan-hoo:
@gsketefian @JeffBeck-NOAA , I've tested the modified one. It works well on the wcoss dell and cray. Please review it again.

scripts/exregional_run_fcst.sh
gsketefian: @chan-hoo Ok, no problem.

scripts/exregional_run_fcst.sh
gsketefian: Huh, weird.  Ok, if you prefer this, it's ok with me.

gsketefian:
@chan-hoo This looks good to me.  Couple of things:
1) Can you update the PR message to remove the lines about separate template XMLs and new test config files (since you removed all those)?
2) Can you try running a test on Hera to make sure everything still works there (e.g. to make sure the jinja changes didn't break anything on Hera)?
Thanks.

gsketefian:
@chan-hoo Thanks.  I will approve.
@JeffBeck-NOAA Are there changes that you'd still like @chan-hoo to make?  If not, can he merge?

gsketefian (approved):

JeffBeck-NOAA (approved):

JeffBeck-NOAA:
> @chan-hoo Thanks. I will approve.
> @JeffBeck-NOAA Are there changes that you'd still like @chan-hoo to make? If not, can he merge?

@gsketefian, I'm good with everything!  Going to approve.  I'll also merge, since I don't believe @chan-hoo has permissions.

JeffBeck-NOAA:
@chan-hoo, I added you to the repo with write permission.  Can you try to merge?  Thanks!

PR #396, opened 2021-01-12
Make DT_ATMOS, LAYOUT_X/Y, and BLOCKSIZE user-specifiable when using predefined grids

gsketefian:
## DESCRIPTION OF CHANGES:
* When using a predefined grid, allow the default values of DT_ATMOS, LAYOUT_X, LAYOUT_Y, and BLOCKSIZE for that grid to be overwritten with values specified by the user in the experiment configuration file.
* Add a WE2E test (config.grid_RRFS_CONUS_25km_modify_DT_ATMOS_LAYOUT_XY_BLOCKSIZE.sh) that uses the RRFS_CONUS_25km predefined grid but changes DT_ATMOS, LAYOUT_X, LAYOUT_Y, and BLOCKSIZE from the default values for that grid.
* Update comments.

## TESTS CONDUCTED:
* Ran the new WE2E test (grid_RRFS_CONUS_25km_modify_DT_ATMOS_LAYOUT_XY_BLOCKSIZE) on Hera.  It was successful.


ush/config_defaults.sh
```diff
@@ -833,21 +834,28 @@ WRTCMP_dy=""
...
 #
+# * If PREDEF_GRID_NAME is set to a valid predefined grid name, the grid 
+#   generation method GRID_GEN_METHOD, the (native) grid parameters, and 
+#   the write-component grid parameters are set to predefined values for 
+#   the specified grid, overwriting any settings of these parameters in 
+#   the user-specified experiment configuration file.  In addition, if 
+#   the time step DT_ATMOS and the computational parameters LAYOUT_X, 
+#   LAYOUT_Y, and BLOCKSIZE are not specified in that configuration file, 
+#   they are also set to predefined values for the specified grid.
+#
```
JeffBeck-NOAA, line 845: Isn't the following part of the removed comment still true if PREDEF_GRID_NAME is set to a valid grid: "In addition, if the time step DT_ATMOS and the computational parameters LAYOUT_X, LAYOUT_Y, and BLOCKSIZE are not specified in that configuration file, they are also set to predefined values for the specified grid."  Actually, weren't they previously just overwritten by the predefined values if this was the case?

JeffBeck-NOAA, line 845: @gsketefian, I thought if the user specifies their own values in config.sh, they should **not** be overwritten by the predefined values in set_predef_grid_params.sh, correct?

ush/config_defaults.sh
```diff
@@ -833,21 +834,28 @@ WRTCMP_dy=""
...
-#   method (GRID_GEN_METHOD), grid parameters, time step (DT_ATMOS), 
-#   computational parameters (e.g. LAYOUT_X, LAYOUT_Y), and write component 
-#   parameters set above (and possibly overwritten by values in the user-
-#   specified workflow configuration file) are used.
+#   method GRID_GEN_METHOD, the (native) grid parameters, and the write-
+#   component grid parameters as well as the time step DT_ATMOS and the 
+#   computational parameters LAYOUT_X, LAYOUT_Y, and BLOCKSIZE must be 
+#   set in the user-specified experiment configuration file; otherwise,
+#   the values of these parameters in this default experiment configuration
+#   file will be used.
```
JeffBeck-NOAA: If PREDEF_GRID_NAME is an empty string, this implies that the user is running with their own configured grid.  We might want to clarify that here.

JeffBeck-NOAA (changes requested):

ush/config_defaults.sh
gsketefian: By "grid parameters", I mean just those parameters that define the native and write-component grids.  Those can't be overwritten (otherwise there is no point to having a predefined grid).  The way I'm defining "grid parameters" here doesn't include DT_ATMOS, LAYOUT_X/Y, and BLOCKSIZE because the latter are affected by the grid but are technically independent of it.  These four parameters can be overwritten by the user via config.sh.  That's what this PR does.  That's what the "In addition, ..." sentence is trying to say.

ush/config_defaults.sh
gsketefian: Ok, I modified the comment to clarify.  Let me know what you think.

ush/config_defaults.sh
JeffBeck-NOAA: OK, thanks for the clarification!

ush/config_defaults.sh
JeffBeck-NOAA: Looks good!

JeffBeck-NOAA (approved):

chan-hoo:
@gsketefian @JeffBeck-NOAA,  I think this change makes setting the grid-specific parameters complicated much more. If DT_ATMOS, LAYOUT_X, LAYOUT_Y, BLOCKSIZE are not specified in 'config.sh', their default values in 'config_default.sh' will be used even though I want to use the values specified in 'set_predef_grid_params.sh'. To make it work, you should remove them from 'config_default.sh'. I tested it without setting them in 'config.sh', but I got the values from 'config_default.sh' (NOT from 'set_predef_grid_params.sh').

JeffBeck-NOAA:
@gsketefian, following on from @chan-hoo's comment, is this the expected behavior?  I would think that if a pre-defined grid is selected, AND if the user doesn't set anything explicitly in config.sh, the values in set_predef_grid_params.sh would be used, correct?

JeffBeck-NOAA:
@chan-hoo, when you said you got the values from config_default.sh and not from set_predef_grid_params.sh, can you confirm that you had PREDEF_GRID_NAME defined in your config.sh file?

chan-hoo:
@JeffBeck-NOAA, yes. I used PREDEF_GRID_NAME="RRFS_CONUS_25km"

JeffBeck-NOAA:
> @JeffBeck-NOAA, yes. I used PREDEF_GRID_NAME="RRFS_CONUS_25km"

@gsketefian, is this a bug?

PR #397, opened 2021-01-13
Make DT_ATMOS, LAYOUT_X/Y, and BLOCKSIZE user-specifiable when using predefined grids

gsketefian:
## DESCRIPTION OF CHANGES:
* When using a predefined grid, allow the default values of DT_ATMOS, LAYOUT_X, LAYOUT_Y, and BLOCKSIZE for that grid to be overwritten with values specified by the user in the experiment configuration file.
* Add a WE2E test (config.grid_RRFS_CONUS_25km_modify_DT_ATMOS_LAYOUT_XY_BLOCKSIZE.sh) that uses the RRFS_CONUS_25km predefined grid but changes DT_ATMOS, LAYOUT_X, LAYOUT_Y, and BLOCKSIZE from the default values for that grid.
* Update comments.

## TESTS CONDUCTED:
None, as agreed upon by the DTC-CAM team for the release/public-v1 branch.  Tests were run for the corresponding PR #396 into the develop branch.  See that PR for details.


chan-hoo:
@gsketefian @JeffBeck-NOAA, I think this change complicate setting the grid-specific parameters much more. If DT_ATMOS, LAYOUT_X, LAYOUT_Y, BLOCKSIZE are not specified in 'config.sh', their default values in 'config_default.sh' will be used even though I want to use the values specified in 'set_predef_grid_params.sh'. If you want to set these parameters in 'config.sh', you should remove them from 'set_predef_grid_params.sh' or 'config_default.sh'. I tested it without setting them in 'config.sh', but I got the values from 'config_default.sh' (NOT from 'set_predef_grid_params.sh').

PR #402, opened 2021-01-19
Add MACOS and Generic Linux options for regional_workflow

### LINKED ISSUE #369: Workflow does not work on MacOS due to bash and UNIX utility differences
A number of instances of UNIX utilities and bash syntax do not work on MacOS due to various incompatibilities.

The main problem is that the bash version installed on macos is very old due to licensing issues (version 3.2 from 2007!), and so it does not recognize the newer syntax that is used extensively in the workflow generation and other bash scripts. In addition, some unix utilities for macos differ in functionality from their gnu counterparts.

The specific problems I've found so far are:
 - bash does not allow the ${VARIABLE^^} syntax for capitalizing strings, nor the converse ${VARIABLE,,} syntax for making them lowercase
 - readlink does not accept the "-f" flag
 - sed does not accept the "-r" flag

Among potentially others. I am currently working on getting the generation script working, and from there will fix problems as I go through the entire workflow.

mkavulich:
## DESCRIPTION OF CHANGES: 
This PR adds generic platforms to the regional_workflow, not specific to any one machine, that should allow users to run the ufs-srweather-app on any UNIX-based machine, so long as the NCEPLIBS and other prerequisites have been properly installed without a workflow manager. This can be done using the scripts described in regional_workflow/ush/wrappers/README.md; additional documentation is currently being written.

Users can utilize these options by setting the MACHINE variable in config.sh to either "LINUX" or "MACOS". The LINUX option should allow most users to run the ufs-srweather-app on a generic Linux OS machine. The MACOS option is for MacOS/Darwin operating systems; this needs to be kept separate because the MacOS version of bash is very old, and missing some functionality, as well as several GNU Linux utilities having different functionality and/or names. This capability is not yet well-tested, as there seem to be issues getting some of the UFS_UTILS running on MacOS (segmentation faults). 

## TESTS CONDUCTED: 
"Generic Linux" test was run on Cheyenne machine as a fresh install, including stand-alone install of NCEPLIBS, with no reference to staged or pre-built input files. This was run without rocoto or directly submitting jobs via PBS, but rather the entire workflow was run interactively on a compute node (using the [`qinteractive` command](https://www2.cisl.ucar.edu/resources/computational-systems/cheyenne/running-jobs/submitting-jobs-pbs#basics) which emulated the running of the workflow on a machine with no job scheduler.

On MacOS (Catalina, 10.15.7), was able to successfully generate workflow, and run most of the tasks successfully. The aforementioned issues prevented a full end-to-end test from succeeding, but tests conducted using input files copied from other platforms were successful, so theoretically once the UFS_UTILS issues are solved this should be fully functional.

These tests were all successful based on an older version of regional_workflow (6833b255). However, when I rebased those changes onto the current head, I was met with failures, which is due to namelist changes in the interim. Investigation of these failures is ongoing but I do not believe it should hold up this PR.

## ISSUE (optional): 
Resolves #369


tests/run_experiments.sh
```diff
@@ -417,7 +422,7 @@ Please correct and rerun."
 #
 #-----------------------------------------------------------------------
 #
-  MACHINE="${machine^^}"
+  MACHINE=`echo $machine| tr '[a-z]' '[A-Z]'`
```
christinaholtNOAA: Would you consider using the `$()` subshell notation as it's a bit more user-friendly and less error-prone than using back ticks? I'm just going to draw attention to other uses later on with a comment "back ticks".

ush/bash_utils/define_macos_utilities.sh
```diff
@@ -0,0 +1,33 @@
...
+    export READLINK=greadlink
+    command -v $READLINK >/dev/null 2>&1 || { echo >&2 "For Darwin-based operating systems (MacOS), the '$READLINK' utility is required to run the UFS SRW Application. Reference the User's Guide for more information about platform requirements. Aborting."; exit 1; }
+    export SED=gsed
+    command -v $SED >/dev/null 2>&1 || { echo >&2 "For Darwin-based operating systems (MacOS), the '$SED' utility is required to run the UFS SRW Application. Reference the User's Guide for more information about platform requirements. Aborting."; exit 1; }
+    export DATE_UTIL=gdate
+    command -v $DATE_UTIL >/dev/null 2>&1 || { echo >&2 "For Darwin-based operating systems (MacOS), the '$DATE_UTIL' utility is required to run the UFS SRW Application. Reference the User's Guide for more information about platform requirements. Aborting."; exit 1; }
+    export LN_UTIL=gln
+    command -v $LN_UTIL >/dev/null 2>&1 || { echo >&2 "For Darwin-based operating systems (MacOS), the '$LN_UTIL' utility is required to run the UFS SRW Application. Reference the User's Guide for more information about platform requirements. Aborting."; exit 1; }
```
christinaholtNOAA: Could you make this error message a function and only have to write it once? Here's an untested suggested syntax:
```
darwinerror () {

  utility=$1
  echo >&2 ""For Darwin-based operating systems (MacOS), the ${utility} utility is required to run the UFS SRW Application. Reference the User's Guide for more information about platform requirements. Aborting."
  exit 1
}
```
And call it with

```
command -v $READLINK >/dev/null 2>&1 || darwinerror $READLINK
```

ush/bash_utils/get_elem_inds.sh
```diff
@@ -123,7 +123,7 @@ The arguments to this function are defined as follows:
 #
 #-----------------------------------------------------------------------
 #
-  inds_to_return="${inds_to_return,,}"
+  inds_to_return=`echo $inds_to_return| tr '[A-Z]' '[a-z]'`
```
christinaholtNOAA: back ticks

ush/create_diag_table_files.sh
```diff
@@ -25,7 +25,7 @@ function create_diag_table_files() {
 #
 #-----------------------------------------------------------------------
 #
-  local scrfunc_fp=$( readlink -f "${BASH_SOURCE[0]}" )
+  scrfunc_fp=$( $READLINK -f "${BASH_SOURCE[0]}" )
```
christinaholtNOAA: Why is this no longer a local variable?

ush/create_model_configure_file.sh
```diff
@@ -97,8 +97,11 @@ run directory (run_dir):
 #
 # Set parameters in the model configure file.
 #
-  dot_quilting_dot="."${QUILTING,,}"."
-  dot_print_esmf_dot="."${PRINT_ESMF,,}"."
+
+  QUILTING_LC=`echo $QUILTING| tr '[A-Z]' '[a-z]'`
+  PRINT_ESMF_LC=`echo $PRINT_ESMF| tr '[A-Z]' '[a-z]'`
```
christinaholtNOAA: back ticks

ush/get_extrn_mdl_file_dir_info.sh
```diff
@@ -37,7 +37,7 @@ function get_extrn_mdl_file_dir_info() {
 #
 #-----------------------------------------------------------------------
 #
-  local scrfunc_fp=$( readlink -f "${BASH_SOURCE[0]}" )
+  scrfunc_fp=$( $READLINK -f "${BASH_SOURCE[0]}" )
```
christinaholtNOAA: Why is this not local now?

ush/get_extrn_mdl_file_dir_info.sh
```diff
@@ -213,7 +213,7 @@ fi
 #
 #-----------------------------------------------------------------------
 #
-  anl_or_fcst="${anl_or_fcst^^}"
+  anl_or_fcst=`echo $anl_or_fcst| tr '[a-z]' '[A-Z]'`
```
christinaholtNOAA: back ticks

ush/get_extrn_mdl_file_dir_info.sh
```diff
@@ -298,7 +298,7 @@ fi
 # Get the Julian day-of-year of the starting date and time of the exter-
 # nal model forecast.
 #
-    ddd=$( date --utc --date "${yyyy}-${mm}-${dd} ${hh}:${mn} UTC" "+%j" )
+    ddd=$( DATE_UTIL --utc --date "${yyyy}-${mm}-${dd} ${hh}:${mn} UTC" "+%j" )
```
christinaholtNOAA: There's no `$` here.

ush/get_extrn_mdl_file_dir_info.sh
```diff
@@ -592,11 +593,13 @@ bination of external model (extrn_mdl_name) and analysis or forecast
       sysdir="$sysbasedir"
       ;;
     *)
-      print_err_msg_exit "\
+      if [ "${USE_USER_STAGED_EXTRN_FILES}" != "TRUE" ]; then
+        print_err_msg_exit "\
```
christinaholtNOAA, line 600: Would you mind making this a function, too? It's a bit much that it's repeated so many times!

ush/launch_FV3LAM_wflow.sh
```diff
@@ -65,7 +70,12 @@ scrfunc_dir=$( dirname "${scrfunc_fp}" )
 #-----------------------------------------------------------------------
 #
 exptdir=$( dirname "$0" )
-exptdir=$( readlink -f "$exptdir" )
+if [[ $(uname -s) == Darwin ]]; then
+  command -v greadlink >/dev/null 2>&1 || { echo >&2 "For Darwin-based operating systems (MacOS), the 'greadlink' utility is required to run the UFS SRW Application. Reference the User's Guide for more information about platform requirements. Aborting."; exit 1; }
+  scrfunc_fp=$( greadlink -f "$exptdir" )
+else
+  scrfunc_fp=$( readlink -f "$exptdir" )
+fi
```
christinaholtNOAA, line 78: Why is this needed twice in the same script?

ush/load_modules_run_task.sh
```diff
@@ -152,7 +152,7 @@ jjob_fp="$2"
 #
 #-----------------------------------------------------------------------
 #
-machine=${MACHINE,,}
+machine=`echo $MACHINE| tr '[A-Z]' '[a-z]'`
```
christinaholtNOAA: back ticks.

ush/set_FV3nml_sfc_climo_filenames.sh
```diff
@@ -30,7 +30,7 @@ function set_FV3nml_sfc_climo_filenames() {
 #
 #-----------------------------------------------------------------------
 #
-  local scrfunc_fp=$( readlink -f "${BASH_SOURCE[0]}" )
+  scrfunc_fp=$( $READLINK -f "${BASH_SOURCE[0]}" )
```
christinaholtNOAA: Not local?

ush/setup.sh
```diff
@@ -139,7 +140,7 @@ check_var_valid_value "VERBOSE" "valid_vals_VERBOSE"
 # Set VERBOSE to either "TRUE" or "FALSE" so we don't have to consider
 # other valid values later on.
 #
-VERBOSE=${VERBOSE^^}
+VERBOSE=`echo $VERBOSE| tr '[a-z]' '[A-Z]'`
```
christinaholtNOAA: back ticks here and for several other lines in this file.

ush/setup.sh
```diff
@@ -358,17 +360,20 @@ check_var_valid_value "MACHINE" "valid_vals_MACHINE"
 #
 #-----------------------------------------------------------------------
 #
+NCORES_PER_NODE=""
 case $MACHINE in
 
   "WCOSS_CRAY")
-    NCORES_PER_NODE="24"
```
christinaholtNOAA, line 364: It seems that this means there's not configurable override for running without rocoto on any of the supported platforms.

ush/setup.sh
```diff
@@ -1947,7 +1975,7 @@ NNODES_RUN_FCST=$(( (PE_MEMBER01 + PPN_RUN_FCST - 1)/PPN_RUN_FCST ))
 #-----------------------------------------------------------------------
 #
 mkdir_vrfy -p "$EXPTDIR"
-
+mkdir_vrfy -p "$LOGDIR"
```
christinaholtNOAA, line 2046: Intentional?

ush/source_util_funcs.sh
```diff
@@ -8,7 +8,11 @@ function source_util_funcs() {
 #
 #-----------------------------------------------------------------------
 #
-  local scrfunc_fp=$( readlink -f "${BASH_SOURCE[0]}" )
+  if [[ $(uname -s) == Darwin ]]; then
+    scrfunc_fp=$( greadlink -f "${BASH_SOURCE[0]}" )
+  else
+    scrfunc_fp=$( readlink -f "${BASH_SOURCE[0]}" )
+  fi
```
christinaholtNOAA, line 15: local?

christinaholtNOAA (approved):
Whoa that's a hefty set of changes! Looks pretty good, and will be quite helpful as we try to move RRFS to AWS. Most of the comments are repeats fo a couple of main ones, and those are really just suggestions.

ush/setup.sh
mkavulich: Yes, LOGDIR is a subdirectory of EXPTDIR that also needs to be created, so I figured 2 birds, 1 stone with the `-p` option. See 14f51b2fc specifically for more info.

tests/run_experiments.sh
mkavulich: Done.

ush/create_diag_table_files.sh
mkavulich: Copy-paste errors 🤦  Fixed.

ush/setup.sh
mkavulich: I think it is correct to require the pre-configured setup use rocoto. If people want to use a custom setup with a different (or no) workflow manager then they are no longer using the platform as intended/supported. 

Users can always use the new generic LINUX setting to set up their own custom environment on an official platform if they wish. And for users that are using a different platform but would still like to use rocoto, we will provide documentation that goes over the modifications needed to add a new platform.

christinaholtNOAA: Thanks for the context. :)

ush/get_extrn_mdl_file_dir_info.sh
mkavulich: It's weird, as far as I can tell this *should* be causing failures on Jet, but it's not. Fixed regardless.

tests/run_experiments.sh
```diff
@@ -9,7 +9,12 @@
 #
 #-----------------------------------------------------------------------
 #
-scrfunc_fp=$( readlink -f "${BASH_SOURCE[0]}" )
+if [[ $(uname -s) == Darwin ]]; then
```
gsketefian, line 12: Why not just change "readlink" to "$READLINK"?  I guess because this is an independent separate script in which READLINK may not yet be defined, or var_defns.sh is not yet sourced?  But you do use READLINK later.

Assuming you can't use $READLINK for whatever reason, in this if-statement is using uname better than just checking whether $MACHINE is set to "MACOS"?

tests/run_experiments.sh
```diff
@@ -757,7 +763,8 @@ boundary conditions specification interval (LBC_SPEC_INTVL_HRS):
       fi
     elif [ "${EXTRN_MDL_NAME_LBCS}" = "HRRR" ] || \
          [ "${EXTRN_MDL_NAME_LBCS}" = "RAP" ]; then
-      EXTRN_MDL_FILES_LBCS=( "${EXTRN_MDL_FILES_LBCS[@]/#/${EXTRN_MDL_NAME_LBCS,,}.out.for_f}" )
+      EXTRN_MDL_NAME_LBCS_LC=$(echo $EXTRN_MDL_NAME_LBCS| tr '[A-Z]' '[a-z]')
```
gsketefian: Can you make the variable name EXTRN_MDL_NAME_LBCS_LC lower case since the convention I try to stick to is lowercase for local variables and uppercase for global workflow variables?  Thx.

tests/run_experiments.sh
```diff
@@ -417,7 +422,7 @@ Please correct and rerun."
 #
 #-----------------------------------------------------------------------
 #
-  MACHINE="${machine^^}"
+  MACHINE=$(echo $machine| tr '[a-z]' '[A-Z]')
```
gsketefian: Can you make a note above this line that the reason we don't use ${machine^^} is because of the MACOS case?  That way there is less of a chance that someone will come around and change it back, only to find out it doesn't work with MACOS!

ush/bash_utils/define_macos_utilities.sh
```diff
@@ -0,0 +1,45 @@
...
+    command -v $SED >/dev/null 2>&1 || darwinerror $SED
+    export DATE_UTIL=gdate
+    command -v $DATE_UTIL >/dev/null 2>&1 || darwinerror $DATE_UTIL
+    export LN_UTIL=gln
+    command -v $LN_UTIL >/dev/null 2>&1 || darwinerror $LN_UTIL
+  else
+    export READLINK=readlink
+    export SED=sed
+    export DATE_UTIL=date
+    export LN_UTIL=ln
```
gsketefian, line 43: Why not use LN instead of LN_UTIL (analogous to SED/sed)?  I understand you wouldn't want to do that with DATE_UTIL since using DATE may lead to confusion with a forecast date variable.

scripts/exregional_make_grid.sh
```diff
@@ -433,11 +441,19 @@ uniform cubed-sphere grid equivalent resolution returned with nonzero exit
...
+The utility 'ncdump' was not found in the environment. Be sure to add the
+netCDF 'bin/' directory to your PATH."
+fi
+
 # Make the following (reading of res_equiv) a function in another file
 # so that it can be used both here and in the exregional_make_orog.sh
 # script.
 res_equiv=$( ncdump -h "${grid_fp}" | \
-             grep -o ":RES_equiv = [0-9]\+" | grep -o "[0-9]" ) || \
+             grep -o ":RES_equiv = [0-9]\+" | sed 's/[^0-9]*//g' ) || \
```
JeffBeck-NOAA: Does the "sed" in this line need to be "$SED", similar to the changes you made in the other j-jobs and ex-scripts?

ush/bash_utils/print_input_args.sh
```diff
@@ -174,7 +174,8 @@ have been set as follows:
 #
 #-----------------------------------------------------------------------
 #
-  if [ ! -v VERBOSE ]; then
+  set +u
+  if [ ! -z {$VERBOSE+x} ]; then
```
JeffBeck-NOAA: Are there unset variables somewhere that require "set +u"?

ush/generate_FV3LAM_wflow.sh
```diff
@@ -19,7 +19,11 @@ function generate_FV3LAM_wflow() {
 #
 #-----------------------------------------------------------------------
 #
-local scrfunc_fp=$( readlink -f "${BASH_SOURCE[0]}" )
+  if [[ $(uname -s) == Darwin ]]; then
+    scrfunc_fp=$( greadlink -f "${BASH_SOURCE[0]}" )
+  else
+    scrfunc_fp=$( readlink -f "${BASH_SOURCE[0]}" )
+  fi
```
JeffBeck-NOAA, line 26: Does "readlink" here need to be "$READLINK"?

ush/bash_utils/print_input_args.sh
gsketefian: I think the $ should be outside of the {...} and the ! should be removed, i.e. replace with this:
`  if [ -z ${VERBOSE+x} ]; then`
Then I don't think you'll need the `set +u`.  Here's a test script I wrote to check this (comment the different lines to test various combinations):
```
#!/bin/bash

#VERBOSE=""
#VERBOSE="abc"

#if [ ! -v VERBOSE ]; then
if [ -z ${VERBOSE+x} ]; then
  echo "VERBOSE is NOT defined!!"
else
  echo "VERBOSE is defined!!"
fi
```

ush/get_extrn_mdl_file_dir_info.sh
```diff
@@ -213,7 +213,7 @@ fi
 #
 #-----------------------------------------------------------------------
 #
-  anl_or_fcst="${anl_or_fcst^^}"
+  anl_or_fcst=$(echo $anl_or_fcst| tr '[a-z]' '[A-Z]')
   valid_vals_anl_or_fcst=( "ANL" "FCST" )
```
JeffBeck-NOAA, line 217: @gsketefian, does this also require a comment on why ${anl_or_fcst^^} isn't used due to MacOS?

ush/cmp_expt_to_baseline.sh
```diff
@@ -28,7 +28,11 @@ module load nccmp
 #
 #-----------------------------------------------------------------------
 #
-scrfunc_fp=$( readlink -f "${BASH_SOURCE[0]}" )
+if [[ $(uname -s) == Darwin ]]; then
```
gsketefian: We should probably remove this script from the workflow since it's ancient and is almost guaranteed to not work.  I haven't used it in more than a year, probably two.  We should keep it in the develop branch though in case one day we update it to work.

Of course unless someone else has been using/updating it...

ush/launch_FV3LAM_wflow.sh
JeffBeck-NOAA, line 78: readlink to $READLINK, or OK?

ush/load_modules_run_task.sh
```diff
@@ -152,7 +152,7 @@ jjob_fp="$2"
 #
 #-----------------------------------------------------------------------
 #
-machine=${MACHINE,,}
+machine=$(echo $MACHINE| tr '[A-Z]' '[a-z]')
 env_fn="build_${machine}_${COMPILER}.env"
```
JeffBeck-NOAA, line 160: @gsketefian, add a comment as above for ^^, but now for ,,?

ush/make_grid_mosaic_file.sh
```diff
@@ -27,7 +27,11 @@ function make_grid_mosaic_file() {
 #
 #-----------------------------------------------------------------------
 #
-  local scrfunc_fp=$( readlink -f "${BASH_SOURCE[0]}" )
+  if [[ $(uname -s) == Darwin ]]; then
+    scrfunc_fp=$( greadlink -f "${BASH_SOURCE[0]}" )
+  else
+    scrfunc_fp=$( readlink -f "${BASH_SOURCE[0]}" )
+  fi
```
JeffBeck-NOAA, line 34: readlink is OK, shouldn't be $READLINK?

ush/cmp_rundirs_ncfiles.sh
```diff
@@ -75,8 +75,13 @@ function cmp_ncfiles_one_dir() {
 #
 #set -x
 
-rundir1="$( readlink -f $1 )"
-rundir2="$( readlink -f $2 )"
+if [[ $(uname -s) == Darwin ]]; then
```
gsketefian: Same for this script as cmp_expt_to_baseline.sh (i.e. probably remove it).

ush/set_extrn_mdl_params.sh
```diff
@@ -49,25 +49,12 @@ if [ "${RUN_ENVIR}" = "nco" ]; then
...
-      ;;
-    "HERA")
-      EXTRN_MDL_SYSBASEDIR_ICS=""
-      ;;
-    "ORION")
-      EXTRN_MDL_SYSBASEDIR_ICS=""
-      ;;
-    "JET")
-      EXTRN_MDL_SYSBASEDIR_ICS=""
-      ;;
```
JeffBeck-NOAA, line 70: Does it matter that EXTRN_MDL_SYSBASEDIR_ICS is undefined if these are removed?

ush/source_util_funcs.sh
```diff
@@ -8,7 +8,11 @@ function source_util_funcs() {
 #
 #-----------------------------------------------------------------------
 #
-  local scrfunc_fp=$( readlink -f "${BASH_SOURCE[0]}" )
+  if [[ $(uname -s) == Darwin ]]; then
+    local scrfunc_fp=$( greadlink -f "${BASH_SOURCE[0]}" )
+  else
+    local scrfunc_fp=$( readlink -f "${BASH_SOURCE[0]}" )
+  fi
   local scrfunc_fn=$( basename "${scrfunc_fp}" )
```
JeffBeck-NOAA, line 16: readlink is OK instead of $READLINK?

ush/setup.sh
```diff
@@ -139,7 +140,7 @@ check_var_valid_value "VERBOSE" "valid_vals_VERBOSE"
 # Set VERBOSE to either "TRUE" or "FALSE" so we don't have to consider
 # other valid values later on.
 #
-VERBOSE=${VERBOSE^^}
+VERBOSE=$(echo $VERBOSE| tr '[a-z]' '[A-Z]')
 if [ "$VERBOSE" = "TRUE" ] || \
```
JeffBeck-NOAA, line 144: @gsketefian, is there some high-level location where @mkavulich can just add a single comment for all these ^^ or ,, changes, so he doesn't have to do it everywhere?

ush/create_model_configure_file.sh
```diff
@@ -97,8 +97,11 @@ run directory (run_dir):
 #
 # Set parameters in the model configure file.
 #
-  dot_quilting_dot="."${QUILTING,,}"."
-  dot_print_esmf_dot="."${PRINT_ESMF,,}"."
+
+  QUILTING_LC=$(echo $QUILTING| tr '[A-Z]' '[a-z]')
```
gsketefian: Make QUILTING_LC and PRINT_ESMF_LC lower case since they are local variables (convention I've tried to stick to for a long time).

ush/generate_FV3LAM_wflow.sh
```diff
@@ -19,7 +19,11 @@ function generate_FV3LAM_wflow() {
 #
 #-----------------------------------------------------------------------
 #
-local scrfunc_fp=$( readlink -f "${BASH_SOURCE[0]}" )
+  if [[ $(uname -s) == Darwin ]]; then
```
gsketefian, line 22: Put "local" in front of scrfunc_fp on both lines.

ush/launch_FV3LAM_wflow.sh
```diff
@@ -65,7 +70,12 @@ scrfunc_dir=$( dirname "${scrfunc_fp}" )
 #-----------------------------------------------------------------------
 #
 exptdir=$( dirname "$0" )
-exptdir=$( readlink -f "$exptdir" )
+if [[ $(uname -s) == Darwin ]]; then
+  command -v greadlink >/dev/null 2>&1 || { echo >&2 "For Darwin-based operating systems (MacOS), the 'greadlink' utility is required to run the UFS SRW Application. Reference the User's Guide for more information about platform requirements. Aborting."; exit 1; }
+  scrfunc_fp=$( greadlink -f "$exptdir" )
```
gsketefian: scrfunc_fp on the left-hand sides of the equal signs on lines 75 and 77 should be changed to exptdir.

Also, since this if-statement (with the long message in it) is repeated in many scripts, it would be nice to make it a function and call it in a subshell, e.g. 
`exptdir=$( generic_readlink "$exptdir" )`
I guess this is kind of like the READLINK variable you use elsewhere...

ush/load_modules_run_task.sh
```diff
@@ -152,7 +152,7 @@ jjob_fp="$2"
 #
 #-----------------------------------------------------------------------
 #
-machine=${MACHINE,,}
+machine=$(echo $MACHINE| tr '[A-Z]' '[a-z]')
```
gsketefian: Thinking about this lowercase/uppercase stuff some more, would it be possible to define a function to do these, like:
`machine=$( to_lowercase $MACHINE )`
Then the use of ,, vs. tr ... can be set in one place (in the `to_lowercase` function).  Same for uppercase.

ush/make_grid_mosaic_file.sh
```diff
@@ -27,7 +27,11 @@ function make_grid_mosaic_file() {
 #
 #-----------------------------------------------------------------------
 #
-  local scrfunc_fp=$( readlink -f "${BASH_SOURCE[0]}" )
+  if [[ $(uname -s) == Darwin ]]; then
```
gsketefian, line 30: Can you add "local" to before "srfunc_fp=" on both lines?

Also, do you want that long error message about greadlink that you have in other scripts here as whell?

ush/set_extrn_mdl_params.sh
```diff
@@ -131,33 +109,34 @@ else
...
 #
-# If EXTRN_MDL_SYSBASEDIR_ICS has not been set (not even to a null string), 
-# print out an error message and exit.
+# If EXTRN_MDL_SYSBASEDIR_ICS has not been set, check if user has specified their own 
+# input data or HPSS_QUEUE is set (in order to pull data from HPSS at runtime). If 
+# none of these avenues for getting data are available, print out an error message and 
+# exit.
 #
-if [ -z "${EXTRN_MDL_SYSBASEDIR_ICS+x}" ]; then
+if [ "$USE_USER_STAGED_EXTRN_FILES" = "FALSE" -a -z "${EXTRN_MDL_SYSBASEDIR_ICS}" -a -z "QUEUE_HPSS" ]; then
```
gsketefian: Forgot $ before QUEUE_HPSS.

For long variable names, especially those that have underscores in them, can we use curly braces when referencing them?  E.g. ${USE_USER_STAGED_EXTRN_FILES} instead of $USE_USER_STAGED_EXTRN_FILES.  Helps me when reading the code.  Thanks.

ush/set_extrn_mdl_params.sh
```diff
@@ -298,17 +252,21 @@ else
...
 #
-# If EXTRN_MDL_SYSBASEDIR_LBCS has not been set (not even to a null string), 
-# print out an error message and exit.
+# If EXTRN_MDL_SYSBASEDIR_LBCS has not been set, check if user has specified their own
+# input data or HPSS_QUEUE is set (in order to pull data from HPSS at runtime). If 
+# none of these avenues for getting data are available, print out an error message and 
+# exit.
 #
-if [ -z "${EXTRN_MDL_SYSBASEDIR_LBCS+x}" ]; then
+if [ "$USE_USER_STAGED_EXTRN_FILES" = "FALSE" -a -z "${EXTRN_MDL_SYSBASEDIR_LBCS}" -a -z "QUEUE_HPSS" ]; then
```
gsketefian: $ before QUEUE_HPSS plus use {} around variables with underscores.

ush/source_util_funcs.sh
```diff
@@ -8,7 +8,11 @@ function source_util_funcs() {
 #
 #-----------------------------------------------------------------------
 #
-  local scrfunc_fp=$( readlink -f "${BASH_SOURCE[0]}" )
+  if [[ $(uname -s) == Darwin ]]; then
```
gsketefian, line 11: Do you want that long error message about greadlink that you have in other scripts here as well?

ush/setup.sh
gsketefian: I suggested this somewhere up top in another file, i.e. make functions to_uppercase and to_lowercase.  He'll still have to change the ^^ and ,, to to_uppercase and to_lowercase, but at least the actual code that is executed will be in one place (in those functions).

ush/setup.sh
```diff
@@ -1995,7 +2022,7 @@ NNODES_RUN_FCST=$(( (PE_MEMBER01 + PPN_RUN_FCST - 1)/PPN_RUN_FCST ))
 #-----------------------------------------------------------------------
 #
 mkdir_vrfy -p "$EXPTDIR"
-
+mkdir_vrfy -p "$LOGDIR"
```
gsketefian, line 2046: Can you state in the comment above why the log directory needs to be made here?  Because it needs to be created sooner if running without a workflow manager, if I understand right.  Thx!

ush/source_util_funcs.sh
gsketefian, line 11: Try using a function maybe?  Not sure if it will work, since BASH_SOURCE[0] may then change within the function...  
Oh wait, you can first get BASH_SOURCE[0] here, then pass it to a generic readlink function.  I think that'll work.

tests/run_experiments.sh
mkavulich: You're correct, this is a bootstrap problem related to how functions and variables are loaded/defined. In several scripts we need to use readlink before ush/source_util_funcs.sh is sourced, so we do not yet have access to all of those utility functions which include the MacOS-specific utility names

ush/bash_utils/define_macos_utilities.sh
mkavulich: It was just an abundance of caution; with such a short variable name I was worried it wouldn't be clear what the variable was and it was also harder to search to make sure that variable wasn't used elsewhere. Also, these variables were added at different times and at that point I had already used DATE_UTIL so I stuck to that convention.

ush/bash_utils/print_input_args.sh
mkavulich: Thanks for catching this...The problem here was that MacOS bash does not support the -v option. But I could not figure out why I kept getting errors when trying the `VERBOSE+x` solution; turns out that the $ in the wrong place will do that!

As a side note, it seems to me like this logic is unnecessary since VERBOSE is set to a default value before this utility is ever used. But I suppose it's okay to keep out of an abundance of caution for future-proofing.

ush/cmp_expt_to_baseline.sh
mkavulich: I have definitely not used it. It looks like @JeffBeck-NOAA made a change to it in August, but I assume he was doing the same as me and grepping lines that needed to be changed without noticing it was an outdated stand-alone script. If you're the only user and you'd like to remove it from the release branch I'm all for it.

scripts/exregional_make_grid.sh
mkavulich: It doesn't need to be gsed in this case because it isn't invoking any functionality that the MacOS native sed does not have. I could change it for consistency if people think I should, thinking about it now that might be a good idea.

ush/launch_FV3LAM_wflow.sh
mkavulich: Because of the bootstrap problem I mentioned above it would really require a fundamental restructuring of all of the scripts in order to avoid all of these Darwin if statements. I have avoided as many as possible with the bash_utils/define_macos_utilities.sh functionality.

ush/cmp_expt_to_baseline.sh
JeffBeck-NOAA: @mkavulich, yep, I'm pretty sure that's what I did.  @gsketefian, was this script replaced by another, or are we just not comparing output for now?  One of our deliverables for the next AOP (UFS SRW App community support) is to advance the regression testing.  It would seem like comparing netcdf output is the next logical step (a true regression test), so maybe it'd be worth keeping.

ush/set_extrn_mdl_params.sh
mkavulich: EXTRN_MDL_SYSBASEDIR_ICS is now set to an empty string at the beginning of the big if block; this just condenses all the repetitive blank assignments

ush/set_extrn_mdl_params.sh
mkavulich: Can do, and thanks for catching that bad variable reference

ush/set_extrn_mdl_params.sh
mkavulich: Done

ush/make_grid_mosaic_file.sh
mkavulich: @gsketefian This may be a bad assumption for all platforms, but for all of my MacOS testing these if statements produced "command not found" errors, but the script continued through into the logic that eventually outputs the full verbose error message. I thought that was a good compromise to prevent cluttering these several if statements with long error messages.

ush/setup.sh
mkavulich: If using rocoto this directory gets created automatically, you're correct that this command is needed for platforms without a workflow manager, but as far as I can tell it is only needed at runtime (so creating it here is sufficient).

mkavulich:
@gsketefian, @JeffBeck-NOAA, and @christinaholtNOAA, I believe I have addressed all comments. Just to clarify, because this was mentioned in a few places and I only addressed it once: there are a few instances where, because of the order that different scripts are sourced with the appropriate functions, it was impossible to remove *all* if statements regarding the utility name differences between MacOS and Linux, but I was able to consolidate the vast majority using the `ush/bash_utils/define_macos_utilities.sh` function.

I'm running a final round of tests tests after these last set of changes, but after that I believe this PR will be ready for approval and merge, barring any final comments or concerns.

JeffBeck-NOAA (approved):

ush/launch_FV3LAM_wflow.sh
gsketefian: Ok, that's fine.  But there is still the bug that scrfunc_fp on the left-hand sides of the equal signs on lines 75 and 77 should be changed to exptdir.  I'm surprised this script works with this bug -- or maybe hasn't been tested...

ush/make_grid_mosaic_file.sh
gsketefian: Ok.  The reason I initially introduced these scrfunc_* variables is to be able to attribute error messages to where they're coming from, kind of like a traceback.  I've found out since then that there basically is a way to do traceback in bash, although I don't know if the MACOS version of bash has it.  We'll have to think about whether it would be better to remove these in the develop branch, but things are fine for now for the release branch.

ush/setup.sh
gsketefian: Ok, I was just asking that you add a note in the comment to clarify why the log directory is created here, mostly as a reminder to developers like me...

gsketefian:
@mkavulich Once this is working on MACOS, do you plan on retesting on Hera and/or Cheyenne to make sure nothing broke on those platforms?

ush/setup.sh
mkavulich: Got it, I misunderstood your comment initially. Added.

mkavulich:
@gsketefian Sorry about the problem in ush/launch_FV3LAM_wflow.sh, I pushed the fix just now. The script has not been tested on MacOS because our current method of running the workflow on MacOS is only via the stand-alone scripts in ush/wrappers. But I still thought it best to update this script for future-proofing. 

I did test it on both Cheyenne and Hera to make sure no existing capabilities had been broken, and it did indeed work which I initially found confusing after seeing this obvious mistake in my changes. It seems like it still works because $exptdir has already been set to ".", and the entire launch script is run locally, so it never needs the full exptdir path. 

Regardless, it has now been fixed. I also had to resolve some merge conflicts related to the Gaea changes, so I will run another round of tests on Hera, Cheyenne, and MacOS today before merging.

ush/make_grid_mosaic_file.sh
mkavulich: 👍

mkavulich:
Some final notes for posterity:

 - The MacOS capability will not work until https://github.com/NOAA-EMC/UFS_UTILS/pull/245 is merged (make_orog will fail)
 - The stand-alone capability on MacOS is not yet documented in the Users guide; this Google doc has some instructions that should be helpful: https://docs.google.com/document/d/1CvCW3wl3Cn2UWkp7Fl3Gj7GsqRhP8Ck1uxPG7sJLitw/edit?usp=sharing
 - The release branch end-to-end tests were run on Hera (/scratch2/BMC/det/kavulich/workdir/PR_402/expt_dirs/) and Jet (/mnt/lfs1/HFIP/dtc-hurr/kavulich/PR_402/expt_dirs) on the final version of this PR; all passed.

PR #381, opened 2020-12-21
Enhancements to Python plotting scripts

BenjaminBlake-NOAA:
## DESCRIPTION OF CHANGES: 
Output files generated by plot_allvars and plot_allvars_diff are now written out to the expt_dir rather than the Python directory under regional_workflow/ush/Python.  In addition, logic has been added to generate plots for a defined set of forecast hours.  The user must now specify the starting/ending forecast hours as well as the forecast hour increment.

## TESTS CONDUCTED: 
These changes were tested on both Hera and Jet.

## ISSUE (optional): 
Resolves issue #376 : https://github.com/NOAA-EMC/regional_workflow/issues/376

## CONTRIBUTORS (optional): 
Ben Blake, David Wright

PR #412, opened 2021-01-21
Add submission scripts for python scripts for hera specifically

jwolff-ncar:
## DESCRIPTION OF CHANGES: 
With the updates to the python scripts to run multiple forecast hours and several variables, it is necessary to submit as a batch script. These are sample slurm submission scripts.

## TESTS CONDUCTED: 
Tested and run on Hera only at this point. These need to be ported and expanded for other machines. (For example, should also create sample qsub scripts for Cheyenne prior to the release)

We will also want to include this in the release branch.

## ISSUE (optional): 
Related to https://github.com/NOAA-EMC/regional_workflow/pull/381

## CONTRIBUTORS (optional): 
@BenjaminBlake-NOAA @dmwright526 @christinaholtNOAA

### DIFF
--- a/ush/Python/sq_job.sh
+++ b/ush/Python/sq_job.sh
@@ -1,12 +1,18 @@
 #!/bin/sh
-#SBATCH -e /scratch2/BMC/det/jwolff/GST/ufs-srweather-app-gst/regional_workflow/ush/Python/plot_allvars.log
-#SBATCH --account=gsd-fv3
+
+#SBATCH --account=an_account
 #SBATCH --qos=batch
 #SBATCH --ntasks=4
 #SBATCH --time=0:20:00
 #SBATCH --job-name="plot_allvars"
+#SBATCH --out=plot_allvars.out
 
-cd /scratch2/BMC/det/jwolff/GST/ufs-srweather-app/regional_workflow/ush/Python
+# Prior to submitting the script the following environment variables
+# must be set using export or setenv
+# HOMErrfs=/path-to/ufs-srweather-app/regional_workflow
+# EXPTDIR=/path-to/expt_dirs/your_experiment
+
+cd ${HOMErrfs}/ush/Python
 set -x
 . /apps/lmod/lmod/init/sh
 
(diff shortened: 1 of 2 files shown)


ush/Python/sq_job.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/sh
+#SBATCH -e /scratch2/BMC/det/jwolff/GST/ufs-srweather-app-gst/regional_workflow/ush/Python/plot_allvars.log
+#SBATCH --account=gsd-fv3
+#SBATCH --qos=batch
+#SBATCH --ntasks=4
+#SBATCH --time=0:20:00
+#SBATCH --job-name="plot_allvars"
+
+cd /scratch2/BMC/det/jwolff/GST/ufs-srweather-app/regional_workflow/ush/Python
```
christinaholtNOAA: Can this be a relative path to this file?

ush/Python/sq_job.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/sh
```
christinaholtNOAA, line 1: Most scripts in this repo are bash. Could this change to be consistent with the others?

ush/Python/sq_job.sh
```diff
@@ -0,0 +1,47 @@
...
+#module load miniconda3
+#conda activate pygraf
+
+############
+# Path to shape files
+############
+#Hera: /scratch2/NCEPDEV/fv3-cam/Chan-hoo.Jeon/tools/NaturalEarth/ 
+#Jet: 
+#Orion: /home/chjeon/tools/NaturalEarth/
+#Gaea:
```
christinaholtNOAA, line 50: Are these shape files needed with the Python environments we have?

ush/Python/sq_job.sh
```diff
@@ -0,0 +1,47 @@
...
+
+############
+# Path to shape files
+############
+#Hera: /scratch2/NCEPDEV/fv3-cam/Chan-hoo.Jeon/tools/NaturalEarth/ 
+#Jet: 
+#Orion: /home/chjeon/tools/NaturalEarth/
+#Gaea: 
+
+python plot_allvars.py 2019061500 6 48 6 /scratch2/BMC/det/jwolff/GST/expt_dirs/test_CONUS_25km_GFSv15p2 /scratch2/NCEPDEV/fv3-cam/Chan-hoo.Jeon/tools/NaturalEarth
```
christinaholtNOAA: A comment on command line use here would be super helpful. Also, consider removing the personal paths.

ush/Python/sq_job_diff.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/sh
```
christinaholtNOAA, line 1: Would it be worth just adding a commented section in the sq_job.sh script to handle the diffs since most of this script is a repeat of the other?

ush/Python/sq_job.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/sh
+#SBATCH -e /scratch2/BMC/det/jwolff/GST/ufs-srweather-app-gst/regional_workflow/ush/Python/plot_allvars.log
+#SBATCH --account=gsd-fv3
```
christinaholtNOAA: It may be worth putting a non-account string here.

I'm also curious if these are the SBATCH directives that are needed for all the platforms you include below (I don't submit jobs to those very often).

christinaholtNOAA (approved):
Jamie, these submit scripts are super helpful! Just leaving a few comments that might make things smoother for end users.

ush/Python/sq_job.sh
jwolff-ncar: Thanks for calling me out on being lazy! I have fixed as many hard coded paths as possible and will be checking in a new version shortly.

ush/Python/sq_job.sh
jwolff-ncar: I based this off of the scripts in ush/wrapper so I would prefer to leave it the same as those unless there are other dissensions.

ush/Python/sq_job.sh
jwolff-ncar: I assume so as it is a required command line argument but I don't actually know the answer. @BenjaminBlake-NOAA do you know?

ush/Python/sq_job_diff.sh
jwolff-ncar: I think I will leave this for now but we can talk about that for future enhancements.

ush/Python/sq_job.sh
jwolff-ncar: Good question on other platforms - I have never run on anything other than hera so I do not know and hope others can help test/port to other platforms.

ush/Python/sq_job.sh
BenjaminBlake-NOAA: Yes, they are cartopy shapefiles and are used for plotting the background.

ush/Python/sq_job.sh
jwolff-ncar: Good idea on the command line usage statement. Thanks!

jwolff-ncar:
Thanks for all the great comments. I have committed an updated version of the scripts and am ready for round 2 of reviews!

BenjaminBlake-NOAA (approved):
I tested the latest changes on Hera and they worked well for me!  I'll go ahead and approve this PR.  @jwolff-ncar  after these changes are finalized, I can add them to PR 415 for adding the Python script enhancements to the release branch.

jwolff-ncar:
Thanks, @BenjaminBlake-NOAA! I will merge these now. They may need to be updated as we port to additional platforms but let's get them in there as is for now.

christinaholtNOAA:
@jwolff-ncar Thanks for taking the time to respond to my questions and comments. It helps a lot with the context. This looks great and it will be super helpful.

PR #416, opened 2021-01-26
Improve the way LSM-related parameters are set.

gsketefian:
## NOTE:
PR #[94](https://github.com/ufs-community/ufs-srweather-app/pull/94) into ufs-srweather-app should be merged right after this PR.

### Main Changes:
* Add the new function check_ruc_lsm.sh that reads in the SDF to check whether it uses the RUC LSM.  If so, it sets the local variable sdf_uses_ruc_lsm to "TRUE", otherwise to "FALSE".  It then sets the environment variable whose name is specified by the input argument output_varname_sdf_uses_ruc_lsm to the value of sdf_uses_ruc_lsm.
* In setup.sh, introduce new workflow variable SDF_USES_RUC_LSM by passing its name as an argument to the new function check_ruc_lsm.sh.  This gets set to "TRUE" if the physics suite uses the RUC LSM and "FALSE" otherwise.
* In exregional_make_ics.sh, set the chgres_cube namelist variable nsoill_out (which is the number of soil levels to have in the NetCDF file that chgres_cube generates) by uisng an if-statement that checks the new workflow variable SDF_USES_RUC_LSM as well as the name of the external model used for ICs.  This if-statemnt replaces the settings of nsoill_out in the external-model-dependent case-statement and, depending on model, the physics-suite-dependent if-statements within that case statement.  This simplifies the script because we no longer need to remember which suites use the RUC LSM and which don't.
* In generate_FV3LAM_wflow.sh, set the number of ice levels in the FV3 input file according to whether the RUC_LSM is used in the physics suite (i.e. by checking the new variable SDF_USES_RUC_LSM).

### Improvements:
* Change the name of the workflow variable THOMPSON_MP_USED to SDF_USES_THOMPSON_MP to be consistent with the new varable SDF_USES_RUC_LSM.
* In set_thompson_mp_fix_files.sh, change the local variable thompson_mp_used to sdf_uses_thompson_mp (to be consistent with the change in the corresponding workflow variable SDF_USES_THOMPSON_MP).

## TESTS CONDUCTED:
Ran the following WE2E tests on Hera using the HEAD (hash #ea8a7aa) of the authoritative ufs-community/ufs-weather-model repo:

* grid_RRFS_CONUS_13km_ics_HRRR_lbcs_RAP_suite_GSD_SAR_2020020800
* grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GSD_SAR
* grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_GSD_SAR_2020020800
* grid_RRFS_CONUS_25km_modify_DT_ATMOS_LAYOUT_XY_BLOCKSIZE
* grid_RRFS_CONUS_3km_ics_HRRR_lbcs_RAP_suite_GSD_SAR_2020020800
* nco_CONUS_25km_GFDLgrid
* nco_RRFS_CONUS_25km_HRRR_RAP
* suite_FV3_GSD_v0

All tests were successful.  Note that the three tests whose names end with "_2020020800" are not in the workflow, but they are generated by taking the respective WE2E tests in the workflow with names without the trailing "_2020020800" and simply changing DATE_FIRST_CYCL and DATE_LAST_CYCL from 20200801 to 20200208 and LBC_SPEC_INTVL_HRS from 3 to 1.  The change in date was to try out a winter case, which was the situation under which failures were common (the setting of kice to 9 for suites that use the RUC LSM seems to have fixed this winter case problem).

### DIFF
--- a/scripts/exregional_make_ics.sh
+++ b/scripts/exregional_make_ics.sh
@@ -302,22 +302,73 @@ lai_from_climo=""
 tg3_from_soil=""
 convert_nst=""
 #
-# If the physics suite uses Thompson microphysics and if the external
-# model for ICs is one that does not provide the aerosol fields needed
-# by Thompson microphysics (currently only the HRRR and RAP provide
-# aerosol data), set the variable thomp_mp_climo_file in the chgres_cube
-# namelist to the full path of the file containing aerosol climatology
-# data.  In this case, this file will be used to generate approximate
+#-----------------------------------------------------------------------
+#
+# If the external model is not one that uses the RUC land surface model
+# (LSM) -- which currently includes all valid external models except the
+# HRRR and the RAP -- then we set the number of soil levels to include
+# in the output NetCDF file that chgres_cube generates (nsoill_out; this
+# is a variable in the namelist that chgres_cube reads in) to 4.  This 
+# is because FV3 can handle this regardless of the LSM that it is using
+# (which is specified in the suite definition file, or SDF), as follows.  
+# If the SDF does not use the RUC LSM (i.e. it uses the Noah or Noah MP 
+# LSM), then it will expect to see 4 soil layers; and if the SDF uses 
+# the RUC LSM, then the RUC LSM itself has the capability to regrid from 
+# 4 soil layers to the 9 layers that it uses.
+#
+# On the other hand, if the external model is one that uses the RUC LSM
+# (currently meaning that it is either the HRRR or the RAP), then what
+# we set nsoill_out to depends on whether the RUC or the Noah/Noah MP
+# LSM is used in the SDF.  If the SDF uses RUC, then both the external
+# model and FV3 use RUC (which expects 9 soil levels), so we simply set
+# nsoill_out to 9.  In this case, chgres_cube does not need to do any
+# regridding of soil levels (because the number of levels in is the same
+# as the number out).  If the SDF uses the Noah or Noah MP LSM, then the
+# output from chgres_cube must contain 4 soil levels because that is what
+# these LSMs expect, and the code in FV3 does not have the capability to
+# regrid from the 9 levels in the external model to the 4 levels expected
+# by Noah/Noah MP.  In this case, chgres_cube does the regridding from 
+# 9 to 4 levels.
+#
+# In summary, we can set nsoill_out to 4 unless the external model is
+# the HRRR or RAP AND the forecast model is using the RUC LSM.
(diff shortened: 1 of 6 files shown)


scripts/exregional_make_ics.sh
```diff
@@ -302,22 +302,73 @@ lai_from_climo=""
...
+   [ "${SDF_USES_THOMPSON_MP}" = "TRUE" ]; then
   thomp_mp_climo_file="${THOMPSON_MP_CLIMO_FP}"
 fi
-
+#
+#-----------------------------------------------------------------------
+#
+# Set other chgres_cube namelist variables depending on the external
+# model used.
+#
```
JeffBeck-NOAA, line 369: At some point we can add specifics about the climo data being used (or not) below.

ush/generate_FV3LAM_wflow.sh
```diff
@@ -508,12 +508,21 @@ Setting parameters in FV3 namelist file (FV3_NML_FP):
...
+# generating the ICs.
 #
-if [ "${CCPP_PHYS_SUITE}" = "FV3_GSD_v0" ] || \
-   [ "${CCPP_PHYS_SUITE}" = "FV3_GSD_SAR" ]; then
+kice=""
+if [ "${SDF_USES_RUC_LSM}" = "TRUE" ]; then
 
+  kice="9"
+
+# Is lsoil needed at all?  It may be a variable that is not used in FV3
```
JeffBeck-NOAA: lsoil is still required.  It is identical to nsoill_out, so it's the number of input soil levels going into FV3 from the chgres_cube output, while lsoil_lsm is the number that the LSM scheme will run with.

JeffBeck-NOAA (approved):

ush/generate_FV3LAM_wflow.sh
gsketefian: Oh ok, good to know.  I'll change this comment accordingly (put in your explanation).

scripts/exregional_make_ics.sh
gsketefian: Which climo data are you referring to?

JeffBeck-NOAA: The *_from_climo flags.

JeffBeck-NOAA:
@gsketefian, thoroughly reviewed this PR again.  Looks good to me.

PR #435, opened 2021-02-12
Fixed an issue that colorlist is one element less than clevs in ush/P…

ywangwof:
…ython/plot_allvars.py

## DESCRIPTION OF CHANGES: 
Add one color "black" to "colorlist" in file _ush/Python/plot_allvars.py_ about line 606.

## TESTS CONDUCTED: 

It is tested on Odin (NSSL Cray computer) using Python version 3.8.3, matplotlib-base 3.3.2  and cartopy  0.18.0.

## ISSUE (optional): 

-  “colorlist” is 2 elements less than “clevs”. It is not allowed. 
-  “colorlist” can only be 1 element less than “clevs”. 
- The error message from Python is
`ValueError: There are 11 color bins including extensions, but ncolors = 10; ncolors must equal or exceed the number of bins`

### DIFF
--- a/ush/Python/plot_allvars.py
+++ b/ush/Python/plot_allvars.py
@@ -9,7 +9,7 @@
 #           David Wright 	Org: University of Michigan
 #
 # Instructions:		Make sure all the necessary modules can be imported.
-#                       Five command line arguments are needed:
+#                       Six command line arguments are needed:
 #                       1. Cycle date/time in YYYYMMDDHH format
 #                       2. Starting forecast hour
 #                       3. Ending forecast hour
@@ -603,7 +603,7 @@ for fhr in fhours:
     units = 'J/kg'
     clevs = [100,250,500,1000,1500,2000,2500,3000,3500,4000,4500,5000]
     clevs2 = [-2000,-500,-250,-100,-25]
-    colorlist = ['blue','dodgerblue','cyan','mediumspringgreen','#FAFAD2','#EEEE00','#EEC900','darkorange','crimson','darkred']
+    colorlist = ['blue','dodgerblue','cyan','mediumspringgreen','#FAFAD2','#EEEE00','#EEC900','darkorange','crimson','darkred','black']
     cm = matplotlib.colors.ListedColormap(colorlist)
     norm = matplotlib.colors.BoundaryNorm(clevs, cm.N)
 


BenjaminBlake-NOAA:
@jwolff-ncar It looks like the changes that were committed to the release/public-v1 branch to fix the CAPE/CIN colorbar were not added to the develop branch.  @ywangwof I actually have a slight modification to the change you are proposing, and this change is also needed for the plot_allvars_diff.py script.

jwolff-ncar:
@BenjaminBlake-NOAA I can issue a PR to develop once things settle down with the release unless someone beats me to it. Thanks for the reminder, though!

BenjaminBlake-NOAA:
@jwolff-ncar  Ok that sounds good!  Or I can if I get a chance to do it this week.  @ywangwof I'll add you as a reviewer so that you can test the changes on Odin.  They are very similar to the changes you made here, I just use 'darkviolet' instead of black and then set the "over" color to 'black'.

ywangwof:
Ben, 

it is good that you have solved the problem. I would be glad to test your change. Unfortunately, Odin is down currently due to unstable power supply caused by the high power demands (historical low temperature). 

Yunheng.

jwolff-ncar:
Any update on your testing @ywangwof? I know it may still be an issue to get onto Odin. We are hoping to get all PRs approved and merged for the release today so let us know your status.
Thanks!

BenjaminBlake-NOAA:
@jwolff-ncar Since this PR is to the develop branch and not the release branch, does it still need to be approved and merged today?  Thanks!

ywangwof:
@JeffBeck-NOAA Sorry. Odin is not powered on yet. Since it is just to align the color list with the contour level list, I do not think it will make much difference on Odin and on other machines. @BenjaminBlake-NOAA I did not receive the review request for your change? Which hash # should I test when Odin is back? Thanks.

jwolff-ncar:
Thanks for noting this is to develop. We need to get this into the release branch to support Odin for the release (it is a Tier 2 platform)

jwolff-ncar:
Let me apologize as there is a lot of going on to get the release out. There is a fix for this in the release branch. It needs to be pushed to the develop branch and that will solve this problem. Let's go with that approach rather than this slightly different approach. In the mean time, @ywangwof if you want to get the scripts from the release branch to test that would be great.

JeffBeck-NOAA:
Closing this PR based on @jwolff-ncar's comment.

Issue #434, opened 2021-02-11
New feature: sub-hourly output from FV3LAM workflow


Jeff-Duda:
Update to code that generates workflow and scripts to produce sub-hourly output in UPP.

Jeff-Duda:
PR implemented

PR #439, opened 2021-02-17
Sub-hourly post-processing enabled by some new config.sh entries.

Jeff-Duda:
## DESCRIPTION OF CHANGES: 
Modified the Jinja-formatted FV3LAM_wflow.xml template workflow to accommodate sub-hourly post-processing tasks that rely on sub-hourly FV3 output as a dependency. All changes are _additions_ to existing code and include the addition of a few keyword variables in the config.sh script. These new flags include...

- SUB_HOURLY_POST: a logical flag indicating whether nor not sub-hourly post-processing is to be used
- DT_SUBHOURLY_POST_MNTS: the increment in minutes to sub-divide the hour

Additional post-processing tasks were added to FV3LAM_wflow.xml to account for the different FV3 output file names depending on whether sub-hourly FV3 output is used (the first FV3 output file has a different naming structure than the remaining output files).

setup.sh was updated to check whether valid entries were used for these two variables and also check that DT_ATMOS divides evenly into DT_SUBHOURLY_POST_MNTS so that the FV3 output is consistent with the requested frequency of UPP output. config_defaults.sh and valid_param_vals.sh were also updated accordingly.

## TESTS CONDUCTED: 
Have run generate_FV3LAM_wflow.sh on a large variety of settings of SUB_HOURLY_POST and DT_SUBHOURLY_POST_MNTS. Note that setup.sh is configured such that DT_SUBHOURLY_POST_MNTS = 0 will cause SUB_HOURLY_POST to be ignored. I have successfully tested cases in which DT_ATMOS _does not_ divide evenly into DT_SUBHOURLY_POST_MNTS and when DT_SUBHOURLY_POST_MNTS is specified as anything other than a two-digit value (strings vs. open integers both work).

The resulting workflows run successfully with rocotorun and output no error messages.

## ISSUE: 
Resolves issue #434

## CONTRIBUTORS:
@gsketefian.  Contributions:

1) Fixed bug in setup.sh in the test that checks whether DT_SUBHOURLY_POST_MNTS is set to 0:  should use the -eq operator instead of ==.
2) Fixed bug in the jinja XML template for rocoto (FV3LAM_wflow.xml) as follows:  rearranged the post-processing tasks so that the post task is run for only the first minute of the last hour (e.g. if the forecast is 3 hours long, post is run for 3:00 but not for 3:15, 3:30, etc).
3) Ran the following 3 WE2E tests [note that tests (b) and (c) are not yet in the regional_workflow repo and will be included in a future PR]:
  a) **grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta**.  This is without subhourly post-processing, i.e. SUB_HOURLY_POST is set to "FALSE".
  b) **subhourly_post**.  This is with subhourly post-processing, i.e. SUB_HOURLY_POST set is set to "TRUE" (with DT_SUBHOURLY_POST_MNTS set to "12" minutes).
  c) **subhourly_post_ensemble_2mems**.  This is with subhourly post-processing and with ensemble forecasts enabled, i.e. SUB_HOURLY_POST and DO_ENSEMBLES are both set to "TRUE" (with DT_SUBHOURLY_POST_MNTS is set to "12" minutes and NUM_ENS_MEMBERS set to "2" members).  This test is run because the changes in the jinja XML template FV3LAM_wflow.xml needed to add subhourly post invovle code that executes ensemble forecasts.
**All three tests were successful.**

Note that this PR changes the names of the grib2 files that UPP generates such that they now always include the minutes -- regardless of whether SUB_HOURLY_POST is set to "TRUE" or "FALSE" (if set to "FALSE", the minutes are always "00").  For example, previously, the grib2 file for forecast hour 1 was named `rrfs.t00z.bgdawpf001.tm00.grib2`; henceforth, it will be named `rrfs.t00z.bgdawpf00100.tm00.grib2`.


scripts/exregional_run_post.sh
```diff
@@ -268,8 +282,29 @@ The \${fhr} variable contains too few or too many characters:
...
+  fi
+else
+  print_err_msg_exit "\
+The \${fmin} variable contains too few or too many characters:
+  fmin = \"$fmin\""
+fi
+
+ls -l BGDAWP.GrbF${post_fhr}*
+ls -l BGRD3D.GrbF${post_fhr}*
+
```
JeffBeck-NOAA: I'm guessing lines 298 and 299 were for testing purposes?  If so, can you remove those to limit logging output?  Thanks.

jobs/JREGIONAL_RUN_POST
```diff
@@ -108,6 +108,15 @@ digits:
...
+#-------------- check if all necessary environmental variables are included -----
+if [[ "${fmin}" ]] && [[ ! "${dt_atmos}" ]]; then
+  echo "ERROR: You set environmental variable \"fmin\" but did not supply \"dt_atmos\""
+  echo "  as well (same as in FV3 namelist). Both of these need to be set so that "
+  echo "  UPP will process the correct file for the first output timestep model fields."
+  exit 2
+fi
+#-------------------------------------------------------------------------
+#
 #-----------------------------------------------------------------------
```
JeffBeck-NOAA, line 124: Could you rewrite the error statement to follow the "print_err_msg_exit" paradigm (see line 135)?  That's how we handle error-then-exit in the workflow.  I see you used it in later commits, so I'm guessing this may have been one of your first ones!

Also, could you rewrite the comment as:

\#
\# Check if all necessary environmental variables are included.
\# 

Thanks!

ush/config.community.sh
```diff
@@ -17,6 +17,8 @@ LBC_SPEC_INTVL_HRS="6"
 DATE_FIRST_CYCL="20190615"
 DATE_LAST_CYCL="20190615"
 CYCL_HRS=( "00" )
+SUB_HOURLY_POST="TRUE"
+DELTA_MIN="15"
```
JeffBeck-NOAA: Can you set SUB_HOURLY_POST to "FALSE" as the default?  Right now, the hourly output is the standard way we run the SRW App, so users who wish to run sub-hourly can just flip this logical when they copy this file to config.sh.  Thanks.

ush/config.nco.sh
```diff
@@ -18,6 +18,8 @@ LBC_SPEC_INTVL_HRS="6"
 DATE_FIRST_CYCL="20190901"
 DATE_LAST_CYCL="20190901"
 CYCL_HRS=( "18" )
+SUB_HOURLY_POST="TRUE"
+DELTA_MIN="15"
```
JeffBeck-NOAA: Same comment as above.

ush/templates/FV3LAM_wflow.xml
```diff
@@ -405,37 +405,50 @@ MODULES_RUN_TASK_FP script.
...
+              <datadep age="05:00"><cyclestr>&CYCLE_BASEDIR;/@Y@m@d@H{{ slash_ensmem_subdir }}/dynf#fhr#:#fmin#:00.nc</cyclestr></datadep>
+              <datadep age="05:00"><cyclestr>&CYCLE_BASEDIR;/@Y@m@d@H{{ slash_ensmem_subdir }}/phyf#fhr#:#fmin#:00.nc</cyclestr></datadep>
+            </and>
+            <and>
+              <datadep age="05:00"><cyclestr>&CYCLE_BASEDIR;/@Y@m@d@H{{ slash_ensmem_subdir }}/dynf#fhr#.nc</cyclestr></datadep>
+              <datadep age="05:00"><cyclestr>&CYCLE_BASEDIR;/@Y@m@d@H{{ slash_ensmem_subdir }}/phyf#fhr#.nc</cyclestr></datadep>
+            </and>
+          </or>
+        </dependency>
+
```
JeffBeck-NOAA, line 572: Should lines 435-447 be wrapped in an {% if sub_hourly_post %}, {% else %}, {% endif %}, as in lines 411-415?  Otherwise you'd get both dependencies regardless of what SUB_HOURLY_POST is set to?

ush/generate_FV3LAM_wflow.sh
```diff
@@ -282,6 +282,11 @@ settings="\
   'ensmem_indx_name': ${ensmem_indx_name}
   'uscore_ensmem_name': ${uscore_ensmem_name}
   'slash_ensmem_subdir': ${slash_ensmem_subdir}
+#
+# Parameters associated with subhourly post-processed output
+#
+  'sub_hourly_post': ${SUB_HOURLY_POST}
+  'del_min': ${DELTA_MIN}
 " # End of "settings" variable.
```
JeffBeck-NOAA, line 291: Any reason 'del_min' can't be 'delta_min'?

JeffBeck-NOAA (changes requested):
@Jeff-Duda, thanks for getting this PR opened!  I took a quick look through and found only minor things so far.   To update a PR, you only need to make changes, commit then, then issue a 'git push' and it will automatically update everything in GitHub.  I'll check out your feature branch next week to run a test without sub-hourly data, but to test with sub-hourly data, I'll need to manually modify the model_configure file.  Therefore, one thing we should add in the future is a user-defined option to automatically set those model_configure parameters for sub-hourly output (leveraging your DELTA_MIN entry should be nearly all that's required to do that, I would think).  If you wanted, you could include that feature in this PR, or we could create a second PR later.  Up to you!

scripts/exregional_run_post.sh
Jeff-Duda: Correct...I was testing. Removed them.

jobs/JREGIONAL_RUN_POST
Jeff-Duda: Made requested formatting changes

ush/config.community.sh
Jeff-Duda: Made suggested change

ush/config.nco.sh
Jeff-Duda: Made suggested change.

ush/generate_FV3LAM_wflow.sh
Jeff-Duda: Changed del_min to delta_min in this script as well as the FV3LAM_wflow.xml template in templates/

jobs/JREGIONAL_RUN_POST
```diff
@@ -88,7 +88,7 @@ else
 fi
 mkdir_vrfy -p "${postprd_dir}"
 
-fhr_dir="${postprd_dir}/$fhr"
+fhr_dir="${postprd_dir}/$fhr$fmin"
```
gsketefian: Can we change "fmin" to something like "fmn"?  The "min" makes me think of "minimum", and "mn" is used in several other places in the scripts for "minutes".  Thanks.

jobs/JREGIONAL_RUN_POST
```diff
@@ -108,6 +108,15 @@ digits:
   fhr = \"${fhr}\""
 fi
 #
+#-------------- check if all necessary environmental variables are included -----
+if [[ "${fmin}" ]] && [[ ! "${dt_atmos}" ]]; then
```
gsketefian: For consistency with the rest of the scripts, modify this to:
```
if [ -n "$fmin" ] && [ -z "${dt_atmos}" ]; then
  print_err_msg_exit "\
You set environmental variable \"fmin\" but did not supply \"dt_atmos\"
as well (same as in FV3 namelist). Both of these need to be set so that
UPP will process the correct file for the first output timestep model fields."
fi
```
Note that the "print_err_msg_exit" function will exit the script with a nonzero error code, so there's no need to include a call to "exit 2".

ush/templates/FV3LAM_wflow.xml
Jeff-Duda: There is more than one way this can be done.

I went with a suggestion by Christina Holt to engineer this task such that there will _always_ be a set of nested metatasks regardless of whether sub-hourly output was requested. Christina thought that would be more helpful to users in understanding what is going on at the task level in the workflow. The idea is that, when sub-hourly processing is not being used, the workflow <var=fmin> array will just be the single entry "00". Leaving the dependency structure as is would mean the dependencies would always be appropriate no matter whether sub-hourly processing was used or not. But yes, it does result in a <dependency> structure that seems a bit kooky.

If other reviewers feel strongly about adopting your suggestion, that can certainly be done. Let's get the opinions of others, though.

jobs/JREGIONAL_RUN_POST
Jeff-Duda: Changed $fmin to $fmn throughout jobs/JREGIONAL_RUN_POST and scripts/exregional_run_post.sh, but not in the template FV3LAM_wflow file. I feel that in the workflow, the meaning is pretty straightforward and isolated from other usages within the system.

jobs/JREGIONAL_RUN_POST
Jeff-Duda: Already modified as per Jeff Beck's comments.

ush/config.community.sh
```diff
@@ -17,6 +17,8 @@ LBC_SPEC_INTVL_HRS="6"
 DATE_FIRST_CYCL="20190615"
 DATE_LAST_CYCL="20190615"
 CYCL_HRS=( "00" )
+SUB_HOURLY_POST="TRUE"
```
gsketefian: Probably don't want to include these in the sample config files (config.community.sh and config.nco.sh) since most people won't be using this capability (not yet at least).  But we should definitely add a workflow end-to-end (WE2E) test that exercises this capability so that we can be sure future PRs don't break it.  That's what I usually do (i.e. whenever I add a new capability to the workflow, I add a corresponding WE2E test).  We can do that once the rest of the issues in this PR are resolved, or as part of a separate PR.

ush/config_defaults.sh
```diff
@@ -381,6 +381,8 @@ DATE_FIRST_CYCL="YYYYMMDD"
 DATE_LAST_CYCL="YYYYMMDD"
 CYCL_HRS=( "HH1" "HH2" )
 FCST_LEN_HRS="24"
+SUB_HOURLY_POST="FALSE"
```
gsketefian, line 393: Please include in the comments above descriptions of what these variables are.

ush/config.nco.sh
```diff
@@ -18,6 +18,8 @@ LBC_SPEC_INTVL_HRS="6"
 DATE_FIRST_CYCL="20190901"
 DATE_LAST_CYCL="20190901"
 CYCL_HRS=( "18" )
+SUB_HOURLY_POST="TRUE"
```
gsketefian: Same as above, i.e. remove from this file.

ush/config.community.sh
Jeff-Duda: Removed DELTA_MIN and SUB_HOURLY_POST from the sample config files.

ush/config_defaults.sh
Jeff-Duda: Same as above

ush/config.nco.sh
Jeff-Duda: Same as above

ush/setup.sh
```diff
@@ -581,6 +604,20 @@ fi
 #
 #-----------------------------------------------------------------------
 #
+# Check that DELTA_MIN is a string consisting of exactly 2 digits between "00" and "59"
+#
+#-----------------------------------------------------------------------
+#
+MIN_OR_NULL=$( printf "%s" "${DELTA_MIN}" | \
```
gsketefian: To follow the convention usually used in the rest of the script, make MIN_OR_NULL lowercase (min_or_null) since it is a local variable.  Usually, global workflow/experiment variables (those that get recorded in var_defns.sh) are uppercase and local variables (those that are used only within a script) are lowercase.

ush/config_defaults.sh
```diff
@@ -381,6 +381,8 @@ DATE_FIRST_CYCL="YYYYMMDD"
...
+DELTA_MIN="15"
```
gsketefian: The "MIN" again threw me off at first because I thought this was some sort of minimum value for a variable named DELTA.  It becomes clear what it is after one looks at the code, but we'd like it to be clearer than that so we pretty much know what it is right when we read it.  So maybe rename to something like DT_SUBHOURLY_POST_MNTS to indicate that it is a time interval (DT), used only when doing subhourly post (SUBHOURLY_POST), and is in units of minutes (MNTS; the workflow uses MNTS elsewhere to indicate units are in minutes).

ush/generate_FV3LAM_wflow.sh
```diff
@@ -282,6 +282,11 @@ settings="\
   'ensmem_indx_name': ${ensmem_indx_name}
   'uscore_ensmem_name': ${uscore_ensmem_name}
   'slash_ensmem_subdir': ${slash_ensmem_subdir}
+#
+# Parameters associated with subhourly post-processed output
+#
+  'sub_hourly_post': ${SUB_HOURLY_POST}
+  'del_min': ${DELTA_MIN}
```
gsketefian: For clarity, use same variable name (except lowercase) in jinja template as in workflow, i.e. use delta_min instead of del_min.

ush/setup.sh
```diff
@@ -975,6 +1012,18 @@ The forecast model main time step (DT_ATMOS) is set to a null string:
 Please set this to a valid numerical value in the user-specified experiment
 configuration file (EXPT_CONFIG_FP) and rerun:
   EXPT_CONFIG_FP = \"${EXPT_CONFIG_FP}\""
+else
```
gsketefian, line 1019: Three comments here:
1) This check can be a separate if-statement later below (e.g. after the check of BLOCKSIZE) instead of being here as an else-clause.
2) Probably should do this check only if SUB_HOURLY_POST is set to TRUE.  
3) The error message can be clarified to specify that DELTA_MIN (after converting to seconds) must be evenly divisible by DT_ATMOS.

Overall, maybe something like this (note: I haven't tested this):
```
if [ "${SUB_HOURLY_POST}" = "TRUE" ]; then
  rem=$(( (DELTA_MIN*60) % DT_ATMOS ))
  if [ "$rem" -ne "0" ]; then
    print_err_msg_exit "\
When performing sub-hourly post (i.e. SUB_HOURLY_POST set to \"TRUE\"), the time
interval specified by DELTA_MIN (after converting to seconds) must be evenly divisible 
by the time step DT_ATMOS used in the forecast model, i.e. the remainder (rem) must 
be zero.  In this case, it is not:
  SUB_HOURLY_POST = \"${SUB_HOURY_POST}\"
  DELTA_MIN = \"${DELTA_MIN}\"
  DT_ATMOS = \"${DT_ATMOS}\"
  rem = \$(( (DELTA_MIN*60) % DT_ATMOS )) = $rem
Please reset DELTA_MIN and/or DT_ATMOS so that the remainder is zero."
  fi
fi
```

ush/setup.sh
Jeff-Duda: I made the requested change, but I do not understand what makes MIN_OR_NULL different from other nearby similar variables such as DATE_OR_NULL or CYCL_OR_NULL that are also in setup.sh and seem to have the same purpose but are in upper-case.

ush/config_defaults.sh
Jeff-Duda: Changed global variable DELTA_MIN to DT_SUBHOURLY_POST_MNTS throughout

scripts/exregional_run_post.sh
```diff
@@ -204,20 +206,32 @@ tmmark="tm00"
 #
 #-----------------------------------------------------------------------
 #
-dyn_file="${run_dir}/dynf${fhr}.nc"
-phy_file="${run_dir}/phyf${fhr}.nc"
+if [ ! "${fmin}" ]; then
```
gsketefian: Can reduce repetition and hard-coding by doing something like this:

```
mnts_secs_str=""
if [ "${SUB_HOURLY_POST}" = "TRUE" ]; then
  if [ ${fhr} = "000" ]; then
    mnts_secs_str=":${fmin}:45"
  else
    mnts_secs_str=":${fmin}:00"
  fi
else
  fmin="00"
fi

dyn_file="${run_dir}/dynf${fhr}${mnts_secs_str}.nc"
phy_file="${run_dir}/phyf${fhr}${mnts_secs_str}.nc"
post_time=$( date --utc --date "${yyyymmdd} ${hh} UTC + ${fhr} hours + ${fmin} minutes" "+%Y%m%d%H%M" )
```

ush/generate_FV3LAM_wflow.sh
Jeff-Duda: Done

scripts/exregional_run_post.sh
```diff
@@ -268,8 +282,29 @@ The \${fhr} variable contains too few or too many characters:
   fhr = \"$fhr\""
 fi
 
-mv_vrfy BGDAWP.GrbF${post_fhr} ${postprd_dir}/${NET}.t${cyc}z.bgdawpf${fhr}.${tmmark}.grib2
-mv_vrfy BGRD3D.GrbF${post_fhr} ${postprd_dir}/${NET}.t${cyc}z.bgrd3df${fhr}.${tmmark}.grib2
+len_fmin=${#fmin}
+if [ ${len_fmin} -eq 2 ]; then
+  post_fmin=${fmin}
+elif [ ${len_fmin} -eq 3 ]; then
```
gsketefian: Under what circumstances is fmin set to something with more than 2 digits?

ush/setup.sh
Jeff-Duda: Made exact suggested changes

scripts/exregional_run_post.sh
```diff
@@ -268,8 +282,29 @@ The \${fhr} variable contains too few or too many characters:
...
+else
+  print_err_msg_exit "\
+The \${fmin} variable contains too few or too many characters:
+  fmin = \"$fmin\""
+fi
+
+ls -l BGDAWP.GrbF${post_fhr}*
+ls -l BGRD3D.GrbF${post_fhr}*
+
+if [[ ! "${fmin}" ]] || [[ ${fmin} -eq 00 ]]; then
```
gsketefian: With the suggested code above where fmin gets set to "00" if SUB_HOURLY_POST is not set to "TRUE", you don't need this if-statement, just the else part of it, i.e. just use
```
mv_vrfy BGDAWP.GrbF${post_fhr}.${fmin} ${postprd_dir}/${NET}.t${cyc}z.bgdawpf${fhr}${fmin}.${tmmark}.grib2
mv_vrfy BGRD3D.GrbF${post_fhr}.${fmin} ${postprd_dir}/${NET}.t${cyc}z.bgrd3df${fhr}${fmin}.${tmmark}.grib2
```

scripts/exregional_run_post.sh
Jeff-Duda: Made exact suggested change and also changed "45" to ${dt_atmos} (passed in from JREGIONAL_RUN_POST) to reflect actual behavior of FV3 output at forecast hour 0.

scripts/exregional_run_post.sh
```diff
@@ -280,9 +315,9 @@ mv_vrfy BGRD3D.GrbF${post_fhr} ${postprd_dir}/${NET}.t${cyc}z.bgrd3df${fhr}.${tm
 # instead of calling sed.
 start_date=$( echo "${cdate}" | sed 's/\([[:digit:]]\{2\}\)$/ \1/' )
 basetime=$( date +%y%j%H%M -d "${start_date}" )
-ln_vrfy -fs ${postprd_dir}/${NET}.t${cyc}z.bgdawpf${fhr}.${tmmark}.grib2 \
+ln_vrfy -fs ${postprd_dir}/${NET}.t${cyc}z.bgdawpf${fhr}00.${tmmark}.grib2 \
```
gsketefian: Why is the change in the file name here always "00"?  Shouldn't it be "$fmin" in general?

scripts/exregional_run_post.sh
Jeff-Duda: Probably none. I was just absent-mindedly modeling this section off of the one for forecast hour above. I removed that elif block.

scripts/exregional_run_post.sh
Jeff-Duda: Just note that there are instances during which UPP outputs a different file naming structure depending on whether sub-hourly output is requested or not. When no sub-hourly output is requested, the output file name from UPP does not contain the ".${fmn}" syntax. I am not 100% sure how this is yet determined, though. I think it is through the time entry in the itag file, but it might be something different.

As of now, I have removed the if-then statement.

scripts/exregional_run_post.sh
Jeff-Duda: Made suggested change.

gsketefian:
@Jeff-Duda Can you push your latest changes to your fork so we can see them under the "Files Changed" tab?  Thanks.

Jeff-Duda:
> 
> 
> @Jeff-Duda Can you push your latest changes to your fork so we can see them under the "Files Changed" tab? Thanks.

@gsketefian I am going to re-run some tests to confirm the changes still work before pushing changes.

scripts/exregional_run_post.sh
Jeff-Duda: Okay, so I can confirm that when the minutes-past-the-hour in the time is "00", UPP outputs a file with a different naming format, so this if statement needs to be there for now.

Jeff-Duda:
Completed bug fixes as suggested by @JeffBeck-NOAA and @gsketefian. Notes on those changes are addressed for each comment.

Referring back to the original PR request message, the variable ${DT_ATMOS} in exregional_run_post.sh is no longer unused (it is now used to specify the very first FV3 output file). Moreover, ${DT_ATMOS} and ${DT_SUBHOURLY_POST_MNTS} have been added to exregional_run_fcst.sh so that they get into create_model_configure_file.sh to write the model_configure file to produce sub-hourly output from FV3. This set of changes also means that sub-hourly FV3 output can now be enabled by using the same settings as for sub-hourly UPP output, although we may need to re-visit separating those two (note: getting sub-hourly FV3 output is not dependent on the SUB_HOURLY_POST logical flag in the config file, so a user should be able to setup a workflow to enable sub-hourly FV3 output without also performing sub-hourly post-processing.

ush/config_defaults.sh
Jeff-Duda: I mis-read this comment the first time around, and thus the first of the new commits (hash f4f7fc568eb57cfce03aa90c89fc19062e5401bf) doesn't actually reflect this comment. But I made a second commit that does include descriptions for these new variables in config_defaults.sh.

ush/templates/FV3LAM_wflow.xml
JeffBeck-NOAA: OK, that makes sense.  So the way this is written now, if you run with sub-hourly output, you need all four "phyf#fhr#:#fmin#:{{ dt_atmos }}", "phyf#fhr#:#fmin#:00.nc", "dynf#fhr#:#fmin#:{{ dt_atmos }}", and "dynf#fhr#:#fmin#:00.nc" files before the post task will kick off for a combined set of fhr and fmn values?  Otherwise, if it's hourly output, then it operates as before (because you have an \<or\> in the dependency)?

scripts/exregional_run_post.sh
JeffBeck-NOAA: Do the "00"s in lines 311 and 313 still need to be replaced by ${fmn}?

ush/setup.sh
JeffBeck-NOAA: Have you tested this code block with a value of DT_ATMOS that isn't evenly divisible by DELTA_MIN to ensure you get the new error message?

ush/templates/FV3LAM_wflow.xml
Jeff-Duda: Yes, that is correct.

ush/setup.sh
Jeff-Duda: Good call. Yes, the check worked as far as quitting the generate script if an inconsistent set of values was used, but the line 

`rem = \$(( (DELTA_MIN*60) % DT_ATMOS )) = $rem`

was causing a secondary error, so I replaced that with

`rem = ${val}`

and commit-pushed it (commit hash 3621f16c).

scripts/exregional_run_post.sh
Jeff-Duda: Yes. They now have been.

scripts/exregional_run_fcst.sh
```diff
@@ -451,7 +451,10 @@ fi
 create_model_configure_file \
   cdate="$cdate" \
   nthreads=${OMP_NUM_THREADS:-1} \
-  run_dir="${run_dir}" || print_err_msg_exit "\
+  run_dir="${run_dir}" \
+  sub_hourly_post="${SUB_HOURLY_POST}" \
+  delta_fmin="${DT_SUBHOURLY_POST_MNTS}" \
```
gsketefian: To keep consistent with the rest of the variables, change delta_fmin to dt_subhourly_post_mnts.  Thx.

scripts/exregional_run_post.sh
```diff
@@ -268,8 +282,22 @@ The \${fhr} variable contains too few or too many characters:
   fhr = \"$fhr\""
 fi
 
-mv_vrfy BGDAWP.GrbF${post_fhr} ${postprd_dir}/${NET}.t${cyc}z.bgdawpf${fhr}.${tmmark}.grib2
-mv_vrfy BGRD3D.GrbF${post_fhr} ${postprd_dir}/${NET}.t${cyc}z.bgrd3df${fhr}.${tmmark}.grib2
+len_fmin=${#fmn}
```
gsketefian, line 285: Don't need this if-statement because fmn is hard-coded in the template xml to be a 2-digit string.

gsketefian: Don't need this if-statement at all since post_fmn is set but not used.

scripts/exregional_run_post.sh
```diff
@@ -204,20 +206,32 @@ tmmark="tm00"
...
+dyn_file="${run_dir}/dynf${fhr}${mnts_secs_str}.nc"
+phy_file="${run_dir}/phyf${fhr}${mnts_secs_str}.nc"
+post_time=$( date --utc --date "${yyyymmdd} ${hh} UTC + ${fhr} hours +${fmn} minutes" "+%Y%m%d%H%M" )
 
-post_time=$( date --utc --date "${yyyymmdd} ${hh} UTC + ${fhr} hours" "+%Y%m%d%H" )
 post_yyyy=${post_time:0:4}
 post_mm=${post_time:4:2}
 post_dd=${post_time:6:2}
 post_hh=${post_time:8:2}
+post_min=${post_time:10:2}
```
gsketefian: For consistency with the other variables, change post_min to post_mn.

scripts/exregional_run_post.sh
```diff
@@ -268,8 +282,22 @@ The \${fhr} variable contains too few or too many characters:
...
+  print_err_msg_exit "\
+The \${fmn} variable contains too few or too many characters:
+  fmn = \"$fmn\""
+fi
+
+if [ "${fmn}" = "00" ]; then
+  mv_vrfy BGDAWP.GrbF${post_fhr} ${postprd_dir}/${NET}.t${cyc}z.bgdawpf${fhr}00.${tmmark}.grib2
+  mv_vrfy BGRD3D.GrbF${post_fhr} ${postprd_dir}/${NET}.t${cyc}z.bgrd3df${fhr}00.${tmmark}.grib2
+else
+  mv_vrfy BGDAWP.GrbF${post_fhr}.${fmn} ${postprd_dir}/${NET}.t${cyc}z.bgdawpf${fhr}${fmn}.${tmmark}.grib2
```
gsketefian: For consistency with what is done with the hours, use post_mn instead of fmn in the BGDAWP and BGRD3D file names.

ush/setup.sh
```diff
@@ -365,6 +365,29 @@ fi
...
+SUB_HOURLY_POST=${SUB_HOURLY_POST^^}
+if [ "$SUB_HOURLY_POST" = "TRUE" ] || \
+   [ "$SUB_HOURLY_POST" = "YES" ]; then
+  SUB_HOURLY_POST="TRUE"
+elif [ "$SUB_HOURLY_POST" = "FALSE" ] || \
+     [ "$SUB_HOURLY_POST" = "NO" ]; then
+  SUB_HOURLY_POST="FALSE"
+fi
+if [ ${DT_SUBHOURLY_POST_MNTS} == 00 ]; then
+  SUB_HOURLY_POST="FALSE"
```
gsketefian, line 387: Can you include an informational message to the user that this variable is being changed from what it was in the user's configuration file (and why)?  Use the print_info_msg function to do that (lots of examples in this and other files of that).  Thanks!

ush/setup.sh
```diff
@@ -975,6 +1012,22 @@ The forecast model main time step (DT_ATMOS) is set to a null string:
...
+    if [ ${val} -ne 0 ]; then
+      print_err_msg_exit "\
+When performing sub-hourly post (i.e. SUB_HOURLY_POST set to \"TRUE\"), the time
+interval specified by DT_SUBHOURLY_POST_MNTS (after converting to seconds) must be evenly divisible 
+by the time step DT_ATMOS used in the forecast model, i.e. the remainder (rem) must 
+be zero.  In this case, it is not:
+  SUB_HOURLY_POST = \"${SUB_HOURLY_POST}\"
+  DT_SUBHOURLY_POST_MNTS = \"${DT_SUBHOURLY_POST_MNTS}\"
+  DT_ATMOS = \"${DT_ATMOS}\"
+  rem = ${val}
```
gsketefian: The problem with this line was probably the percent sign.  Try escaping it by using a double-percent, like this:
`  rem = \$(( (DELTA_MIN*60) %% DT_ATMOS )) = $rem
`
That worked for me.  It would be good to show the formula so the user knows exactly what the calculation being performed is.

ush/create_model_configure_file.sh
```diff
@@ -49,6 +49,9 @@ function create_model_configure_file() {
 cdate \
 run_dir \
 nthreads \
+sub_hourly_post \
+delta_fmin \
```
gsketefian: For consistency with the rest of the scripts, change delta_fmin to dt_subhourly_post_mnts.  Having the same name for the same variable everywhere in the code comes in handy when grepping all the scripts to find where it's used.

ush/create_model_configure_file.sh
```diff
@@ -170,6 +173,17 @@ run directory (run_dir):
 
   fi
 
+  if [ "${sub_hourly_post}" = "TRUE" ]; then
+   ((nsout = delta_fmin*60 / dt_atmos))
```
gsketefian: For consistency with other arithmetic lines, change to:
`nsout=$(( dt_subhourly_post_mnts*60 / dt_atmos ))`
Actually, this seems to be the same variable as the remainder calculated in setup.sh that has to be zero when doing subourly post; otherwise, setup.sh quits.  Doesn't that mean it will always be zero here?

ush/create_model_configure_file.sh
```diff
@@ -170,6 +173,17 @@ run directory (run_dir):
 
   fi
 
+  if [ "${sub_hourly_post}" = "TRUE" ]; then
```
gsketefian, line 180: Can you add a comment right above here explaining what nfhout and nsout are?

ush/templates/FV3LAM_wflow.xml
gsketefian: Is all this valid when dt_atmos is greater than 59sec?  In that case, do you have to convert some of dt_atmos to minutes?  Say dt_subhourly_post_mnts is 15 minutes (900 sec) and dt_atmos is 90 sec, then the remainder is 0 so the test in setup.sh will pass, but here the file name will be strange -- phyf....90.nc.

jobs/JREGIONAL_RUN_POST
```diff
@@ -110,6 +110,19 @@ fi
 #
 #-----------------------------------------------------------------------
 #
+# check if all necessary environmental variables are included
+#
+#-----------------------------------------------------------------------
+#
+if [[ "${fmn}" ]] && [[ ! "${DT_ATMOS}" ]]; then
```
gsketefian: For consistency with other if-statements, change to 
if [ -n "${fmn}" ] && [ -z "${DT_ATMOS}" ]; then

ush/config_defaults.sh
gsketefian: Ok, thanks for including those descriptions.

scripts/exregional_run_post.sh
gsketefian: Also, to avoid repeating the move commands (including the hard-coded file names) in two places, can rewrite this if-statement this way:
```
dot_post_mn_or_null=""
if [ "${post_mn}" = "00" ]; then
  dot_post_mn_or_null=".${post_mn}"
fi
post_fn_suffix=".GrbF${post_fhr}${dot_post_mn_or_null}"
mv_vrfy BGDAWP${post_fn_suffix} ${postprd_dir}/${NET}.t${cyc}z.bgdawpf${fhr}${fmn}.${tmmark}.grib2
mv_vrfy BGRD3D${post_fn_suffix} ${postprd_dir}/${NET}.t${cyc}z.bgrd3df${fhr}${fmn}.${tmmark}.grib2
```

ush/templates/FV3LAM_wflow.xml
Jeff-Duda: That's a good point, Gerard. I have never run FV3 with dt_atmos > 59, so I can't say I'll know how the output file name will look. I would imagine it would get converted into minutes, so perhaps that will have to be examined and tested first.

scripts/exregional_run_fcst.sh
Jeff-Duda: Changed here as well as in create_model_configure_file.sh

scripts/exregional_run_post.sh
Jeff-Duda: Removed if statement after seeing this duty already accomplished by this earlier command:

`post_mn=${post_time:10:2}`

scripts/exregional_run_post.sh
Jeff-Duda: Made suggested change.

scripts/exregional_run_post.sh
Jeff-Duda: The logical operator is inverted from what it should be, but otherwise I implemented this change as suggested.

ush/setup.sh
Jeff-Duda: Done as suggested.

ush/create_model_configure_file.sh
Jeff-Duda: Done as indicated from a previous comment.

ush/create_model_configure_file.sh
Jeff-Duda: Done.

jobs/JREGIONAL_RUN_POST
Jeff-Duda: Done as suggested.

ush/create_model_configure_file.sh
Jeff-Duda: This calculation is the integer value ('/'), not the remainder ('%'). So no, $nsout will not always be 0.

Code format was changed as suggested.

ush/setup.sh
Jeff-Duda: Replaced line as suggested, but the output is not accurate. For example, I ran the following test. The remainder value should print out as 38, but printed out as 0.

When performing sub-hourly post (i.e. SUB_HOURLY_POST set to "TRUE"), the time
interval specified by DT_SUBHOURLY_POST_MNTS (after converting to seconds) must be evenly divisible
by the time step DT_ATMOS used in the forecast model, i.e. the remainder (rem) must
be zero.  In this case, it is not:
  SUB_HOURLY_POST = "TRUE"
  DT_SUBHOURLY_POST_MNTS = "06"
  DT_ATMOS = "46"
  rem = $(( (DELTA_MIN*60) % DT_ATMOS )) = 0
Please reset DT_SUBHOURLY_POST_MNTS and/or DT_ATMOS so that the remainder is zero.

gsketefian:
@Jeff-Duda Thanks for addressing the comments.  Can you also push your latest to your fork so we can see the changes?  I see your comments stating the changes you've made, but I don't see the actual changes in the files.  I assume it's because you haven't yet pushed your latest.  Thx.

ush/templates/FV3LAM_wflow.xml
Jeff-Duda: I ran a test with dt_atmos > 59 and it turns out the FV3 output is properly formatted using fFFF:mm:SS, so it correctly converts seconds-past-59 to minutes & seconds. So i modified the template file and generation script accordingly.

During this investigation I discovered I had modified the workflow template incorrectly...all post tasks would have their dependency met as soon as the very first FV3 file was available, which is undesirable. Fixing this required me to separate the one metatask set into two separate tasks; the new one runs UPP only on that very first FV3 output file ("f00000"). However, it is essentially impossible to remove that same time from the metatask setup without further separating the tasks, which seems unnecessarily cumbersome. So the result is that there will be two f00000 post tasks, and one of them will never run (since the exact "dynf000:00:00.nc" file will never exist).

ush/setup.sh
```diff
@@ -975,6 +1015,22 @@ The forecast model main time step (DT_ATMOS) is set to a null string:
 Please set this to a valid numerical value in the user-specified experiment
 configuration file (EXPT_CONFIG_FP) and rerun:
   EXPT_CONFIG_FP = \"${EXPT_CONFIG_FP}\""
+else
+  if [ "${SUB_HOURLY_POST}" = "TRUE" ]; then
+    ((val = (DT_SUBHOURLY_POST_MNTS*60) % DT_ATMOS))
```
gsketefian: Change line 1020-1021 to the following, otherwise the message will fail (because rem further below would not be defined:
```
    rem=$(( (DELTA_MIN*60) % DT_ATMOS ))
    if [ "$rem" -ne "0" ]; then

```

ush/create_model_configure_file.sh
```diff
@@ -170,6 +173,20 @@ run directory (run_dir):
 
   fi
 
+  # nfhout is the forecast hour output interval (provided forecast length is at most 60 hours)
```
gsketefian: Can you work the following in your comments here?  Thanks.

If performing sub-hourly post-processing, specify the output frequency in units of number of forecast model time steps (i.e. specify nsout).  Note that nsout is guaranteed to be an integer because a check has been placed in the experiment generation scripts to ensure that in this case, the remainder of dt_subhourly_post_mnts*60 divided by dt_atmos is zero.  If not performing sub-hourly post-processing, specify the output frequency in units of hours (i.e. specify nfhout).

ush/setup.sh
Jeff-Duda: Done.

gsketefian: @Jeff-Duda Hey Jeff, it is easier for the reviewers if you push your changes and let the reviewers resolve the conversation once they can see that it's actually been pushed.  Often people make the change on their clone but forget to push.  For example, in this case, I don't yet see any changes.  Thx.

ush/setup.sh
```diff
@@ -975,6 +1015,22 @@ The forecast model main time step (DT_ATMOS) is set to a null string:
 Please set this to a valid numerical value in the user-specified experiment
 configuration file (EXPT_CONFIG_FP) and rerun:
   EXPT_CONFIG_FP = \"${EXPT_CONFIG_FP}\""
+else
+  if [ "${SUB_HOURLY_POST}" = "TRUE" ]; then
+    ((rem = (DT_SUBHOURLY_POST_MNTS*60) % DT_ATMOS))
```
gsketefian, line 1021: For consistency with the rest of the scripts, let's do math this way:
`rem=$(( (DT_SUBHOURLY_POST_MNTS*60) % DT_ATMOS ))
`
Thanks.

ush/setup.sh
Jeff-Duda: I don't know what to say then...that is exactly the change I made, committed, and pushed...

[Jeffrey.Duda@hfe09 ush]$ git diff b5d8c4701a5b3622f9b9a0577ad278536f1fbe6d f49aee6e66223df34453753fbedb070eb8ecc235
diff --git a/ush/create_model_configure_file.sh b/ush/create_model_configure_file.sh
index 463e48f4..11c3a2d3 100644
--- a/ush/create_model_configure_file.sh
+++ b/ush/create_model_configure_file.sh
@@ -173,10 +173,9 @@ run directory (run_dir):

   fi

-  # IF using sub-hourly FV3/UPP output, then the FV3 output interval needs to be specified in
-  # units of multiples of model time steps (nsout). nsout is guaranteed to be an integer because
-  # consistency between ${DT_SUBHOURLY_POST_MNTS} and ${DT_ATMOS} was enforced at the time of script generation.
-  # ELSE only nfhout is used and set to 1-hour output.
+  # nfhout is the forecast hour output interval (provided forecast length is at most 60 hours)
+  # nsout is the output interval in multiples of dt_atmos; it is useful for controlling sub-hourly FV3 output
+  # in the below formulation, it is explicitly designed for minute-specific (rather than second-specifc) output
   if [ "${sub_hourly_post}" = "TRUE" ]; then
    nsout=$(( dt_subhourly_post_mnts*60 / dt_atmos ))
    nfhout=0
diff --git a/ush/setup.sh b/ush/setup.sh
index 1858b184..245479b3 100755
--- a/ush/setup.sh
+++ b/ush/setup.sh
@@ -1017,8 +1017,8 @@ configuration file (EXPT_CONFIG_FP) and rerun:
   EXPT_CONFIG_FP = \"${EXPT_CONFIG_FP}\""
 else
   if [ "${SUB_HOURLY_POST}" = "TRUE" ]; then
-    ((rem = (DT_SUBHOURLY_POST_MNTS*60) % DT_ATMOS))
-    if [ ${rem} -ne 0 ]; then
+    ((val = (DT_SUBHOURLY_POST_MNTS*60) % DT_ATMOS))
+    if [ ${val} -ne 0 ]; then
       print_err_msg_exit "\
 When performing sub-hourly post (i.e. SUB_HOURLY_POST set to \"TRUE\"), the time
 interval specified by DT_SUBHOURLY_POST_MNTS (after converting to seconds) must be evenly divisible

ush/templates/FV3LAM_wflow.xml
gsketefian: Haven't yet gotten a chance to look at this issue in detail due to urgent work on the release of the SRW app and it being somewhat complicated.

gsketefian:
@Jeff-Duda Can you point me to experiment directories where you've run with the subhourly post (and preferably another one without) so I can better understand the output file names, etc in the two cases?  Having a task that doesn't ever run will throw off the WE2E test system (the part where it reports whether the workflow for each test completed successfully or not), so I'd like to see what it will take to fix that.  Thx.

Jeff-Duda:
> 
> 
> @Jeff-Duda Can you point me to experiment directories where you've run with the subhourly post (and preferably another one without) so I can better understand the output file names, etc in the two cases? Having a task that doesn't ever run will throw off the WE2E test system (the part where it reports whether the workflow for each test completed successfully or not), so I'd like to see what it will take to fix that. Thx.

You can check out this space on Hera: /scratch1/BMC/wrfruc/jdduda/FV3_workflows/derecho_2020

gsketefian:
@Jeff-Duda Thanks, that helps to see what's going on.  Do you have one when SUB_HOURLY_POST is set to FALSE?  I'd like to see how the file and task names are going to change from what they are now for that case.

Jeff-Duda:
> 
> 
> @Jeff-Duda Thanks, that helps to see what's going on. Do you have one when SUB_HOURLY_POST is set to FALSE? I'd like to see how the file and task names are going to change from what they are now for that case.

Not at this time, but I can create one in /scratch1/BMC/wrfruc/jdduda/FV3_workflows/test_hourly

gsketefian:
That would be great.  I would test your PR myself but am way over hours for this FV3LAM support project.

Jeff-Duda:
Okay...a workflow is running at that location. Check back periodically throughout the day to see the progress.

gsketefian:
If/when you try without sub-hourly post, just one cycle will be sufficient.

Btw, in the workflow's community mode (which is what you're using, I think), there is a way to skip the pre-processing tasks if you've already generated the grid, orography, and surface climatology files for the grid.  Below is from an email I sent to another user who asked about this:

```
Ok, to skip the make_grid, make_orog, and make_sfc_climo tasks, first create a directory in which to place pregenerated 
copies of these files.  For example, for the 25km CONUS grid, I've put the files in this directory (on Hera):

/scratch2/BMC/det/FV3LAM_pregen/RRFS_CONUS_25km

Then you'll have to add the following lines to your config.sh file (where I'm using the directory above for GRID_DIR, 
OROG_DIR, and SFC_CLIMO_DIR):

#
# Set RUN_TASK_MAKE_GRID to "FALSE" to skip the MAKE_GRID_TN task.  In 
# this case, the directory containing the pregenerated grid files (GRID_DIR)
# must be specified.
#
RUN_TASK_MAKE_GRID="FALSE"
GRID_DIR="/scratch2/BMC/det/FV3LAM_pregen/RRFS_CONUS_25km"
#
# Set RUN_TASK_MAKE_OROG to "FALSE" to skip the MAKE_OROG_TN task.  In 
# this case, the directory containing the pregenerated orography files 
# (OROG_DIR) must be specified.
#
RUN_TASK_MAKE_OROG="FALSE"
OROG_DIR="/scratch2/BMC/det/FV3LAM_pregen/RRFS_CONUS_25km"
#
# Set RUN_TASK_MAKE_SFC_CLIMO to "FALSE" to skip the MAKE_SFC_CLIMO_TN 
# task.  In this case, the directory containing the pregenerated surface
# climatology files (SFC_CLIMO_DIR) must be specified.
#
RUN_TASK_MAKE_SFC_CLIMO="FALSE"
SFC_CLIMO_DIR="/scratch2/BMC/det/FV3LAM_pregen/RRFS_CONUS_25km"

An example of an experiment that does this can be found here on Hera:

/scratch2/BMC/det/Gerard.Ketefian/UFS_CAM/PR_feature_nco_fix_links/expt_dirs/links02/pregen_grid_orog_sfc_climo

Please let me know if you run into any difficulties.  Thanks,
Gerard
```

ush/templates/FV3LAM_wflow.xml
```diff
@@ -405,39 +405,87 @@ MODULES_RUN_TASK_FP script.
...
+        <dependency>
+          <or>
+            <taskdep task="&RUN_FCST_TN;{{ uscore_ensmem_name }}"/>
+            <or>
+              <and>
+                <datadep age="05:00"><cyclestr>&CYCLE_BASEDIR;/@Y@m@d@H{{ slash_ensmem_subdir }}/dynf#fhr#:#fmin#:00.nc</cyclestr></datadep>
+                <datadep age="05:00"><cyclestr>&CYCLE_BASEDIR;/@Y@m@d@H{{ slash_ensmem_subdir }}/phyf#fhr#:#fmin#:00.nc</cyclestr></datadep>
+              </and>
+              <and>
```
gsketefian: Are the dynf and phyf file names always going to include the minute now, i.e. even when SUB_HOURLY_POST is set to FALSE?  If so, then are lines 443-446 needed?

gsketefian: After looking at the output netcdf file names with and without subhourly post that you ran, it seems like the minutes and seconds will not be appended to the file names in the "without" case.  So to simplify the final xml that is generated, the `<or>` can be removed and either the first `<and>` (for subhourly enabled) or the second `<and>` (for subhourly disabled) can be kept, i.e. I'm thinking of replacing lines 438-447 with something like this:
```
            <and>
{% if sub_hourly_post %}
              <datadep age="05:00"><cyclestr>&CYCLE_BASEDIR;/@Y@m@d@H{{ slash_ensmem_subdir}}/dynf#fhr#:#fmin#:00.nc</cyclestr></datadep>
              <datadep age="05:00"><cyclestr>&CYCLE_BASEDIR;/@Y@m@d@H{{ slash_ensmem_subdir}}/phyf#fhr#:#fmin#:00.nc</cyclestr></datadep>
{% else %}
              <datadep age="05:00"><cyclestr>&CYCLE_BASEDIR;/@Y@m@d@H{{ slash_ensmem_subdir}}/dynf#fhr#.nc</cyclestr></datadep>
              <datadep age="05:00"><cyclestr>&CYCLE_BASEDIR;/@Y@m@d@H{{ slash_ensmem_subdir}}/phyf#fhr#.nc</cyclestr></datadep>
{% endif %}
            </and>

```

ush/templates/FV3LAM_wflow.xml
```diff
@@ -405,39 +405,87 @@ MODULES_RUN_TASK_FP script.
 -->
   <metatask name="&RUN_POST_TN;{{ uscore_ensmem_name }}">
```
gsketefian, line 459: The problem with having a task that never launches (run_post_f00000) is that it will throw off the way the workflow end-to-end testing system because that system checks that all workflow tasks have completed successfully before declaring the workflow status a "SUCCESS".  But I think there is a way to fix this for the subhourly-enabled case by splitting the loop over hours into tasks for the first hour and a loop over the remaining hours, as follows:
```
{% if sub_hourly_post %}
  ==> Move lines 458-487 (in the new file) that define the run_post_f00000 task to here.  This portion 
      does not need fhr or fmin because it will just use {{ first_fcst_file_tstr }}.
  ==> For the first hour (fhr=0), loop over the minutes, not starting with the 0th but instead with delta_min:
      <var name="fmin">{% for min in range(delta_min, 60,delta_min) %}{{ " %02d" % min  }}{% endfor %}</var>
      <metatask>
      ...
      </metatask>
{% endif %}
==> Now make the starting hour of the main loop over fhr start on the second hour (fhr=1) if subhourly is 
    enabled; otherwise start on the first hour (fhr=0):
{% if sub_hourly_post %}
    <var name="fhr">{% for h in range(1, fcst_len_hrs+1) %}{{ " %03d" % h  }}{% endfor %}</var>
{% else %}
    <var name="fhr">{% for h in range(0, fcst_len_hrs+1) %}{{ " %03d" % h  }}{% endfor %}</var>
{% endif %}
==> Put lines 410-455 (in the new file) here to do the usual (including looping over all minutes, starting 
    with 0).
```
You can combine the two if-statements above into one if you like.  I didn't because it was clearer to me this way.

gsketefian (changes requested):

gsketefian:
@Jeff-Duda Sorry for the late reply to your changes.  I had actually made the most recent sent of comments mid last-week but apparently forgot to press the right button for you to see them.

If you'd like to have an in-person meeting to work through these, I'm happy to do it.

ush/setup.sh
Jeff-Duda: That is how the code currently looks.

ush/templates/FV3LAM_wflow.xml
Jeff-Duda: addressed

ush/templates/FV3LAM_wflow.xml
Jeff-Duda: addressed

gsketefian:
@Jeff-Duda This looks pretty good.  I'm trying testing it now and may add one or two WE2E tests so that future PRs don't break the subhourly post capability.

In the meantime, could you merge the latest from the develop branch into your branch and push it to your fork?  I just merged develop into my copy of your branch and there were no conflicts, so it should be straightforward.  Thanks.

Jeff-Duda:
Not sure I did the merging correctly, but have a look and see if this is ready.

gsketefian:
@Jeff-Duda I ran tests and had to fix a couple of bugs in the PR.  The biggest change was the rearrangement of the XML.  I created a PR (#[1](https://github.com/Jeff-Duda/regional_workflow/pull/1)) from my fork into yours so you can pull my changes into your fork.  I'm waiting for the 3 WE2E tests I launched to finish; I don't think there will be a problem, but I'll let you know for sure when they're done.  Feel free to retest if you like.

The only other thing left is to make sure that the PR message above is up-to-date because I'm going to use it as the commit message, i.e. I'm going to cut-and-paste when I do the "squash-and-merge", so the PR message is what will show up when you do a "git log".

gsketefian:
@Jeff-Duda The three WE2E tests I ran were all successful.  Please let me know if/when your PR message is updated and we can merge.

Jeff-Duda:
I did perform a merge with your changes a few hours ago. What more do I need to do?

gsketefian:
@Jeff-Duda I don't mean that merge, I mean the final merge for your PR where we (I, really) take all of your commits in this PR, squash them into one, and merge that single commit into the authoritative branch.  This is done via the "Squash and merge" button below.  I have to do that.  That process requires a commit message that will show up when someone does a "git log" on the develop branch, and I'm going to use your PR message above as that message.  So I just want to make sure that it is up-to-date.  For example, I see that it refers to the DELTA_MIN variable, but we changed that to DT_SUBHOURLY_POST_MNTS.  Just update those details and let me know when it's ready and I'll merge it and close this PR.  Thanks.

Jeff-Duda:
Update them where?

JeffBeck-NOAA:
@Jeff-Duda, at the top of this conversation, you'll see the original location where you described the PR (Description of Changes, Tests Conducted, Issue (Optional)).  There's an option to edit that message.

gsketefian:
@Jeff-Duda At the top of this PR page, there is a box containing the message you wrote a while ago that describes the changes this PR makes.  It has sections like "Description of Changes", "Test Conducted", and "Issue (optional)" that you filled out.  Once you find that message, you'll see a button with three dots (...) in the upper-right corner of the box containing the message.  Press that, then press "Edit", and edit your message so it's up-to-date.  Then press the green "Update Comment" button that will show up at the bottom of the message box.  You can repeat this process as many times as you like until you're satisfied that the PR message adequately and correctly described the changes that this PR makes to the repo.  Once you're done, let me know and I'll merge your PR into the authoritative "develop" branch of this repo (I'll use your PR message as the message for the single commit that all your changes will be grouped under).

Jeff-Duda:
Okay, I think I am comfortable with the edits to my original message now. Youc an move forward.

gsketefian:
@Jeff-Duda I just merged this PR.  In the PR message, I added a few lines about my bug fixes and noted that this PR changes the way the UPP grib2 output files are named even in the case of SUB_HOURLY_POST set to "FALSE", in which case "00" minutes is added to the names of the files.

I will submit another PR soon that will introduce a couple of WE2E tests involving subhourly post-processing.

PR #444, opened 2021-02-23
Changed multiprocessing pool call to serial call to avoid issue on Ma…

ywangwof:
…cOS/Windows

## DESCRIPTION OF CHANGES: 
To eliminate the multiprocessing pool call in the python plotting scripts and change them to serial calls since we have only one domain in the SRW Application

## TESTS CONDUCTED: 
Tests have been performed on Odin and a MacOS machine.

## ISSUE (optional): 
The original multiprocessing call will cause this error message on MacOS,
```
An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.
        This probably means that you are not using fork to start your
        child processes and you have forgotten to use the proper idiom
        in the main module:
            if __name__ == '__main__':
                freeze_support()
                ...
        The "freeze_support()" line can be omitted if the program
        is not going to be frozen to produce an executable.
```
It is because on Windows or MacOS the subprocesses will import (i.e. execute) the main module at start, which creates subprocesses recursively. It can be avoided by inserting an if __name__ == '__main__': guard in the main module. However, this python script was written mainly as a running script (the same style as Bash etc.) and it is not easy to wrap them in a main function.
The easy fix is just to eliminate the use of multiprocessing. Anyway, we only have one domain for SRW Application right now.

## Notes:

The changes are mainly in the `main` function of _plot_allvars.py_ & _plot_allvars_diff.py_. All other changes are just the removing of extra blank spaces at end of all code lines (a feature of my editor and a code convention I believe we should adapt for code development).

## CONTRIBUTORS (optional): 
It has been discussed during the testing meeting with @BenjaminBlake-NOAA, @climbfuji and @JeffBeck-NOAA  etc.

### DIFF
--- a/ush/Python/plot_allvars.py
+++ b/ush/Python/plot_allvars.py
@@ -1,6 +1,6 @@
 ################################################################################
 ####  Python Script Documentation Block
-#                      
+#
 # Script name:       	plot_allvars.py
 # Script description:  	Generates plots from FV3-LAM post processed grib2 output
 #			over the CONUS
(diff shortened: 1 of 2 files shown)


BenjaminBlake-NOAA:
@ywangwof Thanks for submitting, I will test on Hera and then approve the PR assuming everything is fine.

BenjaminBlake-NOAA (approved):
The changes worked for me on Hera, so I'll go ahead and approve

PR #451, opened 2021-03-02
Feature/refc updates for release/public-v1

jwolff-ncar:
## DESCRIPTION OF CHANGES: 
With the updates to the documentation Graphics chapter, paths and environments needed to be updated in the example batch scripts (sq_job(_diff).sh). This makes the batch scripts consistent with the documentation

A line was mistakenly added to the diag_table.FV3_GFS_v15p2 to output refl_10cm in a previous PR. This is a Thompson MP field and is not included in the GFDL MP. Given this, refl_10cm remains empty in the output, and post uses it for REFC instead of calculating the reflectivity using the microphysics tracers. In that same PR a file was erroneously added and that is being removed diag_table.FV3_RRFS_v1alpha.new.

## TESTS CONDUCTED: 
Tests run on Hera. Would greatly appreciate assistance with additional tests on other platforms.


climbfuji (commented):
No problem with the sq_job / sq_job_diff script changes, but can we be 100% sure that the diag_table changes don't break anything?

JeffBeck-NOAA:
@climbfuji, we just need to be extremely careful about those changes.  I don't think it will be an issue, but we need to thoroughly test on different platforms through Python plotting to ensure the composite reflectivity field looks OK.

jwolff-ncar:
Testing successful on Cheyenne.

llpcarson:
Testing successful on jet.

JeffBeck-NOAA:
Works on Orion, approving.

JeffBeck-NOAA (approved):

gsketefian:
@jwolff-ncar Can you remove commented-out lines 198-280 from regional_workflow/ush/load_modules_run_task.sh?  It is a bunch of stuff that was left in case things didn't work with the new build/module system, but now it's just a bunch of clutter.  Thanks.

jwolff-ncar:
@gsketefian While I am sure it is harmless, I prefer to keep this PR as isolated as possible. If others disagree with me, I can remove those lines. It certainly can be removed from develop so we don't forget in the future. The goal is to get emergency fixes into the release at this very late stage in the game. We don't have time/resources to continue testing when another (even very small) change goes in.

jwolff-ncar:
@mkavulich Any results on a tier 3 platform to report? If we have a thumbs up on that I would be comfortable merging.

JeffBeck-NOAA:
Works on Gaea too.

mkavulich:
Testing was successful on MacOS, aside from the problem with graphics after 24 hours. But on that subject.... 

I discovered something concerning about the python plotting script issue : it is *not* an issue with the script: it appears to be an issue with the grib2 input files: after 24hr, all grib2 files have identical data. Presumably this means there is an error in the post script. Did we want to attempt to include a fix in this PR?

jwolff-ncar:
Thanks for the update @mkavulich. Bad news about the grib2 files. We will need to figure out the cause of this as soon as possible (obviously) and go from there. This is likely a show stopper.

mkavulich:
@jwolff-ncar Actually, the problem is even worse than I thought: the **ncfiles** from model output are identical after 24 hours. For example, if you use "ncdiff" to compare dynf024.nc and dynf025.nc, the fields are identical. Differences can be seen in fields up to 24 hours, but after dynf024.nc, they are all identical.

This is clearly a serious problem. I am not sure how this could be happening, but I doubt this is a simple scripting fix, and it's well beyond my capability to debug, especially at this late hour.

gsketefian:
@mkavulich @jwolff-ncar @JeffBeck-NOAA In the scripts, the only places I can think of that might (even remotely) cause this frozen-ness problem are:
1) The diag_table file.
2) The model_configure file.
3) The output hours in the run_post task in FV3LAM_wflow.xml.
I checked these in the GST 25km run in which I first saw this problem, and they seem ok.  But it would be good if someone else took a look as well.

ush/templates/FV3.input.yml
```diff
@@ -40,7 +40,7 @@ FV3_RRFS_v1alpha:
     lsm: 2
     lsoil_lsm: 4
   atmos_model_nml:
```
gsketefian, line 42: Remove these two lines since this file is only for CHANGES to the default settings in input.nml.FV3 (and fdiag is already set to 1 in that file).

ush/templates/FV3.input.yml
```diff
@@ -52,7 +52,7 @@ FV3_RRFS_v1beta:
...
```
gsketefian, line 54: Remove lines 54-55.

ush/templates/FV3.input.yml
```diff
@@ -187,7 +187,7 @@ FV3_GFS_2017_gfdlmp_regional:
 
 FV3_GFS_v15p2:
   atmos_model_nml:
```
gsketefian, line 189: Remove lines 189-190.

ush/templates/FV3.input.yml
```diff
@@ -272,7 +272,7 @@ FV3_GFS_v15p2:
 
 FV3_GFS_v16beta:
   atmos_model_nml:
-    fdiag: 3
+    fdiag: 1
```
gsketefian, line 275: Remove only line 275 (not 274).

ush/templates/FV3.input.yml
```diff
@@ -343,7 +343,7 @@ FV3_GFS_v16beta:
 
 FV3_CPT_v0:
   atmos_model_nml:
```
gsketefian, line 345: Remove lines 345-346.

ush/templates/FV3.input.yml
jwolff-ncar: In order to be the least intrusive as possible we will simply leave fdiag hard-coded to 1 for this release. Your suggested edits (or some other approach) can be considered for the develop branch.

JeffBeck-NOAA:
fdiag change works on Jet.  Approving.

climbfuji (approved):
Worked in all my tests so far.

llpcarson:
same, works for cheyenne, hera.  Orion still waiting in the queue

PR #458, opened 2021-03-24
Rename GSD_RAP13km grid and adjust its write-component grid

gsketefian:
## DESCRIPTION OF CHANGES:
* Rename GSD_RAP13km grid to RRFS_NA_13km.
* In ush/set_predef_grid_params.sh:
  * Change parameters of the (newly renamed) RRFS_NA_13km grid to make its write-component grid be completely within the native grid.
  * For the (newly renamed) RRFS_NA_13km grid, automate calculation of WRTCMP_dlon and WRTCMP_dlat from ESGgrid_DELX and ESGgrid_DELY, respectively.
* NCL scripts:
  * Make modifications to make scripts more general and have the find_wrtcmp_grid_params.ncl script work for a write-component grid of type "rotated_latlon".
  * Remove unused code in NCL scripts.
* Fix up comments.

## TESTS CONDUCTED: 
Ran the (renamed) WE2E test grid_RRFS_NA_13km on Hera.  Was successful.

## CONTRIBUTORS (optional): 
This grid was also tested by @jaymes-kenyon on Jet.


ush/set_predef_grid_params.sh
```diff
@@ -317,8 +334,10 @@ case ${PREDEF_GRID_NAME} in
 #      WRTCMP_lat_lwr_left="-13.56176982"
 #      WRTCMP_lon_upr_rght="18.47206579"
 #      WRTCMP_lat_upr_rght="13.56176982"
-#      WRTCMP_dlon="0.11691181"
-#      WRTCMP_dlat="0.11691181"
+##      WRTCMP_dlon="0.11691181"
+##      WRTCMP_dlat="0.11691181"
+#      WRTCMP_dlon=$( printf "%.9f" $( bc -l <<< "(${ESGgrid_DELX}/${radius_Earth})*${degs_per_radian}" ) )
```
JeffBeck-NOAA, line 339: @gsketefian, do lines 337-338 still need to be included?  Did you intend to change the RRFS_AK_13km grid here (that's the section these changes are in)?

gsketefian: Yes, that was intentional.  Everywhere there is a hard-coded value for WRTCMP_dlon/dlat for an ESG grid, I meant to replace with the automated calculation using the printf/bc statements so there won't be a need to copy/paste a hard-coded value.  This just happened to be in a commented-out section; I want to keep the old values there so we know what values we'd like those print statements to generate.  The RRFS_AK_13km grid is not finalized because I don't think it's the same one that GSL is using.  Needs further testing; just want to keep the original values around for when we get to testing it.

JeffBeck-NOAA (approved):

Issue #452, opened 2021-03-11
Explicitly define OMP_STACKSIZE and OMP_NUM_THREADS


JeffBeck-NOAA:
A bug was recently identified where chgres_cube will fail in ESMF if OMP_STACKSIZE and OMP_NUM_THREADS are not explicitly defined for Hera.  Only a few platforms/tasks currently have these defined in the ex-scripts.  It's possible that other tasks on other platforms could fail aside from chgres_cube if these aren't defined.

The settings that resulted in a successful chgres_cube job on Hera for the new NA 3-km domain are:

export OMP_STACKSIZE=1024m
export OMP_NUM_THREADS=1

Optimal settings should be identified for each platform and task, then placed in all ex-scripts. These should be made user-defined values for each task so that they can be tweaked if necessary in the config.sh script.

@christinaholtNOAA @LarissaReames-NOAA

JeffBeck-NOAA:
This was resolved with PR #459.  Closing this issue.

PR #459, opened 2021-03-31
Add functionality for user-defined OMP settings and for the RUN_FCST "--cpus-per-task" value.

JeffBeck-NOAA:
## DESCRIPTION OF CHANGES: 
Create user-defined OMP settings for each workflow task, and a user-defined "--cpus-per-task" for the RUN_FCST task.  Default values can be found in config_defaults.sh.

## TESTS CONDUCTED: 
Tested on Hera with both 3-km and 25-km domains, running FV3_GFS_v15p2 and FV3_GSD_SAR SDFs, respectively.  Tested on Jet with the 3-km domain.

## ISSUE (optional): 
This PR resolves Issue #452 and addresses a related [UFS Forum discussion thread](https://forums.ufscommunity.org/threads/srw-forecasting-model-performance-jet).

## CONTRIBUTORS (optional): 
@LarissaReames-NOAA, @christinaholtNOAA, @ywangwof, @dmwright526, and Ted Mansell


scripts/exregional_run_fcst.sh
```diff
@@ -122,8 +132,7 @@ case $MACHINE in
   "JET")
     ulimit -s unlimited
     ulimit -a
-    APRUN="srun"
-    OMP_NUM_THREADS=4
+    APRUN="mpiexec"
```
christinaholtNOAA: Why mpiexec here? In the release branch, it was changed to mpirun. I'm referring to PR #446.

JeffBeck-NOAA: This was reverted to "srun" in this PR after testing on Jet showed that timing "mpiexec" and "srun" are nearly identical, and that "mpirun" can be prone to hang.

scripts/exregional_run_fcst.sh
```diff
@@ -83,6 +83,17 @@ print_input_args valid_args
 #
 #-----------------------------------------------------------------------
 #
+# Set and export OpenMP variables.
+#
+#-----------------------------------------------------------------------
+#
+export KMP_AFFINITY=${KMP_AFFINITY_RUN_FCST}
+export OMP_NUM_THREADS=${OMP_NUM_THREADS_RUN_FCST} #Needs to be 1 for
```
gsketefian: Can remove comment about dynamic build since we can't do that any more anyway.

ush/config_defaults.sh
```diff
@@ -1579,4 +1579,61 @@ COMPILER="intel"
...
+# OMP_STACKSIZE_*:
+# Controls the size of the stack for threads created by the OpenMP 
+# implementation.
+#
+# CPUS_PER_TASK_RUN_FCST:
+# Sets the number of MPI tasks per CPU for the RUN_FCST task. 
+#
+#-----------------------------------------------------------------------
+#
+KMP_AFFINITY_MAKE_GRID="scatter"
```
gsketefian: For clarity, can you leave a blank line between groups of variables that pertain to each rocoto task?  Thx.

scripts/exregional_make_ics.sh
```diff
@@ -72,6 +72,16 @@ print_input_args valid_args
 #
 #-----------------------------------------------------------------------
 #
+# Set and export OpenMP variables.
+#
+#-----------------------------------------------------------------------
+#
+export KMP_AFFINITY=${KMP_AFFINITY_MAKE_ICS}
```
gsketefian: Can you try a run without exporting?  We shouldn't do it if it's not absolutely necessary to keep from polluting the environment with unnecessary variables.  I don't think it is necessary since below, we're calling the executable directly (as opposed to another script that then calls the executable).  

Same in other scripts.

scripts/exregional_run_fcst.sh
JeffBeck-NOAA: Removed.

ush/config_defaults.sh
JeffBeck-NOAA: Changed.

scripts/exregional_make_ics.sh
JeffBeck-NOAA: Just confirmed (with a negative number of threads) that these commands work without exporting.  I removed the export commands from each ex-script.

scripts/exregional_make_orog.sh
```diff
@@ -76,16 +76,6 @@ print_input_args valid_args
 #
 #-----------------------------------------------------------------------
 #
-# The orography code runs with threads.  On Cray, the code is optimized
```
gsketefian, line 79: This comment indicates that the orography code(s) (there is more than one executable called in this script) is parallelized over threads (OMP) (but probably not with MPI). Maybe the make_orog task should keep these parameters, and on WCOSS_CRAY, set OMP_NUM_THREADS_MAKE_OROG to 6 (don't know about the values for other platforms).  Best way to know is to check the code(s) in UFS_UTILS to see if they contain OMP directives.

JeffBeck-NOAA: It looks like only orog.x has OMP directives in the code.  I will add the OMP user-defined variable options back into the make_orog ex-script.

JeffBeck-NOAA: Done.

scripts/exregional_make_orog.sh
```diff
@@ -76,13 +76,14 @@ print_input_args valid_args
...
-# for six threads.  Do not change.
+# Set and export OpenMP variables.  The orog executable runs with OMP. On
+# WCOSS (Cray), it is optimized for six threads, which is the default.
 #
 #-----------------------------------------------------------------------
 #
-export OMP_NUM_THREADS=6
-export OMP_STACKSIZE=2048m
+KMP_AFFINITY=${KMP_AFFINITY_MAKE_OROG}
```
gsketefian, line 84: Below, for WCOSS_CRAY, there is a line KMP_AFFINITY set here will be overwritten with "disabled".  Should that line (108?) be removed?

JeffBeck-NOAA: Yes, thank you.  It's now removed.

scripts/exregional_make_orog.sh
```diff
@@ -76,13 +76,14 @@ print_input_args valid_args
...
+OMP_NUM_THREADS=${OMP_NUM_THREADS_MAKE_OROG}
+OMP_STACKSIZE=${OMP_STACKSIZE_MAKE_OROG}
 #
```
gsketefian, line 87: This comment is actually for lines further below that have "export APRUN.=.." and not directly related to your changes.  Can you remove the "export"s from those to be consistent with what we did above?  Thx.

JeffBeck-NOAA: All export commands are now removed.  I also updated the comments in the ex-scripts to remove any mention of exporting.  I tested make_orog one more time on Jet and everything looks OK.

gsketefian: Many of the other ex-scripts have unnecessary export commands that we should get rid of at some point, but let's get this PR in.

gsketefian (approved):

JeffBeck-NOAA:
Yes.  We should probably make that an issue in regional_workflow.

PR #473, opened 2021-04-08
Make EXTRN_MDL_SYSBASEDIR_[ICS|LBCS] user-specifiable parameters

gsketefian:
## DESCRIPTION OF CHANGES: 
Currently, the workflow variables EXTRN_MDL_SYSBASEDIR_ICS and EXTRN_MDL_SYSBASEDIR_LBCS cannot be specified by the user.  They get set according to the platform and external model (for ICSs and LBCs, respecitively) specified by the user.

@chunhuazhou requested that these be user-specifiable.  Thus, this PR modifies the script set_extrn_mdl_params.sh such that if one or both of these are specified by the user in the experiment configuration file (config.sh), then those values are used, and if they are not, default values based on the platform and external model are used (as is currently the case).

This PR also adds a new WE2E test named specify_EXTRN_MDL_SYSBASEDIR_ICS_LBCS in which EXTRN_MDL_SYSBASEDIR_ICS and EXTRN_MDL_SYSBASEDIR_LBCS get set to user-specified directories rather than the default system directory for the machine and the specified external models.  The specific test added uses the FV3GFS for both the external models (i.e. the one for ICs and the one for LBCs), and the user-specified directory it specifies (which is the same for ICs and LBCs) via the run_experiments.sh script is on Hera and contains files (for a single starting date and hour) from the FV3GFS external model in the same format as on the system directory on Hera (which is at /scratch1/NCEPDEV/rstprod/com/gfs/prod).

Note that this PR modifies the testing script (run_experiments.sh) to enable it to handle the case of user-specified values for EXTRN_MDL_SYSBASEDIR_ICS and EXTRN_MDL_SYSBASEDIR_LBCS.

## TESTS CONDUCTED:
The newly added WE2E test (named specify_EXTRN_MDL_SYSBASEDIR_ICS_LBCS) was successfully run on Hera.

## CONTRIBUTORS (optional): 
@chunhuazhou


JeffBeck-NOAA (approved):
Reviewed and looks good.  Approving.

PR #479, opened 2021-04-16
develop: Changes to processing and output related to sub-hourly simulation settings

JeffBeck-NOAA:
## DESCRIPTION OF CHANGES: 

- Incorporate changes to allow for previous file/directory naming convention when running in hourly mode, with modifications only applicable to sub-hourly mode.  
- Change job and log file names to be mode-dependent (sub-hourly or hourly)
- Fix a number of bugs related to handling of "fhr" and "fmn" depending on simulation mode.
- Add two new sub-hourly WE2E tests.
- Changes made to the XML workflow template to avoid blocks of blank lines due to Jinja "if" statements.

## TESTS CONDUCTED: 
Tested successfully with three WE2E tests (hourly, sub-hourly, and sub-hourly ensemble) on Hera, here:
- /scratch2/BMC/det/beck/FV3-LAM/expt_dirs/grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
- /scratch2/BMC/det/beck/FV3-LAM/expt_dirs/subhourly_post
- /scratch2/BMC/det/beck/FV3-LAM/expt_dirs/subhourly_post_ensemble_2mems

## CONTRIBUTORS (optional): 
The majority of this PR was originally in a feature branch in @gsketefian's fork of regional_workflow.


chan-hoo:
@JeffBeck-NOAA, this PR resolves the issue on metpuls, too. I got the result from the sample metplus case on Hera. Thank you!

JeffBeck-NOAA:
> @JeffBeck-NOAA, this PR resolves the issue on metpuls, too. I got the result from the sample metplus case on Hera. Thank you!

@chan-hoo, fantastic!  Thank you for confirming!

scripts/exregional_run_post.sh
```diff
@@ -252,69 +268,79 @@ EOF
...
+# generated by UPP.
 #
 len_fhr=${#fhr}
 if [ ${len_fhr} -eq 2 ]; then
   post_fhr=${fhr}
 elif [ ${len_fhr} -eq 3 ]; then
   if [ "${fhr:0:1}" = "0" ]; then
     post_fhr="${fhr:1}"
+# What should happen in the "else" case?  Would just setting post_fhr to 
+# fhr work?  Need to test.
```
jwolff-ncar, line 302: You may want to clean up this comment if it has been tested.

jwolff-ncar (commented):
I was also able to run through the verification tasks with this update. 

Would be good to have @Jeff-Duda test this PR to confirm it works for his test case.

scripts/exregional_run_post.sh
JeffBeck-NOAA: @gsketefian, what do you think about this comment?  It would require 100+ forecast hour data to test.

I'm actually not quite sure why all this logic is necessary to set "post_fhr".  I only see it used on this line below:

post_fn_suffix="GrbF${post_fhr}${dot_post_mn_or_null}"

gsketefian: I think this if-statement was originally put in by @Jeff-Duda and I may have tweaked it, I can't remember exactly.  It ensures that post_fhr contains only two digits, probably because for forecast hours less than 100, the file names UPP generates contain two digits for the forecast hour.  But what does it do for forecast hours of 100 and more?  I guess it would just use the 3-digit forecast hour, but that's never been tested in the FV3LAM workflow.  That's why I put in this comment to draw attention to this.

JeffBeck-NOAA: OK, let's leave it as is for now.  It may work fine with post_fhr=fhr, but we'll save that for a future PR.

JeffBeck-NOAA:
> I was also able to run through the verification tasks with this update.
> 
> Would be good to have @Jeff-Duda test this PR to confirm it works for his test case.

If @Jeff-Duda doesn't have time to test things, the PR did succeed with a sub-hourly test case (one of the new WE2E tests).

ush/setup.sh
```diff
@@ -1110,6 +1074,98 @@ fi
...
+#   
+# This restriction will be removed in a future version of the workflow, 
+# For now, check that DT_SUBHOURLY_POST_MNTS is one of the above values.
+#
+  if [ "${SUB_HOURLY_POST}" = "TRUE" ]; then
+    check_var_valid_value "DT_SUBHOURLY_POST_MNTS" "valid_vals_DT_SUBHOURLY_POST_MNTS"
+  fi
+
+fi
+print_info_msg "hello!"
```
gsketefian: You can remove these Hello and exit lines.

ush/fill_jinja_template.py
```diff
@@ -46,7 +46,7 @@
```
gsketefian: The changes to this python script were needed when I was trying to use "include"s in the jinja template xml (ush/templates/FV3LAM_wflow.xml).  Since those "include"s are not part of this PR, I don't think the changes to this script are needed (they don't make sense without the "include"s).  I recommend undoing the changes in this script and retesting (just comparing the generated XMLs should be sufficient with the ones for the tests you've already done should be sufficient).

JeffBeck-NOAA: Thanks for catching this.  I made the changes (aside from the two typos) and pushed the change.  The XML looks the same (except for directory differences).

ush/setup.sh
JeffBeck-NOAA: Fixed!

gsketefian (approved):

PR #480, opened 2021-04-27
Add the RRFS_NA_3km pre-defined domain to the SRW App

JeffBeck-NOAA:
## DESCRIPTION OF CHANGES: 
This PR adds an option to run the pre-defined RRFS_NA_3km domain to the SRW App.  Note that in order to run longer simulations, the model code should be compiled in 32-bit mode, not 64-bit.  Changes have also been made to k/n_split values in the namelist template which optimize run time.  Nodes/core settings must be modified for chgres_cube and post due to the size of this domain.  A WE2E test was added and more information on all of these settings can be found within the related config.sh script. 

Note that CPUS_PER_TASK_RUN_FCST is changed from "4" to "2" in this PR.  Setting this field to "4" was doubling the requested nodes for the run_fcst task.  For example, a 3-km CONUS run that normally requests 25 nodes (based on predefined layout_x/y values) was asking for 50, simply because CPUS_PER_TASK_RUN_FCST=4, which was unacceptable.  When it is set to "2", the number of nodes remains unchanged, in line with the layout_x/y values.  EMC is using CPUS_PER_TASK_RUN_FCST=2 for their runs.

## TESTS CONDUCTED: 
Multiple tests on Hera were run, including a full 36-hr forecast here: /scratch2/BMC/det/beck/FV3-LAM/expt_dirs/test_RRFS_NA_3km_36hr

## CONTRIBUTORS (optional): 
Thanks are due to @JamesAbeles-NOAA for his recommendations for build/namelist changes and help troubleshooting run times.  Thanks to @BenjaminBlake-NOAA and @JacobCarley-NOAA for their help with the domain configuration.


ush/config_defaults.sh
```diff
@@ -1657,7 +1657,7 @@ KMP_AFFINITY_RUN_FCST="scatter"
 OMP_NUM_THREADS_RUN_FCST="4"
 OMP_STACKSIZE_RUN_FCST="1024m"
 
-CPUS_PER_TASK_RUN_FCST="4"
+CPUS_PER_TASK_RUN_FCST="2"
```
mkavulich, line 1660: @JeffBeck-NOAA I'm testing this PR on Cheyenne right now. Aside from that, my only comment is about this line: Can you provide a brief justification for this change to the defaults in your PR message? This (along with the other changes you already mention) has the potential to impact others significantly.

JeffBeck-NOAA: Thanks @mkavulich.  Good catch on this.  Setting this field to "4" was doubling the requested nodes for the run_fcst task.  For example, a 3-km CONUS run that normally requests 25 nodes (based on predefined layout_x/y values) was asking for 50, simply because CPUS_PER_TASK_RUN_FCST=4, which was unacceptable.  When it is set to "2", the number of nodes remains unchanged, in line with the layout_x/y values.  EMC is using CPUS_PER_TASK_RUN_FCST=2 for their runs.  I can add this to the PR message.

tests/baseline_configs/config.grid_RRFS_NA_3km.sh
```diff
@@ -0,0 +1,49 @@
+RUN_ENVIR="community"
+PREEXISTING_DIR_METHOD="rename"
+
+PREDEF_GRID_NAME="RRFS_NA_3km"
+QUILTING="TRUE"
+
+CCPP_PHYS_SUITE="FV3_GSD_SAR"
```
mkavulich, line 7: One more comment: Should this use the RRFS_v1alpha suite instead?

JeffBeck-NOAA: Yes, it probably should be.  I can change it in the PR, but with the problems we're having with RRFS_v1alpha at the moment, we can't test with that suite (at least not on Hera).

JeffBeck-NOAA:
The pre-defined RRFS_NA_3km domain has been incorporated into PR #492.  Closing this PR.

JeffBeck-NOAA:
The pre-defined RRFS_NA_3km domain has been incorporated into PR #492. Closing this PR.

Issue #490, opened 2021-05-12
GST_release_public_v1 test occasionally fails due to running out of wallclock time on Hera


mkavulich:
We need to either increase the wall time or increase the cores allocated for the run_fcst step.

mkavulich:
This issue is only seen with newer versions of the weather model compared to the current hash from the ufs-srweather-app.  The problem exists in https://github.com/ufs-community/ufs-weather-model/commit/2f1c8e1883370b30828dea87d4250e10c35499f4 and does not exist in https://github.com/ufs-community/ufs-weather-model/commit/ea8a7aa9a8c6b8f85f50c20b7567a846c7e81fd9. It will take some time to narrow down exactly what change resulted in this longer runtime. I am also investigating if there are other tests/domains/suites that are affected.

mkavulich:
This ended up not being an issue with the weather model, but with ESMF libraries. Using ESMF 8.1.0 beta snapshot 47 (which was loaded by the weather model by default for many hashes) resulted in very long runtimes due to a bug that caused threading to be disabled (OMP_NUM_THREADS=1). Updating all modulefiles to use ESMF 8.1.1 fixed this issue, and restored runtimes to near previous levels.

mkavulich:
Turns out I closed this issue prematurely. The GST still times out on Orion, even when compiling with ESMF v8.1.1. This could be because of a machine-specific issue, but it should be investigated before closing this issue.

mkavulich:
Issue migrated to ufs-srweather-app repository: https://github.com/ufs-community/ufs-srweather-app/issues/347

PR #492, opened 2021-05-12
Update for latest version of weather model (hash 2f1c8e1), add RRFS_NA_3km pre-defined domain, update timestep and MPI settings

mkavulich:
## DESCRIPTION OF CHANGES: 
This PR accomplishes three things: 

1. A new pre-defined domain (RRFS_NA_3km) has been added to the SRW App. Nodes/core settings must be modified for chgres_cube and post due to the size of this domain.  A WE2E test was added and more information on all of these settings can be found within the related config.sh script (tests/baseline_configs/config.grid_RRFS_NA_3km.sh). 
2. The default k_split value is updated for a faster model integration. With k_split=2, we see model integration close to X% faster than the previous settings for the same weather model hash. **This will not affect physics suites that have specified other k_split values**
3. In order to properly run the above domain with the intended FV3_RRFS_v1alpha physics suite, the weather model needed to be updated to a more recent hash. This more up-to-date weather model version also has renamed the FV3_GFS_v16beta suite to FV3_GFS_v16; this required a number of changes to the workflow and end-to-end tests. In addition, several changes to default settings are occurring in this PR. Changes have also been made to k/n_split values in the namelist template which optimize run time.  CPUS_PER_TASK_RUN_FCST is changed from "4" to "2" in this PR.  Setting this field to "4" was doubling the requested nodes for the run_fcst task.  For example, a 3-km CONUS run that normally requests 25 nodes (based on predefined layout_x/y values) was asking for 50, simply because CPUS_PER_TASK_RUN_FCST=4, which was unacceptable.  When it is set to "2", the number of nodes remains unchanged, in line with the layout_x/y values.  EMC is using CPUS_PER_TASK_RUN_FCST=2 for their runs, so this should be uncontroversial.

Note that the changes for the new RRFS_NA_3km domain and default decomposition/CPU settings were originally proposed by @JeffBeck-NOAA in PR #480, however since that PR requires the updates to the weather model in order to be tested with the intended physics suite, I have incorporated those changes into mine. As Jeff is on vacation right now I hope he won't mind me stealing his thunder :) 

This PR will need to be accompanied by changes in the ufs-srweather-app for updating the weather model hash and incorporating some necessary build changes (including compiling with 32-bit reals by default); this PR has been created (https://github.com/ufs-community/ufs-srweather-app/pull/140) but it still a draft pending some platform-specific fixes and the merger of this PR.

## TESTS CONDUCTED: 
For the initial changes for PR #480, multiple tests on Hera were run, including a full 36-hr forecast here: /scratch2/BMC/det/beck/FV3-LAM/expt_dirs/test_RRFS_NA_3km_36hr

With the additional changes and updates to the weather model, and updates to the Hera environment file, all end-to-end tests (aside from nco tests) were run on Hera (intel). There were a few pre-existing failures, and aside from an occasional GST failure due to wallclock time issues (see #490) the only new failures were for grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_GSD_SAR, grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_HRRR, and grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_RRFS_v1beta, which all had a new failure in make_ics and make_lbcs. Currently investigating this issue, though it is almost certainly related to the build environments which need to be addressed in https://github.com/ufs-community/ufs-srweather-app/pull/140

Tests are ongoing on Cheyenne, Jet and Orion; will update with the results once those are complete. However, I believe the Hera tests are sufficient for this PR, as any failures specific to those platforms need to be addressed at the app level.

## CONTRIBUTORS: 
@JeffBeck-NOAA authored the half of these changes originating from #480, and offered the following credits on his original PR:

Thanks are due to @JamesAbeles-NOAA for his recommendations for build/namelist changes and help troubleshooting run times.  Thanks to @BenjaminBlake-NOAA and @JacobCarley-NOAA for their help with the domain configuration.


tests/baseline_configs/config.grid_RRFS_NA_3km.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# The model should be built in 32-bit mode (64-bit will result in much
+# longer run times.
+
+# Use k_split=5 and n_split=2, the previous namelist values (k_split=4
+# and n_split=5) will result in significantly longer run times.
+
+# For Hera:
+# 1) With 64-bit build and k_split=4/n_split=5, each time step takes ~21 seconds
+# 2) 1) with k_split=5/n_split=2, each time step takes ~15 seconds
```
gsketefian: Remove extraneous list numbers.

tests/baseline_configs/config.grid_RRFS_NA_3km.sh
```diff
@@ -0,0 +1,49 @@
...
+# 3) 2) with 32-bit build, each time step takes ~6 seconds
+
+NNODES_MAKE_ICS="12"
```
gsketefian, line 31: Are these settings specific to Hera?  If we need to change these according to platform, we can put that kind of code in run_experiments.sh.  That's where I've been doing platform-specific differentiation of settings.

tests/baselines_list.txt
```diff
@@ -43,17 +43,18 @@ grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_HRRR
 grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_RRFS_v1beta
 grid_RRFS_CONUS_25km_modify_DT_ATMOS_LAYOUT_XY_BLOCKSIZE
 grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
-grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16beta
+grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
 grid_RRFS_CONUS_3km_ics_HRRR_lbcs_RAP_suite_GFS_v15p2
 grid_RRFS_CONUS_3km_ics_HRRR_lbcs_RAP_suite_GSD_SAR
 grid_RRFS_CONUS_3km_ics_HRRR_lbcs_RAP_suite_HRRR
 grid_RRFS_CONUS_3km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
 grid_RRFS_NA_13km
```
gsketefian, line 51: To be consistent with the naming convention used for most of the other tests (especially ones that have been added recently), can we add to the test name the external models used and the physics suite?  I guess that goes for both tests grid_RRFS_NA_13km and grid_RRFS_NA_3km.  Later on, I'm thinking we may change the naming convention, maybe depending on the category a test is in, but that'll take more thought.

ush/templates/input.nml.FV3
```diff
@@ -85,15 +85,15 @@
     hord_vt = 6
     hydrostatic = .false.
     io_layout = 1,1
-    k_split = 4
+    k_split = 5
```
gsketefian: So the new k/n_split values are better (or at least not worse) for all grids?  They are not at all grid-dependent, or maybe even suite-dependent?

tests/baseline_configs/config.grid_RRFS_NA_3km.sh
mkavulich: @JeffBeck-NOAA Can better answer these questions since these changes came from him.

I think the problem is specific to the domain, which is very large and high-resolution, requiring more nodes for make_ics and make_lbcs for memory reasons. I assume that would mean this is required on all platforms.

I do think that these settings are less than ideal.  Making PPN_MAKE_ICS and PPN_MAKE_LBCS lower means we will be under-utilizing the nodes. 

We ddo need to have a separate conversation about making PPN_RUN_FCST (and maybe every PPN setting) platform specific in defaults. Right now we are under-utilizing nodes on most platforms using default settings. Maybe this can be rolled in to issue #452 since the OMP settings will have to be taken into account as well?

tests/baselines_list.txt
mkavulich: I think that's a good idea, but I think that would be best handled in a separate PR, since this is a pre-existing problem unrelated to my changes.

ush/templates/input.nml.FV3
mkavulich: @JeffBeck-NOAA will be able to answer this better, but I did notice that the Graduate Student Test now takes longer to run, often hitting the wallclock limit (#490). This may be due to the different k_split and n_split settings, but I did not do a rigorous comparison between the current develop branch and my changes.

I think it is definitely necessary to address these concerns but I do not know enough about the inner workings of the dycore to have an educated opinion.

tests/baseline_configs/config.grid_RRFS_NA_3km.sh
JeffBeck-NOAA: @gsketefian, what do you mean?  Each list number builds on the last (e.g., #2 is #1 with the additional changes).

tests/baseline_configs/config.grid_RRFS_NA_3km.sh
JeffBeck-NOAA: These settings are required for all platforms due to the domain size and resolution.  The PPN_MAKE_ICS and PPN_MAKE_LBCS values are a specific chgres_cube requirement for large domains (need fewer processes but massive amounts of memory) due to an ESMF limitation/memory bug.

ush/templates/input.nml.FV3
JeffBeck-NOAA: @gsketefian, yes these k/n_split values are better for all grids and are the recommended values from EMC.  I'm not sure why the GST would take longer to run.  Was #490 a problem before this PR?

tests/baseline_configs/config.grid_RRFS_NA_3km.sh
mkavulich: @JeffBeck-NOAA Thanks for the clarification. In that case should these be handled somewhere in the generate workflow calling tree in order to have these settings as a default for this domain specifically? Maybe that's something best handled in an issue and follow-up PR in the future?

JeffBeck-NOAA: @mkavulich, yeah, this is where things can get really thorny.  Do we want to fill the generate script with nested if/case statements for different domains, resolutions, platforms for changes to PPN for individual tasks?  I wasn't sure if we wanted to go that route or just have users source the specific WE2E config.sh file when they want to run this domain?  I think this definitely deserves an issue and potential follow-up PR.

ush/templates/input.nml.FV3
JeffBeck-NOAA: @gsketefian, @mkavulich, I have confirmed that the n_split/k_split values that I received need to be flipped.  So it should be k_split=2/n_split=5 and not k_split=5/n_split=2.  This can go in a future PR, but @mkavulich, this is likely why your GST run is slower.  If we want, we can keep the previous values for now (k_split=4/n_split=5).

mkavulich: I think I'd rather fix it here if this is what needs to go in. I'll swap those values and see what the result is. Assuming it is not slower I'll apply those changes to this PR. 

I am actually not sure if #490 was a specific problem before making these changes but I did notice that my unmodified runs were quite close to reaching the wallclock limit (within a few minutes).

JeffBeck-NOAA: Sounds good to me!  It should be faster for all WE2Es, including the GST.

tests/baseline_configs/config.grid_RRFS_NA_3km.sh
mkavulich: @JeffBeck-NOAA Can you confirm that these notes are still correct? k_split=5/n_split=2 resulted in a speed up even though the ideal settings should be switched? If so I will add these test results to the final commit message (along with the results for the switched setting).

I started my tests with the numbers switched (k_split=2/n_split=5) yesterday afternoon and the forecasts are **still** pending to run on Hera. It's excruciating.

If my Orion access gets restored soon I will switch my testing over to there, the current Hera situation is completely untenable for timely development.

ush/templates/input.nml.FV3
mkavulich: @JeffBeck-NOAA Just FYI, the GST case actually uses a different physics suite which results in different k_split and n_split values for that test. The slower test walltimes are likely due to some weather model updates; I'll put more details in #490 because it appears unrelated to this PR.

mkavulich: @JeffBeck-NOAA My Hera jobs finally completed over the weekend. I can't confirm the timing tests that you ran earlier, but my tests did show a significant speedup in the tests run with the previous (incorrect) settings vs the new, correct settings of k_split=2 and n_split=5 (~11s vs ~9s). I was suspicious that somehow I was not using 32-bit executables since my 11s timesteps with the old version is longer than the 6s timestep you reported from your 32-bit test, but I confirmed the 32-bit build in the cmake logs. So I don't know what's up with that, maybe my domain is different than the one you tested originally?

JeffBeck-NOAA: @mkavulich, I assume you ported over the same domain settings that I had in my original PR, so I doubt it's the domain.  It could be the version of the model that is slowing things down for you, as I built an older version to test the domain.  If you're getting better timing with k_split=5 and n_split=2, then let's just go with that for now in this PR.  Ultimately, the point of the PR was just to get the domain added, and we can work on timing tweaks in a subsequent PR. Does that sound good?

JamesAbeles-NOAA: I can't see how k_split=5,n_split=2 can be faster than k_split=2,n_split=5.  The is more expensive computation done using the former (more calls done with k_split=5)

mkavulich: @JeffBeck-NOAA @JamesAbeles-NOAA Sorry, I just read my initial message and I agree it was super-unclear.

The new test (k_split=2 and n_split=5) is the faster one, approximately 20% faster than (k_split=5 and n_split=2). So it sounds like this is good and expected.

tests/baseline_configs/config.grid_RRFS_NA_3km.sh
mkavulich: Per conversation with @JeffBeck-NOAA we simply removed these comments from the file.

ush/templates/input.nml.FV3
JeffBeck-NOAA: @mkavulich, Ah!  That makes much more sense.  I also couldn't understand how more vertical remappings was resulting in a faster run.  In that case, are we ready for a merge with those new settings?

mkavulich: I pinged Gerard to make sure I had addressed all of his concerns before merging. I'm also hoping to finish up some follow-up tests with the old settings so I can be sure I have the right figures in the commit message, but if that doesn't finish before Gerard gets back to me I'll just merge it anyway, this PR needs to get in.

gsketefian: My original concern was that with the new k_split/n_split values, although we're improving the computational efficiency (and maybe also the forecast quality?) for one grid (e.g. CONUS 3km), we might be degrading the forecast quality for another and not know it -- maybe the Alaska grids where the vertical mapping may have to be done more frequently because of the steep terrain.  This is probably beyond the scope of this PR, but we should keep it in mind.  Maybe the bigger question is whether k_split and n_split should be the same for all grids or should they be grid-dependent.

tests/baseline_configs/config.grid_RRFS_NA_3km.sh
gsketefian: Well if the settings are not platform-specific (which is surprising) but only grid-specific, we can put them in set_predef_grid_params.sh.  Will this test run on platforms other than Hera?

JeffBeck-NOAA: There's likely some variability between platforms, but as a first approximation, chgres_cube needs more nodes and less cores for memory management for this domain across all platforms.  Since there are no NNODES_* or PPN_* settings currently defined in the set_predef_grid_params.sh script, I left the changes in the WE2E test.  The test is designed for Hera, but could be run on any platform.

mkavulich: @gsketefian I was able to run this test on Cheyenne as well (but the forecast step did not complete due to wallclock time, so I can't be sure if it would have fully run successfully). 

I agree that k_split and n_split should likely be handled by domain, however, currently they are *also* set according to the physics scheme, which complicates things. I believe they should *also* be configurable via config.sh for maximum flexibility. But again, that complicates things.

I can take this change out of the current PR...more discussion definitely needs to happen, the question is do we keep this change for now or not.

JeffBeck-NOAA: @mkavulich, if the NA 3-km WE2E test runs with the old k/n_split settings, then lets stick with them and deal with domain-, physics-, and platform-dependent settings in a later PR.

gsketefian: Sounds good to me.

JeffBeck-NOAA (approved):

gsketefian (approved):

tests/baseline_configs/config.grid_RRFS_NA_3km.sh
JamesAbeles-NOAA: On AWS for the RRFS work, we run chgres on a single node that has 768 GB of memory. I wonder if there is a way to specify large memory nodes if available, but will leave that to you

Issue #497, opened 2021-05-19
Add new FIX files for recent update on Thompson scheme


chan-hoo:
- A recent update on Thompson scheme requires additional FIX files (`qr_acr_qgV2.dat` and `qr_arc_qsV2.dat`) that do not exist in the `FIXgsm` directory on each machine.
- The two files should be added to the list of `thompson_mp_fix_files` in `ush/set_thompson_mp_fix_files.sh`.

PR #498, opened 2021-05-19
Add new FIX files for recent update on Thompson scheme

chan-hoo:
## DESCRIPTION OF CHANGES: 
- A recent update on Thompson scheme requires additional FIX files: `qr_acr_qgV2.dat` and `qr_arc_qsV2.dat`.
- The two files are added to the list of `thompson_mp_fix_files` in `ush/set_thompson_mp_fix_files.sh`.
- The two files have been added to the the `FIXgsm` directory on Hera, WCOSS dell and cray, Orion, Jet, and RZDM.

## TESTS CONDUCTED: 
- It was confirmed that the new two files were copied to the `fix_am` directory by running `generate_FV3LAM_wflow.sh`.

## ISSUE: 
- Fixes issue mentioned in #497 

## CONTRIBUTORS: 
@JiliDong-NOAA, @KateFriedman-NOAA

### DIFF
--- a/ush/set_thompson_mp_fix_files.sh
+++ b/ush/set_thompson_mp_fix_files.sh
@@ -130,6 +130,8 @@ string."
       "freezeH2O.dat" \
       "qr_acr_qg.dat" \
       "qr_acr_qs.dat" \
+      "qr_acr_qgV2.dat" \
+      "qr_acr_qsV2.dat" \
       )
 
     if [ "${EXTRN_MDL_NAME_ICS}" != "HRRR" -a "${EXTRN_MDL_NAME_ICS}" != "RAP" ] || \


mkavulich:
@chan-hoo just for clarification, are these new files needed in *addition* to the original `qr_acr_qg.dat` and `qr_acr_qs.dat`, or are they replacing them in future versions of the Thompson scheme?

JiliDong-NOAA:
> @chan-hoo just for clarification, are these new files needed in _addition_ to the original `qr_acr_qg.dat` and `qr_acr_qs.dat`, or are they replacing them in future versions of the Thompson scheme?

For the recent update of Thompson scheme, I think these new files replace the original `qr_acr_qg.dat` and `qr_acr_qs.dat`. The question is if we want to keep the original files in the workflow for backward compatibility?

chan-hoo:
@mkavulich, @JiliDong-NOAA, since these new files are for the recent update, someone might use the old version. So, in my opinion, it would be better to keep both versions (all four files) at this moment.

JiliDong-NOAA:
> @mkavulich, @JiliDong-NOAA, since these new files are for the recent update, someone might use the old version. So, in my opinion, it would be better to keep both versions (all four files) at this moment.

agree with @chan-hoo

JeffBeck-NOAA:
Ideally, it would be great to add an "if" statement to only copy the appropriate fixed files based on the version of Thompson MP that is run.  Aside from looking at the date associated with the ufs-weather-model hash in the Externals.cfg file (which I'm not sure is possible), maybe there's a difference in the SDF XML that can be evaluated?  In any case, I'll approve this as is for now.

JeffBeck-NOAA (approved):

PR #501, opened 2021-06-03
RRFS_baseline: Changes necessary for the RRFS_baseline tests.

JeffBeck-NOAA:
## DESCRIPTION OF CHANGES: 
Changes necessary to run the RRFS_baseline tests.

## CONTRIBUTORS
Thank you @jwolff-ncar and @siwei-noaa for additional required modifications.


ush/config_summer.sh
```diff
@@ -0,0 +1,66 @@
...
+EXTRN_MDL_NAME_ICS="FV3GFS"
+EXTRN_MDL_NAME_LBCS="FV3GFS"
+
+FV3GFS_FILE_FMT_ICS="nemsio"
+FV3GFS_FILE_FMT_LBCS="grib2"
+
+WTIME_RUN_FCST="08:00:00"
+
+MODEL="FV3_RRFS_3km_summer"
+METPLUS_PATH="/contrib/METplus/METplus-3.0"
```
jwolff-ncar: Update to be: /contrib/METplus/METplus-4.0.0

ush/config_winter.sh
```diff
@@ -0,0 +1,66 @@
...
+EXTRN_MDL_NAME_ICS="FV3GFS"
+EXTRN_MDL_NAME_LBCS="FV3GFS"
+
+FV3GFS_FILE_FMT_ICS="nemsio"
+FV3GFS_FILE_FMT_LBCS="grib2"
+
+WTIME_RUN_FCST="08:00:00"
+
+MODEL="FV3_RRFS_3km_winter"
+METPLUS_PATH="/contrib/METplus/METplus-3.0"
```
jwolff-ncar: Update to be: /contrib/METplus/METplus-4.0.0

ush/config_summer.sh
```diff
@@ -0,0 +1,66 @@
...
+
+EXTRN_MDL_NAME_ICS="FV3GFS"
+EXTRN_MDL_NAME_LBCS="FV3GFS"
+
+FV3GFS_FILE_FMT_ICS="nemsio"
+FV3GFS_FILE_FMT_LBCS="grib2"
+
+WTIME_RUN_FCST="08:00:00"
+
+MODEL="FV3_RRFS_3km_summer"
```
jwolff-ncar: Let's make this FV3_RRFS_v1alpha_3km_summer

ush/config_winter.sh
```diff
@@ -0,0 +1,66 @@
...
+
+EXTRN_MDL_NAME_ICS="FV3GFS"
+EXTRN_MDL_NAME_LBCS="FV3GFS"
+
+FV3GFS_FILE_FMT_ICS="nemsio"
+FV3GFS_FILE_FMT_LBCS="grib2"
+
+WTIME_RUN_FCST="08:00:00"
+
+MODEL="FV3_RRFS_3km_winter"
```
jwolff-ncar: Let's make this FV3_RRFS_v1alpha_3km_winter

ush/config_summer.sh
```diff
@@ -0,0 +1,66 @@
...
+METPLUS_PATH="/contrib/METplus/METplus-3.0"
+MET_INSTALL_DIR="/contrib/met/9.0.2"
```
jwolff-ncar, line 31: Since the last testing I did MET 10.0 has also been released so we could update that. If we do that, we need to also update the run_vx.local file accordingly and the MET config files to also use that version. We can do that but we will want to do a careful check of the output. I can suggest the other files that need changing if others want to see this change. Let me know.

ush/config_summer.sh
JeffBeck-NOAA: Updated in latest commit.

ush/config_winter.sh
JeffBeck-NOAA: Updated in latest commit.

ush/config_summer.sh
JeffBeck-NOAA: Keeping things as is after our discussion.

ush/config_winter.sh
JeffBeck-NOAA: Updated in latest commit.

ush/config_summer.sh
JeffBeck-NOAA: Updated in latest commit.

jwolff-ncar (approved):

Issue #503, opened 2021-06-04
Parameterize name of forecast model in setup.sh


chan-hoo:
## Description
The name of the forecast model as the base directory of the external source code is fixed to 'ufs_weather_model' in 'ush/setup.sh'. To apply the regional workflow to other coupled models, its name needs to be parameterized in the configuration files. For example, RRFS-CMAQ uses 'fv3gfs_aqm' for the forecast model.

PR #504, opened 2021-06-04
Parameterize name of forecast model in ush/setup.sh

chan-hoo:
## DESCRIPTION OF CHANGES: 
- The name of the base directory of the external source code for the forecast model is parameterized in 'ush/setup.sh'.
- Its default value is 'ufs_weather_model' in 'ush/config_defaults.sh' and it can be specified in 'ush/config.sh'.
- Update the version of Rocoto for the WCOSS dell and cary.

## TESTS CONDUCTED: 
WE2E Tests on Hera:
- grid_RRFS_NA_13km
- grid_RRFS_NA_3km

## ISSUE: 
Fixes issue mentioned in #503

### DIFF
--- a/ush/config_defaults.sh
+++ b/ush/config_defaults.sh
@@ -295,6 +295,9 @@ DOT_OR_USCORE="_"
 # the directory in which it is created in the build step to the executables
 # directory (EXECDIR; this is set during experiment generation).
 #
+# FCST_MODEL:
+# Name of forecast model (default=ufs_weather_model)
+#
 # WFLOW_XML_FN:
 # Name of the rocoto workflow XML file that the experiment generation
 # script creates and that defines the workflow for the experiment.
@@ -345,6 +348,8 @@ FV3_NML_BASE_ENS_FN="input.nml.base_ens"
 MODEL_CONFIG_FN="model_configure"
 NEMS_CONFIG_FN="nems.configure"
 FV3_EXEC_FN="ufs_model"
+
+FCST_MODEL="ufs_weather_model"
 
 WFLOW_XML_FN="FV3LAM_wflow.xml"
 GLOBAL_VAR_DEFNS_FN="var_defns.sh"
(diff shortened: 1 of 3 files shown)


JeffBeck-NOAA:
@gsketefian, @mkavulich, any concerns with this change?

ush/config_defaults.sh
```diff
@@ -295,6 +295,9 @@ DOT_OR_USCORE="_"
 # the directory in which it is created in the build step to the executables
 # directory (EXECDIR; this is set during experiment generation).
 #
+# FCST_MODEL:
```
gsketefian, line 298: Under what circumstances would you change this to something other than the default?  If using something other than the default, it means you're adding another section for a new external repository in Externals.cfig (in the ufs-srweather-app repo).  Is that the intent?

Also, while you're at it, can you change the line
`local_path = src/ufs_weather_model
`
in Externals.cfg to 
`local_path = src/ufs-weather-model
`
(i.e. use dashes instead of underscores) so that the subdirectory used matches the name of the repo -- unless someone else has an objection?  Thanks.

JeffBeck-NOAA: @gsketefian, from @chan-hoo's related issue #503: 

"The name of the forecast model as the base directory of the external source code is fixed to 'ufs_weather_model' in 'ush/setup.sh'. To apply the regional workflow to other coupled models, its name needs to be parameterized in the configuration files. For example, RRFS-CMAQ uses 'fv3gfs_aqm' for the forecast model."

I'd agree about changing the path for the weather model to make it the same as the repo.

gsketefian: Ok, sounds good to me.

gsketefian (approved):

ush/config_defaults.sh
JeffBeck-NOAA: Approving as well.

JeffBeck-NOAA (approved):

Issue #512, opened 2021-06-14
Launch script does not work correctly for cycled runs


chan-hoo:
# Description
The launch script 'ush/launch_FV3LAM_wflow.sh' does not work correctly for cycled runs. An experiment with multiple cycles stops when the first cycle is complete with 'SUCCESS'.

## Output

![error_launch](https://user-images.githubusercontent.com/60152248/121873915-deeae080-ccd4-11eb-8a6c-1488a0aab656.png)

PR #513, opened 2021-06-14
Fix the array index issue on cycle_staus in the launch script

chan-hoo:
## DESCRIPTION OF CHANGES: 
- An array index of 'cycle_statue' was missing in the conditional statement of the launch script (launch_FV3LAM_wflow.sh).
- An array index is added to 'cycle_status'.

## TESTS CONDUCTED: 

fix_launch

## ISSUE: 
Fixes issue mentioned in #512


JulieSchramm (approved):
Good catch.

Issue #514, opened 2021-06-15
Fix HPSS path for GFSv16 grib2 data (requires addition of new /atmos directory within tar files)


JeffBeck-NOAA:
# Description

Attempts to retrieve recent GFSv16 grib2 files from HPSS will fail.  Need to fix HPSS path (requires addition of new /atmos directory within tar files)

## Steps to Reproduce

Pull GFSv16 grib2 files from HPSS.

JeffBeck-NOAA:
Related PR was merged, closing this issue.

PR #515, opened 2021-06-15
Add /atmos to HPSS path for GFS grib2 files after certain date

JeffBeck-NOAA:
## DESCRIPTION OF CHANGES: 
Add "atmos" to the HPSS path for FV3GFS data after March 20th, 2021.

## TESTS CONDUCTED: 
Tests one day before an after the change on Hera.

## ISSUE (optional): 
This resolves Issue #514.

### DIFF
--- a/ush/get_extrn_mdl_file_dir_info.sh
+++ b/ush/get_extrn_mdl_file_dir_info.sh
@@ -920,12 +920,12 @@ workflow to look for the external model files in a user-staged directory."
 
     arcv_fmt="tar"
     
-    if [ "${cdate_FV3LAM}" -lt "2021032100" ]; then
-    arcvrel_dir="./gfs.${yyyymmdd}/${hh}"
-    else
-    arcvrel_dir="./gfs.${yyyymmdd}/${hh}/atmos"
+    slash_atmos_or_null=""
+    if [ "${cdate_FV3LAM}" -ge "2021032100" ]; then
+      slash_atmos_or_null="/atmos"
     fi
-
+    arcvrel_dir="./gfs.${yyyymmdd}/${hh}${slash_atmos_or_null}"
+     
     is_array arcv_fns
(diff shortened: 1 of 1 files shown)


ush/get_extrn_mdl_file_dir_info.sh
```diff
@@ -919,7 +919,12 @@ workflow to look for the external model files in a user-staged directory."
     fi
 
     arcv_fmt="tar"
+    
+    if [ "${cdate_FV3LAM}" -lt "2021032100" ]; then
```
gsketefian: To avoid having the string "./gfs.${yyyymmdd}/${hh}" appearing in two places, I'd rearrange this to:
```
    slash_atmos_or_null=""
    if [ "${cdate_FV3LAM}" -ge "2021032100" ]; then
      slash_atmos_or_null="/atmos"
    fi
    arcvrel_dir="./gfs.${yyyymmdd}/${hh}/${slash_atmos_or_null}"
```
That way in future PRs, any changes to arcvrel_dir need to be made only in one place.

gsketefian (approved):

ush/get_extrn_mdl_file_dir_info.sh
JeffBeck-NOAA: @gsketefian, should this be "atmos_or_null" without the slash?  The slash is already in the arcvrel_dir path, so just checking, otherwise if we keep the slash, then it's:

    arcvrel_dir="./gfs.${yyyymmdd}/${hh}${slash_atmos_or_null}"

With no slash.  Right?

JeffBeck-NOAA: I'll keep it slash_atmos_or_null and remove the slash from arcvrel_dir.

gsketefian: Yes, you're right, that was a bug on my part.

Issue #516, opened 2021-06-16
restart_interval needs to be selectable in config.sh for warm start runs


chan-hoo:
## Description
The parameter 'restart_interval' for 'model_configure' is not selectable in the configuration. It is fixed to zero in the template of 'model_configure'. For a warm start, this parameter needs to be selectable.

## Solution
Introduce a new parameter 'RESTART_INTERVAL' in the configuration files.

PR #517, opened 2021-06-16
Make restart_interval selectable in config.sh for warm start

chan-hoo:
## DESCRIPTION OF CHANGES: 
Make the parameter 'restart_interval' in 'model_configure' selectable in 'config.sh' for warm start runs.

## TESTS CONDUCTED: 
- GST test for 12h forecast without RESTART_INTERVAL in config.sh on the wcoss dell
- GST test for 12h forecast with RESTART_INTERVAL="1" in config.sh on the wcoss dell

## ISSUE: 
Fixes issue mentioned in #516


JeffBeck-NOAA:
@chan-hoo, thanks for getting us one step closer to cycling with this change.  Does setting the restart interval to one produce restart files every hour that can then be used to initialize the next cycle (based on any of those hourly restart files?)

chan-hoo:
@JeffBeck-NOAA, yes, it does

JeffBeck-NOAA (approved):

ush/config_defaults.sh
```diff
@@ -380,12 +380,16 @@ WFLOW_LAUNCH_LOG_FN="log.launch_FV3LAM_wflow"
 # FCST_LEN_HRS:
 # The length of each forecast, in integer hours.
 #
+# RESTART_INTERVAL:
+# Parameter in model_configure; frequency of the output restart files
```
gsketefian: Can you describe here the units of RESTART_INTERVAL and what it means if it is set to 0?  Thx.

Is it true that restart files are always written at the end of a forecast regardless of whether or not restart_interval is set in the model_configure file?  If so, can you mention that here as well?

ush/setup.sh
```diff
@@ -636,6 +636,18 @@ check_var_valid_value \
 #
 #-----------------------------------------------------------------------
 #
+# Make sure RESTART_INTERVAL is set to an integer value if present
+#
+#-----------------------------------------------------------------------
+#
+if ! [[ "${RESTART_INTERVAL:-0}" =~ ^[0-9]+$ ]]; then
```
gsketefian: Why is a default value also given as part of the test?  Maybe remove the :-0 part since we want to make sure RESTART_INTERVAL is defined at this point in the script.

gsketefian:
@chan-hoo Can you also add a WE2E test that sets RESTART_INTERVAL to a non-default value?  Just a short (say 3-hr) forecast on the 25km grid should be ok.  Thx.

ush/config_defaults.sh
chan-hoo: @gsketefian, only when restart_interval=0, the restart files are written at the end. In the case of restart_interval=1, the restart files are written every hour in the RESTART directory.

ush/setup.sh
chan-hoo: I'll remove it

ush/config_defaults.sh
gsketefian: Ok, thanks.  Can you add your reply to the description of RESTART_INTERVAL on line 384?

chan-hoo: done.

ush/setup.sh
chan-hoo: done.

chan-hoo:
@gsketefian, I've modified the scripts and added a WE2E test for restart_interval.

ush/create_model_configure_file.sh
```diff
@@ -121,6 +121,7 @@ run directory (run_dir):
   'dt_atmos': ${DT_ATMOS}
   'atmos_nthreads': ${nthreads:-1}
   'ncores_per_node': ${NCORES_PER_NODE}
+  'restart_interval': ${RESTART_INTERVAL:-0}
   'quilting': ${dot_quilting_dot}
```
JeffBeck-NOAA, line 125: Is this :-0 necessary?  Should it be just ${RESTART_INTERVAL} since it's getting it either from config.sh or config_defaults.sh?

gsketefian (approved):
Looks good, thanks for addressing my comments.  Feel free to merge whenever you've addressed @JeffBeck-NOAA 's comments.

ush/create_model_configure_file.sh
chan-hoo: removed.

Issue #522, opened 2021-06-21
The parameter CPL in model_configure needs to be selectable for use of coupled models.


chan-hoo:
## Description
- The parameter 'cpl' in the template 'model_configure' is fixed to 'false'.
- This parameter needs to be selectable between true/false to use a coupled model as a forecast model (for example, fv3gfs-aqm).

## Solution
- Make the parameter selectable in the configuration file 'config.sh'.

PR #523, opened 2021-06-21
Make the coupling parameter CPL in model_configure selectable in config.sh

chan-hoo:
## DESCRIPTION OF CHANGES: 
- To use a coupled model as a forecast model, the coupling parameter 'cpl' in model_configure should be switched between false/true in ush/setup.sh.
- For example, FV3-LAM: cpl=.false. , RRFS-CMAQ: cpl=.true.

## TESTS CONDUCTED: 
- GST test for 12h forecast on the wcoss dell

## ISSUE: 
Fixes issue mentioned in #522


gsketefian:
To reduce changes of inconsistent settings, would it be better to set cpl automatically depending on the model being used (FCST_MODEL), e.g. in setup.sh?  That would remove a manual step that the user has to remember to do.

chan-hoo:
@gsketefian, I've modified and tested it successfully.

ush/config_defaults.sh
```diff
@@ -380,19 +380,37 @@ WFLOW_LAUNCH_LOG_FN="log.launch_FV3LAM_wflow"
...
 # with the prefix "YYYYMMDD.HHmmSS." in the RESTART directory
 #
 #-----------------------------------------------------------------------
 #
-DATE_FIRST_CYCL="YYYYMMDD"
-DATE_LAST_CYCL="YYYYMMDD"
-CYCL_HRS=( "HH1" "HH2" )
-FCST_LEN_HRS="24"
+DT_ATMOS=""
+CPL=""
```
gsketefian: Since CPL is now automatically set in setup.sh, it is no longer a user-configurable variable but what I like to call a derived or secondary workflow variable.  For that reason, you can remove it from here (since config_defaults.sh should contain only user-configurable variables) and have it be written to var_defns.sh towards the end of setup.sh like the other derived variables (I'll put a note around where I think it can be added in that file).

ush/setup.sh
```diff
@@ -636,6 +636,20 @@ check_var_valid_value \
...
 #
+# Set CPL to TRUE/FALSE based on FCST_MODEL.
+#
+#-----------------------------------------------------------------------
+#
+if [ "${FCST_MODEL}" = "ufs-weather-model" ]; then
+  CPL="FALSE"
+elif [ "${FCST_MODEL}" = "fv3gfs_aqm" ]; then
+  CPL="TRUE"
+else
```
gsketefian, line 655: For the "else" case here, it might be better to write out a message saying CPL has not been specified for the given FCST_MODEL.  

Also, it would be good to define valid_vals_FCST_MODEL in valid_param_vals.sh and check that FCST_MODEL (in setup.sh but somewhere before FCST_MODEL is used) to make sure it is set to a valid value:
`check_var_valid_value "FCST_MODEL" "valid_vals_FCST_MODEL"
`

ush/setup.sh
```diff
@@ -636,6 +636,20 @@ check_var_valid_value \
...
+  CPL="FALSE"
+fi
+#
+#-----------------------------------------------------------------------
+#
 # Make sure RESTART_INTERVAL is set to an integer value if present
 #
 #-----------------------------------------------------------------------
```
gsketefian, line 665: Ok, if you end up removing CPL from config_defaults.sh, you can add a statement like this:
`CPL="$CPL"
`
somewhere between lines 2546 and 2843 in setup.sh.

ush/config_defaults.sh
chan-hoo: Removed.

ush/setup.sh
chan-hoo: Done.

chan-hoo:
@gsketefian, I've removed cpl from config_default.sh, modified the error message for else, and added the 'cpl' statement to the end of setup.sh.

ush/setup.sh
chan-hoo: Modified.

ush/setup.sh
```diff
@@ -636,6 +636,32 @@ check_var_valid_value \
...
+# Set CPL to TRUE/FALSE based on FCST_MODEL.
+#
+#-----------------------------------------------------------------------
+#
+if [ "${FCST_MODEL}" = "ufs-weather-model" ]; then
+  CPL="FALSE"
+elif [ "${FCST_MODEL}" = "fv3gfs_aqm" ]; then
+  CPL="TRUE"
+else
+  err_msg="\
```
gsketefian: I'd replace the contents of the else-clause with just this:
```
  print_err_msg_exit "\
The coupling flag CPL has not been specified for this value of FCST_MODEL:
  FCST_MODEL = \"${FCST_MODEL}\""
```
You don't need to call `check_var_valid_value` again since you already do that in line 643 above (so you're already sure that whatever value FCST_MODEL has is a valid value).

gsketefian (approved):
Just had one comment on that else-clause, otherwise looks good.  Thanks!

ush/setup.sh
chan-hoo: Done.

PR #525, opened 2021-06-24
RRFS_baseline: Update version of MET in config template files

JeffBeck-NOAA:
## DESCRIPTION OF CHANGES: 
Update the version of MET to 10.0.0 in both config template files.


gsketefian (approved):

gsketefian:
I was surprised that the tasks were still reported as successful in rocotostat even though they failed.  The scripts may need some tweaking to so they return a non-zero value when they fail.  Maybe a separate PR for that...

JeffBeck-NOAA:
> I was surprised that the tasks were still reported as successful in rocotostat even though they failed. The scripts may need some tweaking to so they return a non-zero value when they fail. Maybe a separate PR for that...

Yes, this is a known problem with the verification tasks due to difficulties in determining what is considered a failure.  For example, error messages are thrown when observations aren't available for a given hour; however, that shouldn't be considered as a task failure, since it's somewhat common for obs to be unavailable at certain times.

PR #533, opened 2021-07-01
Changes required to produce the diag_table file at forecast run time.

JeffBeck-NOAA:
## DESCRIPTION OF CHANGES: 
This PR creates the diag_table file during the forecast task instead of when the grid generation script is run.  It contains the same changes as PR#349, but includes the compromise solution for ALL_CDATES based on the size of NUM_CYCLES.

## TESTS CONDUCTED: 
Tested on Hera with single and multiple cycles to ensure proper creation of diag_table files and ALL_CDATES information in var_defns.sh as a function of NUM_CYCLES.

### DIFF
--- a/ush/create_diag_table_file.sh
+++ b/ush/create_diag_table_file.sh
@@ -78,7 +78,7 @@ function create_diag_table_file() {
 #
   print_info_msg "$VERBOSE" "                                            
 Creating a diagnostics table file (\"${DIAG_TABLE_FN}\") in the specified
-run directory..."
+run directory...
 
   run_dir = \"${run_dir}\""
 
@@ -94,7 +94,7 @@ to create:
     diag_table_fp = \"${diag_table_fp}\""
 
     settings="
-    starttime: !datetime ${ALL_CDATES[$i]}
+    starttime: !datetime ${CDATE}
     cres: ${CRES}"
 
     $USHDIR/fill_jinja_template.py -q -u "${settings}" -t "${DIAG_TABLE_TMPL_FP}" -o "${diag_table_fp}" || \
--- a/ush/create_diag_table_file.sh
+++ b/ush/create_diag_table_file.sh
@@ -94,8 +94,8 @@ to create:
     diag_table_fp = \"${diag_table_fp}\""
 
     settings="
-    starttime: !datetime ${CDATE}
-    cres: ${CRES}"
+starttime: !datetime ${CDATE}
+cres: ${CRES}"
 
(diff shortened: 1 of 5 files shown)


ush/create_diag_table_file.sh
```diff
@@ -96,9 +94,8 @@ to create:
     diag_table_fp = \"${diag_table_fp}\""
 
     settings="
-  starttime: !datetime ${ALL_CDATES[$i]}
-  cres: ${CRES}
-"
+    starttime: !datetime ${CDATE}
```
gsketefian: Unindent lines 97-98 completely (4 spaces).

ush/setup.sh
```diff
@@ -730,6 +730,13 @@ set_cycle_dates \
   output_varname_all_cdates="ALL_CDATES"
 
 NUM_CYCLES="${#ALL_CDATES[@]}"
+
+if [ $NUM_CYCLES -gt 30 ] ; then
+unset ALL_CDATES
```
gsketefian: Indent lines 735-736 two spaces.

ush/setup.sh
```diff
@@ -730,6 +730,13 @@ set_cycle_dates \
...
+print_info_msg "$VERBOSE" "
+Too many cycles in ALL_CDATES, defining in abbreviated form."
```
gsketefian: This could be interpreted as an error message, so maybe say something like:
Too many cycles in ALL_CDATES to list, redefining ALL_CDATES in abbreviated form."

gsketefian (approved):

ush/setup.sh
```diff
@@ -730,6 +730,13 @@ set_cycle_dates \
   output_varname_all_cdates="ALL_CDATES"
 
 NUM_CYCLES="${#ALL_CDATES[@]}"
+
+if [ $NUM_CYCLES -gt 30 ] ; then
```
gsketefian, line 734: Did you try a case with more than 30 cycles to see if this outputs the expected abbreviated form in var_defns.sh?

JeffBeck-NOAA: Yes, on Hera here:

/scratch2/BMC/fv3lam/beck/FV3-LAM/expt_dirs/test_diag_table_long

Which produces:

ALL_CDATES=( \
"2019061500...2019072000" \
)

ush/setup.sh
JeffBeck-NOAA: Done.

ush/create_diag_table_file.sh
JeffBeck-NOAA: Done.

ush/setup.sh
JeffBeck-NOAA: Done.

Issue #534, opened 2021-07-02
Update met version and workflow template for WCOSS


chan-hoo:
## Description
- Since MET was updated to 10.0 on Hera, its version needs to be updated for WCOSS.
- Since some verification tasks in the workflow template were modified from hpss queue to default queue, some commands for the hpss queue are not necessary on WCOSS.

## Solution
- Update the version of MET to 10.0.0 for WCOSS.
- Clean up unnecessary commands in the workflow template for WCOSS.

## Related to
PR https://github.com/NOAA-EMC/regional_workflow/pull/528

PR #535, opened 2021-07-02
Update met version and workflow template for WCOSS

chan-hoo:
## DESCRIPTION OF CHANGES: 
- Since the version of MET was updated to 10.0 on Hera, its version on WCOSS are updated to 10.0.0.
- Since some verification tasks in the workflow template were modified from hpss queue to default queue, some commands only for the WCOSS dell and cray in the corresponding tasks are removed.

## TESTS CONDUCTED: 
GST + verification on the wcoss dell

## ISSUE: 
Fixes issue mentioned in #534.
Related to PR https://github.com/NOAA-EMC/regional_workflow/pull/528.


JeffBeck-NOAA (approved):
Looks good!

PR #536, opened 2021-07-02
Streamline symlink creation

gsketefian:
## DESCRIPTION OF CHANGES:
This PR (1) helps reduce code duplication and (2) removes the necessity to check for the machine the workflow is running on before attempting to create relative symlinks by introducing the new function `create_symlink_to_file` for creating symlinks to files (but not directories).  This function:
* Checks for the existence of a symlink's target before creating it.  If the target doesn't exist, it prints out an error message and exits.  If it does, it creates the symlink to the target file.  Currently, checking for the existence of a target is done in several places in the scripts.  This PR changes the code so that instead of performing this check themselves, the scripts call this new function to perform this test and, if the test succeeds, create the symlink. 
* Creates a relative symlink (i.e. a symlink whose target is specified via a relative path) only if requested via an input argument AND if the machine/OS/shell the workflow is running on supports relative symlinks.  The second half of this test (after the AND) is done implicitly via the use of the new experiment variable `RELATIVE_LINK_FLAG`.  This variable gets set during experiment generation to the flag that needs to be passed to the link command on the machine in order to create relative links.  For example, on Hera, `RELATIVE_LINK_FLAG` gets set to "--relative".  On machines that don't support creation of relative links, `RELATIVE_LINK_FLAG` gets set to a null string.

## TESTS CONDUCTED:
The following WE2E tests were conducted:
```
DOT_OR_USCORE
nco_ensemble
nco_grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_GSD_SAR
pregen_grid_orog_sfc_climo
suite_FV3_GFS_v15p2
suite_FV3_GFS_v16
```
All completed successfully except suite_FV3_GFS_v16.  The latter failed in the run_fcst task after hour 2, i.e. all the symlinks, etc were created properly for the forecast to start, so this failure was not due to changes in this PR and was preexisting (confirmed this by running the test with the develop branch).


ush/valid_param_vals.sh
```diff
@@ -1,3 +1,10 @@
+#
+# Source necessary scripts.
+#
+. ./constants.sh
+#
+# Define valid values for various global experiment/workflow variables.
+#
 valid_vals_RUN_ENVIR=("nco" "community")
```
JeffBeck-NOAA, line 4: Why does this file need to source constants.sh?

ush/constants.sh
```diff
@@ -1,11 +1,17 @@
...
+#
 
 # Pi.
 pi_geom="3.14159265358979323846264338327"
+
 # Degrees per radian.
 degs_per_radian=$( bc -l <<< "360.0/(2.0*$pi_geom)" )
+
 # Radius of the Earth in meters.
 radius_Earth="6371000.0"
```
JeffBeck-NOAA, line 17: Should we use 6371200.0 m for the radius of the Earth as is the case now in UFS_UTILS and in FV3?  What in regional_workflow uses it?

ush/valid_param_vals.sh
gsketefian: It doesn't.  This was part of another feature branch that I forgot to remove when creating this branch.  Will remove.

ush/constants.sh
gsketefian: Yes, will change.

JeffBeck-NOAA (approved):

Issue #511, opened 2021-06-12
Add in-line post option to the SRW App


JeffBeck-NOAA:
## Description
Need to add an option to the SRW App to run with in-line post.

## Solution
Changes and user-defined options will be required for model_configure among other files.

## Related to (optional)
@chan-hoo already provided a flag to shut off the run_post task in [PR #510](https://github.com/NOAA-EMC/regional_workflow/pull/510).  This is necessary if a user chooses to run with in-line post activated.

JeffBeck-NOAA:
Need to modify the post build as well, open corresponding issue in ufs-srweather-app repo.

PR #531, opened 2021-06-29
Reorganize the WE2E (workflow end-to-end) testing system

gsketefian:
## DESCRIPTION OF CHANGES: 
This PR reorganizes the WE2E testing system so that it is easier to comprehend, use, and modify.  The major changes are as follows:

* Move the WE2E testing system from the `regional_workflow/tests` directory to `regional_workflow/tests/WE2E`.  This is because we anticipate other types of test in the testing system, e.g. unit tests, which would go under `regional_workflow/tests/unit`, etc.
* Move the WE2E test configuration files that were in `regional_workflow/tests/baseline_configs` to category subdirectories under `regional_workflow/tests/WE2E/test_configs`.  The category subdirectories thus far and the types of tests they contain are: 
  * `grids_extrn_mdls_suites_community`
    Tests in community mode of various combinations of the grid, external models for ICs and LBCs, and physics suites.
  * `grids_extrn_mdls_suites_nco`
    Tests in NCO mode of various combinations of the grid, external models for ICs and LBCs, and physics suites.
  * `release_SRW_v1`
    The graduate student test (GST) used for the UFS SRW App version 1 release.
  * `wflow_features`
    Test of workflow features, e.g. ability to set various parameters to user-specified values instead of using the defaults, ability to fetch external model files from different models and on different dates from NOAA-HPSS, etc.
* Rename some of the WE2E test configuration files to adhere to the naming convention used in each category subdirectory.
* Remove some of the WE2E test configuration files since they are almost-duplicates, e.g. they differ with respect to another test only in the cycle date/time used or the LBC specification interval.
* Changes to contents of test configuration files:
  * Rearrange the order in which experiment variables are specified in the test configuration files so that the predefined grid name and the physics suite are set first, then the external model info is set, then the cycle dates, then the forecast length and LBC update interval, finally followed by other parameters.
  * Remove redundant variable specifications in the test configuration files, e.g. remove `QUILTING="TRUE"` since `QUILTING` is already set to `"TRUE"` by default, remove `GRID_GEN_METHOD="ESGgrid"` since a predefined grid is already specified that in turn implies a value for `GRID_GEN_METHOD`.
  * Include a test purpose/description in each test configuration file.
  * Perform bug fixes in the WE2E configuration files, e.g. a test that was supposed to run on the `RRFS_CONUS_3km` grid was actually running on the `RRFS_CONUS_13km` grid.
* Remove the file `baselines_list.txt` since it it not used.
[...]

PR #538, opened 2021-07-06
Add inline post option to the SRW App

chan-hoo:
## DESCRIPTION OF CHANGES: 
- Add the parameter 'write_dopost' for inline post to model_configure.
- Turn off the 'run_post' task when 'write_dopost'='.true.'
- Add post-processing for inline post to the 'run_fcst' script.
- Modify the dependency for verification in the workflow template.
- Add a new WE2E test for inline post.

## TESTS CONDUCTED: 
- GST + verification on Hera and WCOSS Dell
- New WE2E test for inline post

## DEPENDENCIES:
- https://github.com/ufs-community/ufs-srweather-app/pull/150

## ISSUE: 
- Fixes issue mentioned in #511


ush/config_defaults.sh
```diff
@@ -407,10 +407,15 @@ FCST_LEN_HRS="24"
...
+# WRITE_DOPOST:
+# Flag that determines whether or not to use the INLINE POST option
+# When TRUE, force to turn off run_post (RUN_TASK_RUN_POST=FALSE) in setup.sh
+#
 #-----------------------------------------------------------------------
 #
 DT_ATMOS=""
 RESTART_INTERVAL="0"
+WRITE_DOPOST="FALSE"
 #
```
JeffBeck-NOAA, line 419: Could this variable be called "DO_INLINE_POST" instead?  I assume WRITE_DOPOST refers to running post within the write component, but it might be confusing to those not familiar with how that works.

ush/templates/FV3LAM_wflow.xml
```diff
@@ -693,7 +693,11 @@ the <task> tag to be identical to the ones above for other output times.
...
 
     <dependency>
+  {%- if write_dopost %}
+      <taskdep task="&RUN_FCST_TN;{{ uscore_ensmem_name }}"/>
+  {%- else %}  
       <metataskdep metatask="&RUN_POST_TN;{{ uscore_ensmem_name }}"/>
+  {%- endif %}
     </dependency>
 
   </task>
```
JeffBeck-NOAA, line 703: In addition to these if statements for dependencies, can you wrap the RUN_POST task in an if statement to not include it if inline post is used?

JeffBeck-NOAA (changes requested):

ush/config_defaults.sh
JeffBeck-NOAA: Actually, it looks like the new variable in the model_configure file is "write_dopost", so maybe it makes sense to define it that way everywhere.  You've described what it means here, so it's probably fine.

JeffBeck-NOAA (changes requested):
It doesn't look like WRITE_DOPOST automatically sets RUN_TASK_RUN_POST=FALSE anywhere.  Can you make this change in setup.sh?

chan-hoo:
@JeffBeck-NOAA, You can find it on Line 2139 in setup.sh

JeffBeck-NOAA:
> @JeffBeck-NOAA, You can find it on Line 2139 in setup.sh

Thanks!  Just missed it.

JeffBeck-NOAA (approved):

JeffBeck-NOAA:
@chan-hoo, can you add a header to the WE2E test you've included in this PR?  You can check out [this](https://github.com/NOAA-EMC/regional_workflow/pull/531) PR to see how we're handling that now.  Thanks!

JeffBeck-NOAA (approved):

gsketefian:
@chan-hoo I was just merging this into one of my feature branches, and I started wondering if setting WRITE_DOPOST="TRUE" works with SUB_HOURLY_POST="TRUE"?  If not, we should at least add a check, e.g. quit out of the experiment generation script if both of these are TRUE, and make an issue to fix this later on.  Thanks.

scripts/exregional_run_fcst.sh
```diff
@@ -507,6 +524,46 @@ $APRUN ${FV3_EXEC_FP} || print_err_msg_exit "\
 Call to executable to run FV3-LAM forecast returned with nonzero exit
 code."
 #
+# Only for inline post, create the directory where post-processing output
+# are stored (postprd_dir)
+if [ ${WRITE_DOPOST} = "TRUE" ]; then
+  if [ "${RUN_ENVIR}" = "nco" ]; then
+    COMOUT="${COMOUT_BASEDIR}/$RUN.$PDY/$cyc${SLASH_ENSMEM_SUBDIR}"
```
gsketefian, line 531: I think `cyc` is being used here before being set (below) since I don't see it set in JREGIONAL_RUN_FCST or FV3LAM_wflow.xml.  Maybe this was missed because the WE2E test is in community mode.

scripts/exregional_run_fcst.sh
```diff
@@ -507,6 +524,46 @@ $APRUN ${FV3_EXEC_FP} || print_err_msg_exit "\
...
+  cd_vrfy ${postprd_dir}
+
+  yyyymmdd=${cdate:0:8}
+  hh=${cdate:8:2}
+  cyc=$hh
+  tmmark="tm00"
+  fmn="00"
+
+  for fhr in $(seq -f "%02g" 0 ${FCST_LEN_HRS}); do
+    echo $fhr
```
gsketefian, line 547: Remove echo?

scripts/exregional_run_fcst.sh
chan-hoo: @gsketefian, I'll fix this in a new PR for the nco mode.

scripts/exregional_run_fcst.sh
chan-hoo: I'll fix this in a new PR for the nco mode.

chan-hoo:
@gsketefian, I'll add the check for SUB_HOURLY_POST, but I am not familiar with SUB_HOURLY_POST. Can you assign this 'SUB_HOURLY_POST' issue to others who know better than me?

PR #542, opened 2021-07-12
Update syntax for latest version of lmod.

christinaholtNOAA:
## DESCRIPTION OF CHANGES: 
This PR addresses an error that occurs at run time ```ModuleNotFoundError: No module named 'f90nml'```. A similar error may appear in any of the tasks that use a conda environment, and could potentially list `jinja2` instead.

The version of lmod on Hera was updated. The behavior of the `system` command has changed and now runs only in a subshell. As an lmod-version-independent fix, the module will no longer activate a conda environment, but will set the conda environment that should be activated through an environment variable, then the script will activate any required conda environment. 

Note: Users who do not run with Rocoto, but still use these modules, will have an extra step of activating the environment manually.

## TESTS CONDUCTED: 
I built and ran a single WE2E test successfully on Hera with these modifications. I did not test on other platforms.

### DIFF
--- a/modulefiles/tasks/hera/make_grid.local
+++ b/modulefiles/tasks/hera/make_grid.local
@@ -2,5 +2,5 @@
 module use -a /contrib/miniconda3/modulefiles
 module load miniconda3
 if [module-info mode load] {
-  system "conda activate regional_workflow"
+  puts "conda activate regional_workflow"
 }
--- a/modulefiles/tasks/hera/make_grid.local
+++ b/modulefiles/tasks/hera/make_grid.local
@@ -1,6 +1,5 @@
 #%Module
 module use -a /contrib/miniconda3/modulefiles
 module load miniconda3
-if [module-info mode load] {
-  puts "conda activate regional_workflow"
-}
+
+setenv SRW_ENV regional_workflow
--- a/modulefiles/tasks/hera/make_ics.local
+++ b/modulefiles/tasks/hera/make_ics.local
@@ -2,5 +2,5 @@
 module use -a /contrib/miniconda3/modulefiles
 module load miniconda3
 if [module-info mode load] {
-  system "conda activate regional_workflow"
+  puts "conda activate regional_workflow"
 }
--- a/modulefiles/tasks/hera/make_ics.local
+++ b/modulefiles/tasks/hera/make_ics.local
@@ -1,6 +1,5 @@
 #%Module
 module use -a /contrib/miniconda3/modulefiles
 module load miniconda3
-if [module-info mode load] {
-  puts "conda activate regional_workflow"
-}
+
+setenv SRW_ENV regional_workflow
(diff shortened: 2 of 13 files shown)


gsketefian:
@christinaholtNOAA Thanks for this update.  I haven't run the workflow for a couple of days.  Does this mean the workflow currently fails on Hera?

christinaholtNOAA:
That's right @gsketefian.

JeffBeck-NOAA:
Is the testing on Hera complete for this PR?

christinaholtNOAA:
It is not. I am still having failures on getting the appropriate conda environment to activate properly from the module file.

JeffBeck-NOAA:
> It is not. I am still having failures on getting the appropriate conda environment to activate properly from the module file.

OK, thanks for troubleshooting this!

christinaholtNOAA:
Unfortunately, the "simplest" fix to get this going on Hera immediately is to turn these module files into bash files. Sourcing the files still works when the syntax is changed appropriately. I don't like that necessarily, but it is consistent with what is done for the App in the env files.

JeffBeck-NOAA:
It is interesting that we source the env files instead of using "module load."  @mkavulich, is that due to the "export" commands?  @christinaholtNOAA, what happens when these local module files are loaded as such?

christinaholtNOAA:
@JeffBeck-NOAA When the modulefiles/tasks/hera module files are modified to be bash files and sourced, the desired output is acheived.

The problem seems to be that when the `conda activate regional_workflow` line is run in the modulefile version (what is currently in the PR), the environment variable for PATH is not properly exported in the subshell so we don't get the right version of Python.

christinaholtNOAA:
None of the lmod solutions I'd tried worked to have the necessary environment variables updated in the appropriate shell. I believe that is related to the security issue that was "fixed" with the update to the new version of lmod anyway. 

I took a different approach that seemed clean. For the modules files that need to activate a conda environment, they now set the SRW_ENV env variable that the `load_modules_run_task.sh` script uses to activate the required environment. This change should also be lmod-version-independent, so we can have consistency across platforms with different versions...and maybe won't see failures as they are updated.

I am currently testing the latest changes on Hera and seeing good results. I expect that single test to finish shortly. I have not tested the changes on Orion or Jet, but expect they would work on those platforms similarly.

christinaholtNOAA:
The test passed on Hera.

JulieSchramm:
@christinaholdNOAA Is this the error message you were getting on Hera?

  File "ufs-srweather-app-fork/regional_workflow/ush/set_namelist.py", line
73, in <module>

    import f90nml

ModuleNotFoundError: No module named 'f90nml'

/

gsketefian:
This fix (once Julie's issue is resolved) seems like a good fix to me.  Approving so you can merge when ready.

gsketefian (approved):

JeffBeck-NOAA (approved):
Agree with Gerard.  Approving.

christinaholtNOAA:
@JulieSchramm Yes, that is the error this aims to fix.

mkavulich (approved):
Ran a WE2E test on Orion for a cross-platform sanity check. Modules and conda environment loaded correctly and tasks completed successfully.

Issue #546, opened 2021-07-13
Failure of filter_topo and fix


JiliDong-NOAA:
# Description

Filter_topo failed with the following error message:

Starting filtering of orography...
 Unknown ErrorError reading input.nml
Stopped

## Steps to Reproduce

1. run make_orog task in the regional workflow
2. check make_orog.log

## Additional Context

The make_orog task will continue running without exiting even with the filter_orog failure.

* Machine Hera
* Compiler 
* Reference other issues or PRs in other repositories that this is related to, and how they are related.

## Output

Error message:

Starting filtering of orography...
 Unknown ErrorError reading input.nml
Stopped

## Solution:

Remove "refine_ratio = ${refine_ratio}" in scripts/exregional_make_orog.sh

In regional workflow scripts/exregional_make_orog.sh, the input.nml for filter_topo is specified as follows:

cat > "${filter_dir}/input.nml" <<EOF
&filter_topo_nml
  grid_file = "${mosaic_fp}"
  topo_file = "${filtered_orog_fp_prefix}"
  mask_field = "land_frac"
  regional = .true.
  stretch_fac = ${STRETCH_FAC}
  **refine_ratio = ${refine_ratio}**
  res = $res
/
EOF

The line above highlighted (refine_ratio = ${refine_ratio}) is an extra line which causes filter_topo failure when reading the namelist. For reference, the input.nml specified in UFS_UTILS driver script is:

cat > input.nml <<EOF
&filter_topo_nml
  grid_file = "$mosaic_grid"
  topo_file = "$topo_file"
  mask_field = "land_frac"
  regional = $regional
  stretch_fac = $stretch
  res = $res
  /
EOF
 
refine_ratio is not a valid namelist entry for filter_topo. See UFS_UTILS/sorc/grid_tools.fd/filter_topo.fd/filter_topo.F90

  _namelist /filter_topo_nml/ topo_file, topo_field, mask_field, grid_file, zero_ocean, &
       zs_filter, stretch_fac, res, nested, grid_type, regional_

The make_orog task continues without exiting even with this filter_topo failure and reports success in rocoto workflow. In summary, the topography generated using regional workflow is not filtered/smoothed (orog_raw and orog_filt are the same).

A PR will be submitted later to fix the issue.

PR #547, opened 2021-07-13
change input.nml for filter topo to fix task failure

JiliDong-NOAA:
## DESCRIPTION OF CHANGES: 
Remove "refine_ratio = ${refine_ratio}" in scripts/exregional_make_orog.sh to fix filter_topo failure

## TESTS CONDUCTED: 

regional ESG grids/topography

## ISSUE (optional): 
Fixes issue mentioned in #546

### DIFF
--- a/scripts/exregional_make_orog.sh
+++ b/scripts/exregional_make_orog.sh
@@ -493,7 +493,6 @@ cat > "${filter_dir}/input.nml" <<EOF
   mask_field = "land_frac"
   regional = .true.
   stretch_fac = ${STRETCH_FAC}
-  refine_ratio = ${refine_ratio}
   res = $res
 /
 EOF


JeffBeck-NOAA:
@JiliDong-NOAA, thank you for finding this sneaky bug!

JeffBeck-NOAA (approved):
I'm wondering if at some point, refine_ratio was used in the filter_topo code but was removed.  In any case, given Jili's successful tests after its removal, I'm approving.

JiliDong-NOAA:
> I'm wondering if at some point, refine_ratio was used in the filter_topo code but was removed. 

That is my guess too.

gsketefian:
@JiliDong-NOAA Couple of questions:
1) I also suspect that refine_ratio was used before in filter_topo but got removed at some point from the namelist.  Do you know if filter_topo now works only for ESG grids, or does it still also work for GFDL-type grids (i.e. cubed-sphere-based, like the global grid)?  It still is taking in stretch_fac as a namelist variable, so I suspect it still works for GFDL-type grids, but why then was refine_ratio removed?  Maybe @GeorgeGayno-NOAA knows the answer.
2) In issue #546, you state that:
> The make_orog task continues without exiting even with this filter_topo failure and reports success in rocoto workflow.

There is a check in the exregional_make_orog.sh to catch a failure of filter_topo:
```
$APRUN "${exec_fp}" || \
  print_err_msg_exit "\
Call to executable that generates filtered orography file returned with
non-zero exit code."
```
So if this is not catching an error, it means filter_topo is returning with a 0 exit code (success) even when the namelist error occurs.  I vaguely remember encountering this kind of behavior in some UFS_UTILS codes (not sure if it was filter_topo).  Do you mind making an issue in the UFS_UTILS repo to fix this?  Thanks.  @GeorgeGayno-NOAA

GeorgeGayno-NOAA:
> I'm wondering if at some point, refine_ratio was used in the filter_topo code but was removed. In any case, given Jili's successful tests after its removal, I'm approving.

Yes, "refine_ratio" was used to compute the grid resolution for a regional grid. Here is an old tag: https://github.com/NOAA-EMC/UFS_UTILS/blob/2d2c3387b8fb36159c14ecd6145e178564a41db8/sorc/fre-nctools.fd/tools/filter_topo/filter_topo.F90#L1789

This computation was removed when Gerard's "global_equiv_resol' code was added to UFS_UTILS.

gsketefian:
I am going to merge this as a temporary fix, but there are problems that need to be addressed, probably very soon since filter_topo can easily be returning bad results.  I made new issues in the UFS_UTILS repo for these problems:  #[547](https://github.com/NOAA-EMC/UFS_UTILS/issues/547), #[548](https://github.com/NOAA-EMC/UFS_UTILS/issues/548).

JeffBeck-NOAA:
> I am going to merge this as a temporary fix, but there are problems that need to be addressed, probably very soon since filter_topo can easily be returning bad results. I made new issues in the UFS_UTILS repo for these problems: #[547](https://github.com/NOAA-EMC/UFS_UTILS/issues/547), #[548](https://github.com/NOAA-EMC/UFS_UTILS/issues/548).

Thank you, @gsketefian!

PR #553, opened 2021-07-22
Remove symbolic linking of diag_table when running in ensemble mode.

JeffBeck-NOAA:
## DESCRIPTION OF CHANGES: 
PR #349 (via PR #533) moves diag_table creation from the grid generation step to the run_fcst portion of the workflow.  However, it also uses the "run_dir" location instead of "cycle_dir" as the target for the diag_table file.  While these two locations are identical for deterministic simulations, they are not the same when using the ensemble mode (due to member directories).  Therefore, the diag_table file should no longer be symbolically linked from the "cycle_dir" location to the "run_dir" directory in ensemble mode (the file is now copied directly to "run_dir" for each ensemble member).

## TESTS CONDUCTED: 
Successfully ran a deterministic and two-member ensemble simulation on Hera.

### DIFF
--- a/scripts/exregional_run_fcst.sh
+++ b/scripts/exregional_run_fcst.sh
@@ -515,22 +515,6 @@ Call to function to create a diag table file for the current cycle's
 #
 #-----------------------------------------------------------------------
 #
-# If running ensemble forecasts, create a link to the cycle-specific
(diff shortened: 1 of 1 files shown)


gsketefian (approved):
Looks good.

PR #559, opened 2021-07-23
Revert "Modifications to allow workflow to run to completion with GNU build on Hera"

JeffBeck-NOAA:
Reverts NOAA-EMC/regional_workflow#526

Reverting due to problems with missing Python environments on different NOAA HPC when building with the Intel compiler.  This PR will need to have "if" statements to avoid unloading when anything other than GNU is used.


gsketefian (approved):
We'll have to discuss with @JulieSchramm what the best fix is for this, then make a new PR.

JeffBeck-NOAA:
> We'll have to discuss with @JulieSchramm what the best fix is for this, then make a new PR.

Agreed.  Will merge for now and plan discussion with @JulieSchramm soon.

Issue #562, opened 2021-08-02
ncores_per_node in model_configure is not used


chan-hoo:
## Description
The parameter 'ncores_per_node' is not necessary in 'model_configure'.

## Solution
Remove it from the namelist

PR #563, opened 2021-08-02
Remove ncores_per_node from model_configure and set PPN_RUN_FCST in setup.sh

chan-hoo:
## DESCRIPTION OF CHANGES: 
- Since the parameter 'ncores_per_node' in 'model_configure' is not used, it is removed from the namelist.
- The optimal value of 'PPN_RUN_FCST' is calculated from 'NCORES_PER_NODE' and 'OMP_NUM_THREADS' in 'setup.sh'.
- The parameter 'CPU_PER_RUN_FCST' is removed because it can be specified from the existing parameter 'OMP_NUM_THREADS_RUN_FCST'.

## TESTS CONDUCTED: 
WE2E tests on Hera and WCOSS dell:

Comparison of CPU times on the WCOSS dell (unit: sec):
           domain            | current default  | new default
RRFS_CONUS 25km | 15380                | 8544
RRFS_CONUS 13km | 424470              | 35888
RRFS_CONUS 3km   | 2054889           | 605765

## ISSUE: 
Fixes issue mentioned in #562 

## CONTRIBUTORS: 
@JamesAbeles-NOAA


ush/setup.sh
```diff
@@ -579,6 +579,22 @@ esac
...
+#
+#-----------------------------------------------------------------------
+#
+OMP_NUM_THREADS_ALL="$(( ${OMP_NUM_THREADS_ATMOS} ))"
+OMP_NUM_THREADS_RUN_FCST=${OMP_NUM_THREADS_RUN_FCST:-${OMP_NUM_THREADS_ALL}}
+
+PPN_RUN_FCST_OPT="$(( ${NCORES_PER_NODE} / ${OMP_NUM_THREADS_RUN_FCST} ))"
+PPN_RUN_FCST=${PPN_RUN_FCST:-${PPN_RUN_FCST_OPT}}
+#
+#-----------------------------------------------------------------------
```
JeffBeck-NOAA, line 589: @chan-hoo, I'm having a hard time understanding the difference between:

OMP_NUM_THREADS_ALL
OMP_NUM_THREADS_ATMOS
OMP_NUM_THREADS_RUN_FCST

Are they all truly required?  Also, what is "alpha"?

chan-hoo: @JeffBeck-NOAA, for ufs weather model (fv3 alone), *_ATMOS=*_ALL=*_RUN_FCST. But in case of a coupled model,  *_ATMOS != *_ALL (not always). *_ALL is a temporary variable to separate from *_RUN_FCST because someone might want to input *_RUN_FCST arbitrarily. If a coupled model uses different number of OMP_NUMs between its components, *_RUN_FCST = max(*_ATMOS, *_COMP). I didn't put *_COMP into the script yet. If this is confused, we can remove *_ALL at this time. Let me know what you think.

chan-hoo: @JeffBeck-NOAA, I'll remove OMP_NUM_THREADS_ALL and OMP_NUM_THREADS_ATMOS. We can add them later when they are necessary.

chan-hoo: Removed

JeffBeck-NOAA: Thanks for the explanation, @chan-hoo, this looks good!

ush/config_defaults.sh
```diff
@@ -1697,11 +1694,9 @@ OMP_NUM_THREADS_MAKE_LBCS="1"
 OMP_STACKSIZE_MAKE_LBCS="1024m"
 
 KMP_AFFINITY_RUN_FCST="scatter"
-OMP_NUM_THREADS_RUN_FCST="4"
+OMP_NUM_THREADS_RUN_FCST="2"    # atmos_nthreads in model_configure
 OMP_STACKSIZE_RUN_FCST="1024m"
```
JeffBeck-NOAA, line 1698: @chan-hoo, one last comment:  Is this change (from 4 to 2) specifically for WCOSS, or is this better for all NOAA HPC?

chan-hoo: @JeffBeck-NOAA, I tested it on WCOSS and Hera. As described above, the CPU time on WCOSS was reduced significantly. On Hera, I couldn't find the CPU time in the log file. I only found the wall-clock time. The wall-clock times of "4" and "2" were quite close to each other. So, I think we can set the default value of OMP_NUM_THREADS_RUN_FCST to "2".

JeffBeck-NOAA: OK, I'm a bit confused, because if you look at the config_defaults.sh file in GitHub right now, it's already set to 2 [here](https://github.com/NOAA-EMC/regional_workflow/blob/99f6eb81b361a27cb8b7da2221506750e9387b2b/ush/config_defaults.sh#L1703).  I made this change a while ago since when it was set to "4" it was requesting twice the nodes that were originally defined in the config.sh file, which was excessive.  I'm not sure why your PR shows the current value as 4?

chan-hoo: @JeffBeck-NOAA, Sorry, I forgot one thing. There was another parameter (PPN_RUN_FCST) I changed in this test. So, it is hard to say the change from "4" to "2" made the improvement of CPU time. If you want, I'll test its impact. it is usually set to "2" at EMC.

JeffBeck-NOAA: Sorry, @chan-hoo, I see what's going on here.  You're replacing CPUS_PER_TASK_RUN_FCST with OMP_NUM_THREADS_RUN_FCST.

JeffBeck-NOAA: @chan-hoo, I think it's fine.  The "--cpus-per-task 2 --exclusive" settings have been tested on all platforms, since the "2" was previously set by CPUS_PER_TASK_RUN_FCST, so I think we're good!

JeffBeck-NOAA (approved):

ush/config_defaults.sh
chan-hoo: Yes. I removed CPUS_PER_TASK_RUN_FCST. CPUS_PER_TASK_RUN_FCST is only used in the workflow on Hera. The number of omp threads of FV3 is controlled by OMP_NUM_THREADS_RUN_FCST. They should be the same in ufs-weather-model.

Issue #566, opened 2021-08-04
Pazi parameter not set for the RRFS NA domains


JeffBeck-NOAA:
## Description

The pazi parameter needs to be set to -13 for both RRFS NA domains (3- and 13-km resolution) to be rotated correctly.

## Additional Context

Conversation with EMC (@BenjaminBlake-NOAA and @JacobCarley-NOAA).

PR #567, opened 2021-08-04
Add the pazi parameter option to the creation of the regional_esg_grid namelist; set pazi=-13.0 for the RRFS_NA_3km domain

JeffBeck-NOAA:
## DESCRIPTION OF CHANGES: 
Add the pazi parameter as a definable option to the workflow to allow for its addition to the regional_esg_grid namelist.  Set pazi to -13.0 for the RRFS_NA_3km domain to obtain the proper grid rotation angle.

## TESTS CONDUCTED: 
Test on Hera with the RRFS_NA_3km domain and with the RRFS_CONUS_3km domain, the latter of which does not have a rotation angle, but is correctly set to pazi=0.0.

## DOCUMENTATION:
Added in-line comments to explain the pazi parameter.

## ISSUE (optional): 
Resolves Issue #566.

## CONTRIBUTORS (optional): 
@BenjaminBlake-NOAA, @JacobCarley-NOAA, @gsketefian

### DIFF
--- a/ush/set_predef_grid_params.sh
+++ b/ush/set_predef_grid_params.sh
@@ -1187,6 +1187,8 @@ case ${PREDEF_GRID_NAME} in
   ESGgrid_NX=3640
   ESGgrid_NY=2520
 
+  ESGgrid_PAZI=-13.0
+
   ESGgrid_WIDE_HALO_WIDTH=6
 
   DT_ATMOS="${DT_ATMOS:-36}"
--- a/ush/set_predef_grid_params.sh
+++ b/ush/set_predef_grid_params.sh
@@ -112,6 +112,8 @@ case ${PREDEF_GRID_NAME} in
   ESGgrid_NX="202"
   ESGgrid_NY="116"
 
+  ESGgrid_PAZI="0.0"
+
   ESGgrid_WIDE_HALO_WIDTH="6"
 
   DT_ATMOS="${DT_ATMOS:-40}"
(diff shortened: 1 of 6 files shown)


ush/set_predef_grid_params.sh
```diff
@@ -1187,6 +1187,8 @@ case ${PREDEF_GRID_NAME} in
   ESGgrid_NX=3640
   ESGgrid_NY=2520
 
+  ESGgrid_PAZI=-13.0
```
gsketefian: After thinking about this a bit more, I think it would be better to explicitly set ESGgrid_PAZI to 0.0 for all the other ESG grids in this file and remove its default value from config_defaults.sh (i.e. set it to a null string there).  That's because I had intended this file (set_predef_grid_params.sh) to contain a complete description of each grid, and that's why I left all the other ESGgrid_... (and also GFDL_grid_...) parameters undefined in config_defaults.sh.  Doing this will make the definition of ESGgrid_PAZI consistent with the definitions of all the other grid parameters.

gsketefian (approved):
Looks good, I just have that one comment about defining ESGgrid_PAZI.

ush/set_predef_grid_params.sh
JeffBeck-NOAA: Yep, understand.  I just made the relevant changes.

BenjaminBlake-NOAA (approved):
This PR looks good to me.  I agree with Gerard's comments on setting PAZI to an empty value in config_defaults, and defining it explicitly for each grid in set_predef_grid_params.

JeffBeck-NOAA:
Thanks @BenjaminBlake-NOAA, @gsketefian, I've made the relevant changes, so I'll go ahead and merge.

Issue #486, opened 2021-05-10
Move mrms_pull_topofhour.py out of scripts directory


gsketefian:
The scripts directory should contain only ex-scripts (and the jobs directory should contain only j-jobs).  Other scripts should go into ush or in a subdirectory under ush (e.g. ush/Python).

PR #571, opened 2021-08-17
Move mrms_pull_topofhour.py from scripts/ to ush/ [develop]

jwolff-ncar:
## DESCRIPTION OF CHANGES: 
Moved the mrms_pull_topofhour.py from scripts/ into ush/.
Made a small change to the script to remove a 'tab' and make it 'spaces' instead. 
Added a set +u in the load_modules_run_task.sh to ensure the correct environment for running the python script.

## TESTS CONDUCTED: 
Tested one case on hera including the verification step. The verification tasks are not yet supported on other platforms.

## DEPENDENCIES:
None.

## DOCUMENTATION:
No updates needed.

## ISSUE (optional): 
Fixes issue mentioned in https://github.com/NOAA-EMC/regional_workflow/issues/486


JeffBeck-NOAA (commented):
Should ush/load_modules_run_task.sh have changed?

jwolff-ncar:
As I mentioned in the description of changes, I had to add the set +u to get the environment right for the python script. I had worked with @mkavulich on this so perhaps he wants to also take a look for approval.

JeffBeck-NOAA:
> As I mentioned in the description of changes, I had to add the set +u to get the environment right for the python script. I had worked with @mkavulich on this so perhaps he wants to also take a look for approval.

OK, if it's just to ensure the correct environment, it's likely only necessary for debugging, correct?

ush/load_modules_run_task.sh
```diff
@@ -331,7 +332,6 @@ if [ -n "${SRW_ENV:-}" ] ; then
   conda activate ${SRW_ENV}
 fi
```
mkavulich, line 335: If we want to be entirely consistent with previous behavior, we should include a "set -u" line here. But I do not think it's necessary

mkavulich:
@JeffBeck-NOAA The `set +u` is necessary prior to the call to `conda activate` on some machines due to the way the conda scripts work...they have some bash scripts with unset variables, which end up inheriting our `-u` flag which fails when it encounters unset variables. Ideally best practice would be to not have unset variables, but we do not have control over those scripts, so the next best thing is to use `set +u` to stop checking for unset variables prior to calling `conda activate`. This is the solution @christinaholtNOAA recommended.

ush/load_modules_run_task.sh
jwolff-ncar: I am pretty sure I tested with that and it didn't work so I would say let's not add that if it isn't deemed necessary.

christinaholtNOAA:
The way we approached it for the changes we introduced for Jet/Hera in RRFS_dev1 is [here](https://github.com/NOAA-GSL/regional_workflow/blob/feature/RRFS_dev1/ush/load_modules_run_task.sh) starting at line 332. I suggest wrapping only the single line that needs it (and is out of our control) in the `set +u` and `set -u`.

JeffBeck-NOAA:
Thanks @mkavulich, @christinaholtNOAA.  I'm still a bit confused, since I've not had any problems on Hera with the load_modules_run_task script before this PR.

jwolff-ncar:
@JeffBeck-NOAA you would have only run into this problem if you were pulling/processing MRMS data from the HPSS rather than using staged observations. It was not an issue when the capability was first added but noticed about 1.5 months ago so something changed and it is now required.

JeffBeck-NOAA (approved):

jwolff-ncar:
@mkavulich Were you going to make a change to include the the set +u and set -u updates for me to test?

Issue #576, opened 2021-08-22
The executable name and output prefixes of the 'run_post' task needs to be parameterized


chan-hoo:
## Description
The executable name and output prefixes of 'EMC_post' in the recent hashes are different from those of the current one in the SRW App:
1. Executable name: ncep_post (current) vs. upp.x (recent)
2. Output prefixes: bgdawp, bgrd3d (current) vs. prslev, natlev (recent)

PR #577, opened 2021-08-22
The executable name and output prefixes of 'run_post' are parameterized

chan-hoo:
## DESCRIPTION OF CHANGES: 
The executable name and output prefixes of the ‘run_post’ task are parameterized so that they can be specified in the configuration file with the new parameters POST_EXEC_FN and POST_FIDS.

## TESTS CONDUCTED: 
1. 6h GST tests with the current and latest ‘EMC_post’:
For the latest ‘EMC_post’: add the following two parameters to config.sh: 
POST_EXEC_FN=”upp.x”
POST_FIDS=( “prslev” “natlev” )

2. WE2E test: MET_verification

## ISSUE: 
Fixes issue mentioned in #576

### DIFF
--- a/scripts/exregional_run_post.sh
+++ b/scripts/exregional_run_post.sh
@@ -276,7 +276,7 @@ EOF
 print_info_msg "$VERBOSE" "
 Starting post-processing for fhr = $fhr hr..."
 
-${APRUN} ${EXECDIR}/ncep_post < itag || print_err_msg_exit "\
+${APRUN} ${EXECDIR}/${POST_EXEC_FN} < itag || print_err_msg_exit "\
 Call to executable to run post for forecast hour $fhr returned with non-
 zero exit code."
 #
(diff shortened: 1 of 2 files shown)


JeffBeck-NOAA:
@chan-hoo, we handle the names of the executables in the ufs-srweather-app repository when they are copied to the /bin directory, so should we not modify it there?  Also, I don't anticipate the names of the executables or output prefixes for post to change continually, so a user-defined options for executable name/output prefix may be unnecessary.  It may just be enough to make the change once and update the post hash in Externals.cfg.  I'd like to hear thoughts on this from others, @gsketefian,@mkavulich, @jwolff-ncar, @llpcarson, @JulieSchramm.

chan-hoo:
@JeffBeck-NOAA , OK. I'll wait for your decision. RRFS-CMAQ uses the latest version of EMC_post and its executable name (upp.x) was not modified to 'ncep_post' when the app was compiled. So I had to modify the script for this.

JeffBeck-NOAA:
> @JeffBeck-NOAA , OK. I'll wait for your decision. RRFS-CMAQ uses the latest version of EMC_post and its executable name (upp.x) was not modified to 'ncep_post' when the app was compiled. So I had to modify the script for this.

@chan-hoo, yep, makes sense.  My hope is that these changes are settling down and becoming final.

Issue #605, opened 2021-09-23
Cron command is not removed from crontab once all tasks have completed


chan-hoo:
- When 'USE_CRON_TO_RELAUNCH'="TRUE", the cron command is not removed from the crontab file once all the tasks have completed.
- The log.launch_FV3LAM_wflow file shows 'SED: unbound variable' in the corresponding line.

gsketefian:
@chan-hoo I know this is an issue on Cheyenne.  Which machine did you encounter this?

chan-hoo:
@gsketefian, I had this issue on the WCOSS dell. I've fixed this (PR https://github.com/NOAA-EMC/regional_workflow/pull/606).

PR #606, opened 2021-09-23
Fix unbound SED for cron commands in the launch script

chan-hoo:
## DESCRIPTION OF CHANGES: 
To fix the issue on the launch script when 'USE_CRON_TO_RELAUNCH'="TRUE", '$SED' is modified to 'sed' in the script.

## TESTS CONDUCTED: 
WE2E tests on WCOSS_DELL_P3:
- MET_ensemble_verification

## ISSUE: 
Fixes issue mentioned in #605

### DIFF
--- a/ush/launch_FV3LAM_wflow.sh
+++ b/ush/launch_FV3LAM_wflow.sh
@@ -405,7 +405,7 @@ launch script for this experiment:
 # CRONTAB_LINE with backslashes.  Do this next.
 #
     crontab_line_esc_astr=$( printf "%s" "${CRONTAB_LINE}" | \
-                             $SED -r -e "s%[*]%\\\\*%g" )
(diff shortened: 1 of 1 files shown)


gsketefian:
@chan-hoo I think the fix you're implementing will cause the workflow to break on MACOS.  We should define `SED` as in other scripts, which is by sourcing `source_util_funcs.sh`.  You can do that by including the following lines in the launch script right after where `scrfunc_dir` is defined (I haven't debugged this though!):
```
homerrfs=${scrfunc_dir%/*}
ushdir="$homerrfs/ush"
. $ushdir/source_util_funcs.sh
```
This is the same approach as, for example, in `run_WE2E_tests.sh`.  Thanks! @mkavulich

chan-hoo:
@gsketefian , modified as suggested and tested successfully.

Issue #610, opened 2021-09-29
Add support for Rocoto with generic LINUX platform


christinaholtNOAA:
## Description
PR #402  added a capability to use a generic LINUX platform, but did not include support for workflow management with Rocoto.  

## Solution
The RRFS Cloud work required the use of Rocoto on a generic LINUX platform, and we would like to contribute that work back to the repository for a more complete functionality outside the supported platforms.

## Related to (optional)
PR #402

mkavulich:
Addressed by #617

PR #580, opened 2021-08-25
Reorganize the way external model files are obtained

gsketefian:
### Summary:
This PR:
* Allows the user to:
1) Specify which external model data source(s) to try obtaining files from and in which order (via the new experiment variable `EXTRN_MDL_DATA_SOURCES`).
2) Specify what directory structure and file naming convention to assume when searching for external model files on disk (via the new experiment variable `EXTRN_MDL_DIR_FILE_LAYOUT`).  The location on disk may be a user-specified directory or a system directory.
3) Try fetching FV3GFS grib2 files from NOMADS (NOAA Operational Model Archive and Distribution System) during the experiment generation step, and, if successful, placing the files in the correct locations in the experiment directory and turning off the `get_extrn_ics` and `get_extrn_lbcs` workflow tasks.  This is useful because on many HPC machines, the compute nodes (on which the rocoto tasks run) do not have online access while the front-end/login nodes do.
* Provides a new stand-alone script (`get_FV3GFS_grib2_files_from_NOMADS.sh`) for fetching FV3GFS grib2 files from NOMADS for multiple cycles and for analysis and/or forecast files and placing the files in user-specified locations.
* Adds WE2E tests to check the capability of the workflow scripts to obtain external model files from various sources and for various external models and file formats.

### Details:
* Introduce the new experiment array variable `EXTRN_MDL_DATA_SOURCES` that specifies the data sources from which to try to obtain external model files.  The elements of this array can have one of the following values:
  * `"disk"`
     Try to copy or link to files on disk.  The set of base directories in which to look for the files are specified in the new experiment array variables `EXTRN_MDL_BASEDIRS_ICS` and `EXTRN_MDL_BASEDIRS_LBCS`.  These are user-specifiable and may contain multiple directories to search.  Note that the workflow generation scripts append to these arrays a system directory if one is known for the given machine and external model.  Thus, if the files are not found in the user-specified base directories, the system base directories are also searched.  (Here, by "system directory", we mean a directory in which output files from a given model are regularly placed, e.g. every 6 hours, for use by any user on the machine.  These directories usually contain only recent files, e.g. ones from the last 3 days.)
  * `"noaa_hpss"`
    Try to fetch files from NOAA HPSS (mass store).
  * `"nomads"`
    Try to fetch files from NOMADS (NOAA Operational Model Archive and Distribution System).

[...]

PR #617, opened 2021-10-15
Add support for Rocoto with generic LINUX platform

christinaholtNOAA:
## DESCRIPTION OF CHANGES: 
This modification allows for a user to run with Rocoto on their own Linux-based platform, which still requires modules to be installed. 

## TESTS CONDUCTED: 
Ran with the corresponding linux_target branch on Hera. Passed the grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GSD_SAR. 

Tests for the new functionality were conducted on the RRFS Cloud AWS platform and within a Docker container. Currently no additional test was added to the test framework for this capability since we have no intent on supporting any defined linux configuration.

## DEPENDENCIES:
- https://github.com/ufs-community/ufs-srweather-app/pull/184

## DOCUMENTATION:
I will need guidance on where and how to add documentation for this addition, if it is necessary.

## ISSUE (optional): 
This PR addresses PR #610 

## CONTRIBUTORS (optional): 
@christopherwharrop-noaa

### DIFF
--- a/ush/launch_FV3LAM_wflow.sh
+++ b/ush/launch_FV3LAM_wflow.sh
@@ -84,6 +84,7 @@ fi
 #-----------------------------------------------------------------------
 #
 . $exptdir/var_defns.sh
+. ${USHDIR}/source_util_funcs.sh
 #
 #-----------------------------------------------------------------------
 #
@@ -101,13 +102,7 @@ expt_name="${EXPT_SUBDIR}"
 #
 #-----------------------------------------------------------------------
 #
-if [ "$MACHINE" = "CHEYENNE" ]; then
-  module use -a /glade/p/ral/jntp/UFS_SRW_app/modules/
-  module load rocoto
-elif [ "$MACHINE" = "ORION" ]; then
-  module purge
-  module load contrib rocoto
-elif [ "$MACHINE" = "WCOSS_DELL_P3" ]; then
+if [ "$MACHINE" = "WCOSS_DELL_P3" ]; then
   module purge
   module load lsf/10.1
   module use /gpfs/dell3/usrx/local/dev/emc_rocoto/modulefiles/
@@ -118,8 +113,14 @@ elif [ "$MACHINE" = "WCOSS_CRAY" ]; then
   module use -a /usrx/local/emc_rocoto/modulefiles
   module load rocoto/1.3.0rc2
 else
+  machine=$(echo_lowercase $MACHINE)
+  env_fn=${WFLOW_ENV_FN:-"wflow_${machine}.env"}
+  env_fp="${SR_WX_APP_TOP_DIR}/env/${env_fn}"
   module purge
-  module load rocoto
+  source "${env_fp}" || print_err_msg_exit "\
+    Sourcing platform-specific environment file (env_fp) for
+  the workflow task failed :
+      env_fp = \"${env_fp}\""
 fi
 #
 #-----------------------------------------------------------------------
@@ -162,35 +163,7 @@ cd "$exptdir"
 #-----------------------------------------------------------------------
 #
 
-#rocotorun_output=$( ls -alF )
-#echo
-#echo "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
-#echo "${rocotorun_output}"
-#echo "BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB"
-
-#rocotorun_output=$( \
-#rocotorun -w "${WFLOW_XML_FN}" -d "${rocoto_database_fn}" -v 10 \
-#)
-#rocotorun_output=$( (rocotorun -w "${WFLOW_XML_FN}" -d "${rocoto_database_fn}" -v 10) 2>&1 )  # This freezes the script.
-#rocotorun_output=$( (rocotorun -w "${WFLOW_XML_FN}" -d "${rocoto_database_fn}" -v 10) 1>&2 )  # This leaves rocotorun_output empty.
-#rocotorun_output=$( rocotorun -w "${WFLOW_XML_FN}" -d "${rocoto_database_fn}" -v 10 )
-#{ error=$(command 2>&1 1>&$out); } {out}>&1
-#{ rocotorun_output=$( rocotorun -w "${WFLOW_XML_FN}" -d "${rocoto_database_fn}" -v 10 2>&1 1>&$out); } {out}>&1  # This freezes the script.
-
-#
-# Ideally, the following two lines should work, but for some reason the
-# output of rocotorun cannot be captured in a variable using the $(...)
-# notation.  Maybe it's not being written to stdout, although I tried
-# redirecting stderr to stdout and other tricks but nothing seemed to
-# work.  For this reason, below we first redirect the output of rocoto-
-# run to a temporary file and then read in the contents of that file in-
-# to the rocotorun_output variable using the cat command.
-#
-#rocotorun_cmd="rocotorun -w \"${WFLOW_XML_FN}\" -d \"${rocoto_database_fn}\" -v 10"
-#rocotorun_output=$( eval ${rocotorun_cmd} 2>&1 )
-#
 tmp_fn="rocotorun_output.txt"
-#rocotorun_cmd="rocotorun -w \"${WFLOW_XML_FN}\" -d \"${rocoto_database_fn}\" -v 10 > ${tmp_fn}"
 rocotorun_cmd="rocotorun -w \"${WFLOW_XML_FN}\" -d \"${rocoto_database_fn}\" -v 10"
 eval ${rocotorun_cmd} > ${tmp_fn} 2>&1
 rocotorun_output=$( cat "${tmp_fn}" )
@@ -217,18 +190,9 @@ done <<< "${rocotorun_output}"
 #
 #-----------------------------------------------------------------------
 #
-#rocotostat_cmd="{ pwd; rocotostat -w \"${WFLOW_XML_FN}\" -d \"${rocoto_database_fn}\" -v 10; }"
-#rocotostat_cmd="{ pwd; ls -alF; rocotostat -w ${WFLOW_XML_FN} -d ${rocoto_database_fn} -v 10; }"
-#rocotostat_cmd="{ pwd; ls -alF; rocotostat -w \"${WFLOW_XML_FN}\" -d \"${rocoto_database_fn}\" -v 10; }"
-#rocotostat_cmd="{ pwd; rocotostat -w \"${WFLOW_XML_FN}\" -d \"${rocoto_database_fn}\" -v 10; }"
-#rocotostat_cmd="{ rocotostat -w \"${WFLOW_XML_FN}\" -d \"${rocoto_database_fn}\" -v 10; }"
 rocotostat_cmd="rocotostat -w \"${WFLOW_XML_FN}\" -d \"${rocoto_database_fn}\" -v 10"
 
-#rocotostat_output=$( pwd; rocotostat -w "${WFLOW_XML_FN}" -d "${rocoto_database_fn}" -v 10 2>&1 )
-#rocotostat_output=$( rocotostat -w "${WFLOW_XML_FN}" -d "${rocoto_database_fn}" -v 10 2>&1 )
 rocotostat_output=$( eval ${rocotostat_cmd} 2>&1 )
-#rocotostat_output=$( ${rocotostat_cmd} 2>&1 )
-#rocotostat_output=$( { pwd; ls -alF; } 2>&1 )
 error_msg="DEAD"
 while read -r line; do
   grep_output=$( printf "$line" | grep "${error_msg}" )
--- a/ush/load_modules_run_task.sh
+++ b/ush/load_modules_run_task.sh
@@ -71,30 +71,6 @@ fi
 # ..." and "module load ..." calls later below that are used to load the
 # appropriate module file for the specified task.
 #
-# Note that the build of the FV3 forecast model code generates the shell
-# script at
-#
-#   ${UFS_WTHR_MDL_DIR}/NEMS/src/conf/module-setup.sh
-#
-# that can be used to initialize the Lmod (Lua-based module) system/
-# software for handling modules.  This script:
-#
-# 1) Detects the shell in which it is being invoked (i.e. the shell of
-#    the "parent" script in which it is being sourced).
-# 2) Detects the machine it is running on and and calls the appropriate
-#    (shell- and machine-dependent) initalization script to initialize
-#    Lmod.
-# 3) Purges all modules.
-# 4) Uses the "module use ..." command to prepend or append paths to
-#    Lmod's search path (MODULEPATH).
-#
-# We could use this module-setup.sh script to initialize Lmod, but since
-# it is only found in the forecast model's directory tree, here we pre-
-# fer to perform our own initialization.  Ideally, there should be one
-# module-setup.sh script that is used by all external repos/codes, but
-# such a script does not exist.  If/when it does, we will consider
-# switching to it instead of using the case-statement below.
-#
 #-----------------------------------------------------------------------
 #
 print_info_msg "$VERBOSE" "
@@ -128,10 +104,14 @@ case "$MACHINE" in
     ;;
 #
   *)
-    print_err_msg_exit "\
-The script to source to initialize lmod (module loads) has not yet been
-specified for the current machine (MACHINE):
-  MACHINE = \"$MACHINE\""
+    if [[ -n ${LMOD_PATH:-""} && -f ${LMOD_PATH:-""} ]] ; then
+      . ${LMOD_PATH}
+    else
+      print_err_msg_exit "\
+      The script to source to initialize lmod (module loads) has not yet been
+      specified for the current machine (MACHINE):
+        MACHINE = \"$MACHINE\""
+    fi
     ;;
 #
 esac
@@ -147,13 +127,15 @@ jjob_fp="$2"
 #
 #-----------------------------------------------------------------------
 #
-# Sourcing ufs-srweather-app README file (in directory specified by mod-
-# ules_dir) for the specified task
+# Sourcing ufs-srweather-app build env file
 #
 #-----------------------------------------------------------------------
 #
+
+module purge
+
 machine=$(echo_lowercase $MACHINE)
-env_fn="build_${machine}_${COMPILER}.env"
+env_fn=${BUILD_ENV_FN:-"build_${machine}_${COMPILER}.env"}
 env_fp="${SR_WX_APP_TOP_DIR}/env/${env_fn}"
 source "${env_fp}" || print_err_msg_exit "\
 Sourcing platform- and compiler-specific environment file (env_fp) for the 
@@ -172,25 +154,23 @@ workflow task specified by task_name failed:
 # sets environment variables (including prepending/appending to paths)
 # and loads modules.
 #
-# The regional_workflow repository contains module files for all the
+# The regional_workflow repository contains module files for the
 # workflow tasks in the template rocoto XML file for the FV3-LAM work-
-# flow.  The full path to a module file for a given task is
+# flow that need modules not loaded in the env_fn above.
 #
-#   $HOMErrfs/modulefiles/$machine/${task_name}
+# The full path to a module file for a given task is
+#
+#   $HOMErrfs/modulefiles/$machine/${task_name}.local
 #
 # where HOMErrfs is the base directory of the workflow, machine is the
 # name of the machine that we're running on (in lowercase), and task_-
-# name is the name of the current task (an input to this script). The
-# collection of modulefiles is staged by the generate_workflow.sh
-# script. Please see that script for information on their creation.
+# name is the name of the current task (an input to this script).
 #
 #-----------------------------------------------------------------------
 #
 modules_dir="$HOMErrfs/modulefiles/tasks/$machine"
 modulefile_name="${task_name}"
 default_modules_dir="$HOMErrfs/modulefiles"
-default_modulefile_name="${machine}.default"
-use_default_modulefile=0
 #
 #-----------------------------------------------------------------------
 #
@@ -198,46 +178,28 @@ use_default_modulefile=0
 #
 #-----------------------------------------------------------------------
 #
-  print_info_msg "$VERBOSE" "
+
+print_info_msg "$VERBOSE" "
 Loading modules for task \"${task_name}\" ..."
 
-  module use "${modules_dir}" || print_err_msg_exit "\
+module use "${modules_dir}" || print_err_msg_exit "\
 Call to \"module use\" command failed."
 
-  #
-  # If NOT using the default modulefile...
-  #
-#  if [ ${use_default_modulefile} -eq 0 ]; then
 #
-#     module use -a "${modules_dir}" || print_err_msg_exit "\
-#Call to \"module use\" command failed."
-#    
-    #
-    # Load the .local module file if available for the given task
-    #
-    modulefile_local="${task_name}.local"
-    if [ -f ${modules_dir}/${modulefile_local} ]; then
-      module load "${modulefile_local}" || print_err_msg_exit "\
-Loading .local module file (in directory specified by mod-
-ules_dir) for the specified task (task_name) failed:
-  task_name = \"${task_name}\"
-  modulefile_local = \"${modulefile_local}\"
-  modules_dir = \"${modules_dir}\""    
-    fi
-
-#  else # using default modulefile
+# Load the .local module file if available for the given task
 #
-#    module load "${default_modulefile_name}" || print_err_msg_exit "\
-#Loading of default module file failed:
-#  task_name = \"${task_name}\"
-#  default_modulefile_name = \"${default_modulefile_name}\"
-#  default_modules_dir = \"${default_modules_dir}\""
-#
-#  fi
+modulefile_local="${task_name}.local"
+if [ -f ${modules_dir}/${modulefile_local} ]; then
+  module load "${modulefile_local}" || print_err_msg_exit "\
+  Loading .local module file (in directory specified by mod-
+  ules_dir) for the specified task (task_name) failed:
+    task_name = \"${task_name}\"
+    modulefile_local = \"${modulefile_local}\"
+    modules_dir = \"${modules_dir}\""    
+fi
 
-  module list
+module list
 
-#fi #End if statement for tasks that load no modules
 
 # Modules that use conda and need an environment activated will set the
 # SRW_ENV variable to the name of the environment to be activated. That
--- a/ush/setup.sh
+++ b/ush/setup.sh
@@ -668,9 +668,8 @@ check_var_valid_value "SCHED" "valid_vals_SCHED"
 #
 #-----------------------------------------------------------------------
 #
-# If we are using a workflow manager, run some checks. First, 
-# verify that the ACCOUNT variable is not empty. Second, ensure that the
-# custom RUN_CMD variables are not set.
+# If we are using a workflow manager check that the ACCOUNT variable is
+# not empty.
 #
 #-----------------------------------------------------------------------
 #
--- a/ush/setup.sh
+++ b/ush/setup.sh
@@ -681,9 +681,6 @@ The variable ACCOUNT cannot be empty if you are using a workflow manager:
   ACCOUNT = \"$ACCOUNT\"
   WORKFLOW_MANAGER = \"$WORKFLOW_MANAGER\""
   fi
-  RUN_CMD_UTILS=""
-  RUN_CMD_FCST=""
-  RUN_CMD_POST=""
 fi
 #
 #-----------------------------------------------------------------------
@@ -2949,6 +2946,15 @@ Heredoc (cat) command to append grid parameters to variable definitions
 file returned with a nonzero status."
 
 fi
+#
+#-----------------------------------------------------------------------
+#
+# Because RUN_CMD_FCST can include PE_MEMBER01 (and theoretically other
+# variables calculated in this script), delete the first occurance of it
+# in the var_defns file, and write it again at the end.
+#
+#-----------------------------------------------------------------------
+$SED -i '/^RUN_CMD_FCST=/d' $GLOBAL_VAR_DEFNS_FP
 #
 #-----------------------------------------------------------------------
 #
--- a/ush/setup.sh
+++ b/ush/setup.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 #
 #-----------------------------------------------------------------------
 #
@@ -512,7 +513,7 @@ check_var_valid_value "MACHINE" "valid_vals_MACHINE"
 # several queues.  These queues are defined in the default and local 
 # workflow/experiment configuration script.
 #
-# Also, set the machine-dependent flag RELAITVE_OR_NULL that specifies
+# Also, set the machine-dependent flag RELATIVE_OR_NULL that specifies
 # the flag to pass to the link creation command (ln_vrfy) when attempting 
 # to create relative symlinks.  On machines that don't support relative
 # symlinks, it should be set to a null string.
@@ -520,12 +521,11 @@ check_var_valid_value "MACHINE" "valid_vals_MACHINE"
 #-----------------------------------------------------------------------
 #
 RELATIVE_LINK_FLAG=""
-NCORES_PER_NODE="2" # Need some arbitrary default value to avoid division by zero errors
 case $MACHINE in
 
   "WCOSS_CRAY")
     WORKFLOW_MANAGER="rocoto"
-    NCORES_PER_NODE="24"
+    NCORES_PER_NODE="${NCORES_PER_NODE:-24}"
     SCHED="lsfcray"
     QUEUE_DEFAULT=${QUEUE_DEFAULT:-"dev"}
     QUEUE_HPSS=${QUEUE_HPSS:-"dev_transfer"}
@@ -536,7 +536,7 @@ case $MACHINE in
 
   "WCOSS_DELL_P3")
     WORKFLOW_MANAGER="rocoto"
-    NCORES_PER_NODE=24
+    NCORES_PER_NODE="${NCORES_PER_NODE:-24}"
     SCHED="lsf"
     QUEUE_DEFAULT=${QUEUE_DEFAULT:-"dev"}
     QUEUE_HPSS=${QUEUE_HPSS:-"dev_transfer"}
@@ -547,7 +547,7 @@ case $MACHINE in
 
   "HERA")
     WORKFLOW_MANAGER="rocoto"
-    NCORES_PER_NODE=40
+    NCORES_PER_NODE="${NCORES_PER_NODE:-40}"
     SCHED=${SCHED:-"slurm"}
     PARTITION_DEFAULT=${PARTITION_DEFAULT:-"hera"}
     QUEUE_DEFAULT=${QUEUE_DEFAULT:-"batch"}
@@ -561,7 +561,7 @@ case $MACHINE in
 
   "ORION")
     WORKFLOW_MANAGER="rocoto"
-    NCORES_PER_NODE=40
+    NCORES_PER_NODE="${NCORES_PER_NODE:-40}"
     SCHED=${SCHED:-"slurm"}
     PARTITION_DEFAULT=${PARTITION_DEFAULT:-"orion"}
     QUEUE_DEFAULT=${QUEUE_DEFAULT:-"batch"}
@@ -575,7 +575,7 @@ case $MACHINE in
 
   "JET")
     WORKFLOW_MANAGER="rocoto"
-    NCORES_PER_NODE=24
+    NCORES_PER_NODE="${NCORES_PER_NODE:-24}"
     SCHED=${SCHED:-"slurm"}
     PARTITION_DEFAULT=${PARTITION_DEFAULT:-"sjet,vjet,kjet,xjet"}
     QUEUE_DEFAULT=${QUEUE_DEFAULT:-"batch"}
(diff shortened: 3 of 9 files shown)


ush/launch_FV3LAM_wflow.sh
```diff
@@ -118,8 +113,14 @@ elif [ "$MACHINE" = "WCOSS_CRAY" ]; then
...
 else
+  machine=$(echo_lowercase $MACHINE)
+  env_fn=${WFLOW_ENV_FN:-"wflow_${machine}.env"}
+  env_fp="${SR_WX_APP_TOP_DIR}/env/${env_fn}"
   module purge
-  module load rocoto
+  source "${env_fp}" || print_err_msg_exit "\
+    Sourcing platform-specific environment file (env_fp) for
+  the workflow task failed :
+      env_fp = \"${env_fp}\""
```
christinaholtNOAA, line 123: A user will be defining this in their own wflow env file, anyway, so have the supported platforms do the same for consistency. I do not have enough familiarity with the WCOSS machines to make changes there.

ush/launch_FV3LAM_wflow.sh
```diff
@@ -162,35 +163,7 @@ cd "$exptdir"
...
-# output of rocotorun cannot be captured in a variable using the $(...)
-# notation.  Maybe it's not being written to stdout, although I tried
-# redirecting stderr to stdout and other tricks but nothing seemed to
-# work.  For this reason, below we first redirect the output of rocoto-
-# run to a temporary file and then read in the contents of that file in-
-# to the rocotorun_output variable using the cat command.
-#
-#rocotorun_cmd="rocotorun -w \"${WFLOW_XML_FN}\" -d \"${rocoto_database_fn}\" -v 10"
-#rocotorun_output=$( eval ${rocotorun_cmd} 2>&1 )
-#
```
christinaholtNOAA, line 191: If none of this works, it should not be kept in comments, cluttering the code. It's in the history of the repository for reference.

ush/load_modules_run_task.sh
```diff
@@ -71,30 +71,6 @@ fi
...
-# 4) Uses the "module use ..." command to prepend or append paths to
-#    Lmod's search path (MODULEPATH).
-#
-# We could use this module-setup.sh script to initialize Lmod, but since
-# it is only found in the forecast model's directory tree, here we pre-
-# fer to perform our own initialization.  Ideally, there should be one
-# module-setup.sh script that is used by all external repos/codes, but
-# such a script does not exist.  If/when it does, we will consider
-# switching to it instead of using the case-statement below.
-#
```
christinaholtNOAA, line 97: This is likely outdated as it documents the behavior of another script in another repository. It is confusing to live here as it does not document the code that is actually being run in this script. It will be in the history of the repo for reference.

ush/load_modules_run_task.sh
```diff
@@ -172,72 +154,52 @@ workflow task specified by task_name failed:
...
-  modules_dir = \"${modules_dir}\""    
-    fi
-
-#  else # using default modulefile
 #
-#    module load "${default_modulefile_name}" || print_err_msg_exit "\
-#Loading of default module file failed:
-#  task_name = \"${task_name}\"
-#  default_modulefile_name = \"${default_modulefile_name}\"
-#  default_modules_dir = \"${default_modules_dir}\""
```
christinaholtNOAA, line 234: All of this is outdated, and should be kept in the history of the repo, not in commented code.

ush/setup.sh
```diff
@@ -1,3 +1,4 @@
+#!/bin/bash
```
christinaholtNOAA, line 1: Should be set explicitly.

ush/setup.sh
```diff
@@ -671,9 +681,6 @@ The variable ACCOUNT cannot be empty if you are using a workflow manager:
   ACCOUNT = \"$ACCOUNT\"
   WORKFLOW_MANAGER = \"$WORKFLOW_MANAGER\""
   fi
-  RUN_CMD_UTILS=""
-  RUN_CMD_FCST=""
-  RUN_CMD_POST=""
```
christinaholtNOAA, line 676: Don't clear these, we need them for running properly on generic platform.

JeffBeck-NOAA:
> We would like to use this PR to test some of our CI automation work, too. Could someone with appropriate permissions add the `ci-hera-intel-WE` label to this PR? Thanks!

Done!

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch2/BMC/zrtrr/rrfs_ci/autoci/pr/759317963/20211019142104/ufs-srweather-app
Build was Successful
ci-hera-intel-WE

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch2/BMC/zrtrr/rrfs_ci/autoci/pr/759317963/20211019180130/ufs-srweather-app
Build was Successful
Script /scratch2/BMC/zrtrr/rrfs_ci/autoci/pr/759317963/20211019180130/ufs-srweather-app/regional_workflow/tests/WE2E/end_to_end_tests.sh does not exist in repo
Cannot run WE2E tests
ci-hera-intel-WE

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch2/BMC/zrtrr/rrfs_ci/autoci/pr/759317963/20211029214818/ufs-srweather-app
Build was Successful
Script /scratch2/BMC/zrtrr/rrfs_ci/autoci/pr/759317963/20211029214818/ufs-srweather-app/regional_workflow/tests/WE2E/end_to_end_tests.sh does not exist in repo
Cannot run WE2E tests
ci-hera-intel-WE

christinaholtNOAA:
Will anyone be able to review this PR soon?

ush/launch_FV3LAM_wflow.sh
JeffBeck-NOAA: @gsketefian, can you provide input on the comments in the code here?  It looks like the method has changed to query "rocotorun", so maybe we only need a single comment block with the relevant commands?

ush/launch_FV3LAM_wflow.sh
```diff
@@ -162,35 +163,7 @@ cd "$exptdir"
...
 tmp_fn="rocotorun_output.txt"
-#rocotorun_cmd="rocotorun -w \"${WFLOW_XML_FN}\" -d \"${rocoto_database_fn}\" -v 10 > ${tmp_fn}"
 rocotorun_cmd="rocotorun -w \"${WFLOW_XML_FN}\" -d \"${rocoto_database_fn}\" -v 10"
```
JeffBeck-NOAA, line 167: @gsketefian, same question here (line 193).  Is this the old method?  No longer relevant?

ush/launch_FV3LAM_wflow.sh
```diff
@@ -217,18 +190,9 @@ done <<< "${rocotorun_output}"
...
-#rocotostat_cmd="{ pwd; rocotostat -w \"${WFLOW_XML_FN}\" -d \"${rocoto_database_fn}\" -v 10; }"
-#rocotostat_cmd="{ rocotostat -w \"${WFLOW_XML_FN}\" -d \"${rocoto_database_fn}\" -v 10; }"
 rocotostat_cmd="rocotostat -w \"${WFLOW_XML_FN}\" -d \"${rocoto_database_fn}\" -v 10"
 
-#rocotostat_output=$( pwd; rocotostat -w "${WFLOW_XML_FN}" -d "${rocoto_database_fn}" -v 10 2>&1 )
-#rocotostat_output=$( rocotostat -w "${WFLOW_XML_FN}" -d "${rocoto_database_fn}" -v 10 2>&1 )
 rocotostat_output=$( eval ${rocotostat_cmd} 2>&1 )
-#rocotostat_output=$( ${rocotostat_cmd} 2>&1 )
-#rocotostat_output=$( { pwd; ls -alF; } 2>&1 )
 error_msg="DEAD"
```
JeffBeck-NOAA, line 196: @gsketefian, are any of these lines of commented code useful/still relevant?

ush/load_modules_run_task.sh
JeffBeck-NOAA: @gsketefian, can this block of comments be moved to the UFS SRW App documentation?  If so, where?

ush/load_modules_run_task.sh
JeffBeck-NOAA: @gsketefian, I think these commented lines were part of the restructuring of the module file loads, and we can probably remove them.  Agreed?

ush/launch_FV3LAM_wflow.sh
christinaholtNOAA: If these are indeed still used, they should be well commented with reasons to keep them around. Otherwise they should be left to the history of the repo because the lead to confusion rather than what the code needs to be doing. 

These particular lines seem to be currently irrelevant debugging attempts to get the right information stored in the variable. Since that has been worked out in line 229, they should be discarded.

ush/launch_FV3LAM_wflow.sh
christinaholtNOAA: This seems to be from a previous attempt to store the output on disk instead of being stored in memory. Since this solution has been resolved, it should not be kept here as a comment unless it is well documented with the reason to be kept. Otherwise, it's misleading and confusing.

ush/setup.sh
JeffBeck-NOAA: @christinaholtNOAA, can you please remove the sentence in the comments above referring to removal of the custom RUN_CMD variables?  @gsketefian, I'm guessing this is more legacy code, since we don't use the RUN_CMD_* variables for any NOAA HPC platforms (and there's therefore no longer a need to set them as empty)?

christinaholtNOAA: Sure thing. Totally missed that comment!

ush/setup.sh
```diff
@@ -2942,6 +2949,15 @@ fi
...
+# Because RUN_CMD_FCST can include PE_MEMBER01 (and theoretically other
+# variables calculated in this script), delete the first occurance of it
+# in the var_defns file, and write it again at the end.
+#
+#-----------------------------------------------------------------------
+$SED -i '/^RUN_CMD_FCST=/d' $GLOBAL_VAR_DEFNS_FP
+#
+#-----------------------------------------------------------------------
+#
 # Continue appending variable definitions to the variable definitions
```
JeffBeck-NOAA, line 2965: Can we not just write the RUN_CMD_FCST field to var_defns.sh file after all other (related) fields are set?

I tried running the generate script (after commenting out the reset lines for the RUN_CMD_* fields), and I see the following in the var_defns.sh file:

PE_MEMBER01="12"
RUN_CMD_FCST="mpirun -np ${PE_MEMBER01}"

Is this not what we'd expect?

ush/templates/FV3LAM_wflow.xml
```diff
@@ -414,7 +414,7 @@ MODULES_RUN_TASK_FP script.
 
     &RSRV_FCST;
     <command>&LOAD_MODULES_RUN_TASK_FP; "&RUN_FCST_TN;" "&JOBSDIR;/JREGIONAL_RUN_FCST"</command>
-  {%- if machine in ["JET", "HERA"]  %}
+  {%- if machine in ["JET", "HERA", "LINUX"]  %}
     <cores>{{ ncores_run_fcst }}</cores>
```
JeffBeck-NOAA, line 418: Can we assume that all LINUX-based systems will need `<cores>` and `<native>` in the workflow as follows?

'ncores_run_fcst': ${PE_MEMBER01}
'native_run_fcst': --cpus-per-task ${OMP_NUM_THREADS_RUN_FCST} --exclusive

ush/setup.sh
christinaholtNOAA: I don't think we want any variable representations that need to be expanded in the var_defns.sh file, do we?

At least I am currently working under that assumption in another PR given the current state of that config file.

ush/templates/FV3LAM_wflow.xml
christinaholtNOAA: I think we can assume that this will work for the Slurm-based linux systems. Let's see if we can get some input from @christopherwharrop-noaa on that conjecture.

ush/setup.sh
JeffBeck-NOAA: So long as the variable representations can be expanded (i.e., the required definitions are provided earlier in the file), I would think it's OK.  @gsketefian, thoughts?  I agree that an eventual goal of explicitly defining all fields/variables in the var_defn.sh file would be useful, and less prone to error.

ush/templates/FV3LAM_wflow.xml
JeffBeck-NOAA: OK, I'm wondering if we may want to have a user-defined, Linux-based job scheduler option to handle these templating blocks.  If that's too much work, then a comment somewhere saying this is specifically for Slurm-based Linux systems is probably fine.

christopherwharrop-noaa: I believe @christinaholtNOAA is correct.  That should work.

ush/setup.sh
JeffBeck-NOAA: @christinaholtNOAA, minor typo in the comment block (occurance -> occurrence).

christinaholtNOAA: The reason I am bringing this up is that assuming that variables should not be expanded when the file is sourced means that we can include templates in this file. Not in this particular PR, but for example, if we wanted to set a template for a path to a file on a platform, we can leave it in the var_defns.sh file like this:

```
EXTRN_MDL_SYSBASEDIR_ICS='/scratch1/NCEPDEV/rstprod/com/gfs/prod/gfs.${yyyymmdd}/${hh}/atmos/'
```
And then fill it in when we're ready with run-time variables like this (where data variables are already set):

```
sysdir=$(eval echo ${EXTRN_MDL_SYSBASEDIR_ICS})
```

ush/templates/FV3LAM_wflow.xml
christinaholtNOAA: @christopherwharrop-noaa Is it a fair assumption that Slurm + linux would account for a majority of the installations on unsupported platforms? Is there any good reason to choose PBS on any CSP? Would this Rocoto flag still work appropriately in that case? Can we only support so much flexibility?

christopherwharrop-noaa: This is tricky to do in a portable and generic way across multiple batch system schedulers.  Rocoto offers a tpp (threads per process) option in the `<nodes>` tag which should translate into this without using a `<custom>` tag.  However, then you have to calculate the number of nodes needed, which in turn would require knowing how many cores per node there are, which is specific to a given machine.  This is a hard problem to solve in a clean way given the machinery we have to work with.

christopherwharrop-noaa: @christinaholtNOAA - Yes.  That is a pretty safe assumption.  Slurm appears dominant in that arena.  Moreover, both AWS and Google (not sure about Azure) use Slurm for their HPC clustering solutions in the cloud.

christinaholtNOAA: I think that the `<cores>` and `<native>` flags are the simplest direction to go for the generic Linux + Slurm + Rocoto setup.

We have the configurable options we need to support these: `NCORES_PER_NODE` and `PE_MEMBER01` and `OMP_NUM_THREADS_RUN_FCST`.

ush/setup.sh
JeffBeck-NOAA: The var_defns.sh script was intended to be a comprehensive set of paths/variables that were explicitly defined, which would actually be an argument to not have any variable representations.  The current workflow design appends the cycle/hour strings to avoid needing templating in the var_defns.sh file.  For example, if:

`EXTRN_MDL_SYSBASEDIR_ICS="/scratch1/NCEPDEV/rstprod/com/gfs/prod"`

(which then becomes "sysbasedir" in ush/get_extrn_mdl_file_dir_info.sh), then ${yyyymmdd}/${hh} are used as follows:

`sysdir="$sysbasedir/gfs.${yyyymmdd}/${hh}/atmos"`

@gsketefian, does your pending [PR](https://github.com/NOAA-EMC/regional_workflow/pull/580) change this configuration?

ush/templates/FV3LAM_wflow.xml
JeffBeck-NOAA: OK, that works.  Can you add a comment somewhere that these settings are a Slurm-specific implementation of Linux with Rocoto?  Thanks.

ush/setup.sh
christinaholtNOAA: I understand that is how the current workflow addresses the issue. I am working on new development (potentially conflicting with PR #580) that works under the assumption that no variables will be expanded, and that we can store templates like in my example above in that file.

The specific example doesn't matter for the scope of this PR. What matters here is that we should or should not assume that we expand variables upon sourcing the the var_defns.sh file. To date, we can assume that we do not immediately need to expand any variables since none of the values there ever include any variable references. 

I personally would like to keep it that way so that we can more easily template variables like the one in the example above.

ush/templates/FV3LAM_wflow.xml
christinaholtNOAA: Sure thing. Put it in config_default.sh.

ush/setup.sh
JeffBeck-NOAA: OK. If RUN_CMD_FCST is the only variable that currently needs to be expanded (and I believe it is), then we should avoid doing that, which is what your modification to the code originally proposed.   I'm just wondering if we can expand RUN_CMD_FCST directly within config.sh or when it's sourced to create var_defn.sh.

christinaholtNOAA: Not trying to throw @mkavulich under the bus, but this was his implementation in the release branch. I might need him to defend this one. :)

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch2/BMC/zrtrr/rrfs_ci/autoci/pr/759317963/20211103173604/ufs-srweather-app
Build was Successful
Script /scratch2/BMC/zrtrr/rrfs_ci/autoci/pr/759317963/20211103173604/ufs-srweather-app/regional_workflow/tests/WE2E/end_to_end_tests.sh does not exist in repo
Cannot run WE2E tests
ci-hera-intel-WE

christinaholtNOAA:
@JeffBeck-NOAA @gsketefian @mkavulich Can we move forward on this PR. It would be very helpful to get this set of changes into develop for a couple of different cloud porting activities.

christopherwharrop-noaa:
@JeffBeck-NOAA  @gsketefian @mkavulich - I'd like to emphasize the importance of @christinaholtNOAA 's request to move this along to resolve concerns so this can be merged as soon as possible.  The LINUX target work is crucial and is needed for other projects to move forward.  This is the long pole in the tent keeping other work from getting done.  I am not suggesting we compromise on quality of the review process, but am requesting that this PR get the attention required to bring it to the finish line.  We need this now.

JeffBeck-NOAA:
@christinaholtNOAA @christopherwharrop-noaa, my review is more or less complete.  I was waiting to hear from @mkavulich regarding the expansion of RUN_CMD_FCST directly within config.sh or when it's sourced to create var_defn.sh.

ush/load_modules_run_task.sh
JeffBeck-NOAA: @gsketefian, any comment on this?

ush/launch_FV3LAM_wflow.sh
JeffBeck-NOAA: @gsketefian, any comment on this?

ush/launch_FV3LAM_wflow.sh
JeffBeck-NOAA: @gsketefian, any comments here?

ush/launch_FV3LAM_wflow.sh
JeffBeck-NOAA: @gsketefian, comments?

ush/load_modules_run_task.sh
JeffBeck-NOAA: @gsketefian, comments?

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch2/BMC/zrtrr/rrfs_ci/autoci/pr/759317963/20211109160632/ufs-srweather-app
Build was Successful
Script /scratch2/BMC/zrtrr/rrfs_ci/autoci/pr/759317963/20211109160632/ufs-srweather-app/regional_workflow/tests/WE2E/end_to_end_tests.sh does not exist in repo
Cannot run WE2E tests
ci-hera-intel-WE

ush/setup.sh
```diff
@@ -632,10 +632,20 @@ case $MACHINE in
...
+    ;;
+
+  "*")
+    NCORES_PER_NODE="2" # Need some arbitrary default value to avoid division by zero errors
+
+    print_info_msg "\
+      You are running on an unknown platform! The default value of
+    NCORES_PER_NODE = ${NCORES_PER_NODE} will be used unless you set
+    this variable in your configuration file."
     ;;
```
JeffBeck-NOAA, line 645: @christinaholtNOAA, can you explain why this is necessary? If a default NCORES_PER_NODE is defined for LINUX above, then all supported platforms will have a value, and there will be no division by zero errors since a user can't choose an unsupported platform (they have to choose a value listed in the valid_param_vals.sh file, or the generation of the workflow will fail).

christopherwharrop-noaa: @JeffBeck-NOAA - Almost every project I've ever been involved with entails running on a machine that is not supported.  Those of us doing pioneering work need default values such as this to make life easier.  When we need to modify the workflow infrastructure to support the functionality we need to do our work, having fail safes and warnings such as this helps a lot.  Strictly speaking "LINUX" isn't really a platform so much as a workaround to enable users to run outside the confines of the "supported platforms".  Even when a user chooses LINUX a lot of customization may be necessary to all of the "LINUX" cases scattered throughout the code.  I have a lot to say about that but will reserve those comments for the appropriate discussion forum.

JeffBeck-NOAA: @christinaholtNOAA, what I'm saying is this code will not be used.  If a user defines anything other than a defined platform, the generation of the workflow will fail with:

MACHINE must be set to one of the following:
  "WCOSS_CRAY" "WCOSS_DELL_P3" "HERA" "ORION" "JET" "ODIN" "CHEYENNE" "STAMPEDE" "LINUX"

christopherwharrop-noaa: I take your point, but it is usually considered a bug to have a case statement that does not include the  default `*` section in it to handle unexpected conditions.  The problem is that it absolutely **will** be executed (but not on purpose) if someone adds a new platform to the list, and forgets to add the appropriate cases to all the various places in the code.  In the absence of the default clause, the behavior of this code relies on assumed behavior elsewhere, which is subject to change.  This replaces an undefined, difficult to diagnose, behavior with a defined one.

JeffBeck-NOAA: OK, understand.  Last question, I noticed you changed the NCORES_PER_NODE to be definable for each platform.  What was the goal in doing that?  I can see this would be an option for Jet, since the different partitions have different cores per node, and the LINUX target, but the NOAA HPC platforms are completely fixed.  Thanks.

christopherwharrop-noaa: @JeffBeck-NOAA - The number of cores per node is determined by the hardware of a given machine.  On cloud resources, this would depend on which type of machine instance was chosen for use.  It could be almost anything...  Manual control over `NCORES_PER_NODE` is a hard requirement for making use of the `LINUX` target, which basically just means "an unsupported platform".

JeffBeck-NOAA: @christopherwharrop-noaa, yep, definitely understand when using the LINUX target.  I was just curious why it was made an option for systems where that value is fixed (e.g., Hera, Orion, Stampede, Cheyenne, etc).

christinaholtNOAA: @JeffBeck-NOAA I left the default values here, because as you suggest, the number of cores per node is more or less fixed per machine as a hardware limit. When a configuration setting is made configurable (and not hard-coded) it should be treated as such everywhere so that when a user sets a different value and thinks they are running with that value, they can trust that the system didn't choose something else for them.

Let's use the example of running out of memory on hera. All of a sudden, I want to test whether I can submit a job to user a smaller number of cores per node so that the memory issue can be resolved (this is a bit contrived as I know we have done this for other components, but not necessarily the model *yet*). I go in and reconfigure with a configure file that says I want to use 20 cores per node on hera...I reconfigure and wait the half a day for all my jobs to get to through the queue on Hera to learn that it fails exactly the same way. And that's because my configure file that was supposed to set this for me because it's listed in configure_defaults.sh was silently overwritten by the system and then I throw my laptop out the window because I've lost a half day of work! 

That was a bit dramatic. The general rule: if a setting is listed as configurable for ANY case, it should be universally configurable for all cases with reasonable defaults so that it does not HAVE to be set. This one is no different. Because it needs to be configurable for linux, it HAS TO BE configurable for all platforms so users don't go insane.

JeffBeck-NOAA: @christinaholtNOAA, OK, makes sense.  I can also see how this would be an option for debugging, as you mention.

christinaholtNOAA: As far as setting `NCORES_PER_NODE` for an unnamed machine, I will ask for preference, here. As @christopherwharrop-noaa mentioned, we have to have a catch all because it will inevitably get exercised.

1) Should the script exit if it reached that point?
2) Should I take the default setting out here and move it to an explicit check below the case statement? Should that set the default, or exit?

```
if [ -z $NCORES_PER_NODE ] ; then
 # fail because we don't want to make assumptions 
 # OR 
# set it to 2 because that seems safe? But maybe not on macos?
fi
```

Thoughts? For reference, the existing behavior is to set it to 2 and move on.

ush/setup.sh
JeffBeck-NOAA: I'd like to get input from @mkavulich on this last issue.  If there's a way to resolve the expansion of variables in the config.sh and var_defns.sh scripts without having to use the added $sed command in this PR, we should do it.

ush/setup.sh
JeffBeck-NOAA: @christinaholtNOAA, @christopherwharrop-noaa, I would think 2 would be best for a default value of an unknown system, since it will also work for systems that have more cores per node.  An argument could also be made for 1.

ush/setup.sh
christinaholtNOAA: This is literally the same implementation that is in the release branch. There are hundreds of sed commands sprinkled through this framework. I am unclear why you think this one might be special, or specifically a problem.

ush/setup.sh
christopherwharrop-noaa: I think 1 would probably be the most safe.  But there is no magic default value here that makes sense.  If you wanted to get sophisticated, you might be able to find a platform independent way to query the system and find the number of cores.  But I'm not sure there is a portable way to do that.

christinaholtNOAA: Will all the components run with 1 core on all the platforms, even when parallel run commands are invoked?

christinaholtNOAA: Does it make sense to fail, and let the user set it explicitly? It's probably not helpful if it's using resources in a way that is totally unintended. Especially if you're porting to a new machine that is costing a bunch of money to run these experiments.

christopherwharrop-noaa: It probably is not possible to come up with a "good" default that works for all scenarios.  I say either detect the number of cores automatically (which might be wrong, too, if hyperthreading is enabled) and use that as a default, or fail.  Failing is a lot easier to implement.

christopherwharrop-noaa: This is not well tested:

Linux:
```
[Christopher.W.Harrop@awsnoaa-3 ~]$ grep 'cpu cores' /proc/cpuinfo | uniq
cpu cores	: 4
```

Macos:
```
christopher.w.harrop@harrop-lt ~ % sysctl -a | grep machdep.cpu | grep core_count
machdep.cpu.core_count: 8
```

christinaholtNOAA: This is an interesting solution, but will only get us the core count for the processor where the configuration script is run.

ush/setup.sh
mkavulich: Sorry I haven't weighed in here yet @JeffBeck-NOAA @christinaholtNOAA. 

As Christina said, this is from the original implementation that I made on the release branch and subsequently to develop. The reason for doing this was to allow users to simply reference ${PE_MEMBER01} (the number of MPI tasks used by the model) in any custom way they wish in order to specify their run command, which can vary widely from platform to platform. Additionally, this allowed for a "sane" default value for RUN_CMD_FCST that would work with the majority of machines and settings. If we just had a default number of cores, and/or if we forced users to enter the number of MPI tasks manually for their default run command, there is a really good chance they could get it wrong (i.e. incompatible with the other auto-generated parameters in the workflow), which can cause model crashes and/or poor performance with very little clue as to what the issue is. 

The SED command specifically here is because of a quirk in how setup.sh operates with default values: I didn't want to just place the "RUN_CMD_FCST" variable at the end of config_defaults.sh and assume that the needed variable(s) were set, so I put in a custom bit in setup.sh to always place config_defaults.sh at the end and delete the initial instance of it: since we have the "-u" bash flag active in the workflow, whenever var_defns.sh is sourced by the workflow it will fail if a reference to PE_MEMBER01 or other not-yet-defined variable is referenced, even if that reference is eventually overwritten.

I think not allowing settings in var_defns.sh to reference other variables is a huge limiting factor to how the workflow can function in my opinion. I think this is something we should make work, even if the ultimate implementation ends up different.

ush/setup.sh
christopherwharrop-noaa: That is an excellent point!  But I think it's the only viable default.  We may be splitting hairs here.  The current code already prints a warning.  If we feel that the warning is easily missed and will cause hard-to-find problems, we should fail.

mkavulich: I know my input wasn't requested here specifically but I feel the need to chime in regardless: having had time to think about it and after reading the discussion here, I believe that NCORES_PER_NODE should be a mandatory user-specified variable for custom environments (i.e. for the generic LINUX target, not including NCORES_PER_NODE would quit with a descriptive error message). We have seen this (even on officially supported platforms!) in the past that if NCORES_PER_NODE is set to some incorrect value for that machine, we will see varying levels of poor performance that can be quite pernicious and go unnoticed for a very long time, costing money and resources for no reason. For this and the other reasons mentioned above I think it is reasonable to require users to know and specify this value for their system.

christopherwharrop-noaa: I think that makes a lot of sense, @mkavulich and I fully support your recommendation.  I was employing way too many gymnastics trying to come up with a viable strategy for a default that would work.

christinaholtNOAA: I will go with failure here, if the value isn't set! Thanks for the input, y'all!

ush/setup.sh
JeffBeck-NOAA: @christinaholtNOAA, sed isn't the problem here.  The problem is that the code in this line is required to "delete the first occurrence of" RUN_CMD_FCST "in the var_defns file, and write it again at the end", which we should try to avoid.

@mkavulich, sorry, I'm still having some trouble understanding the reasoning here.  If RUN_CMD_FCST relies on PE_MEMBER01, and PE_MEMBER01 is defined beforehand in setup.sh (it currently is), why is this code required? PE_MEMBER01 depends on LAYOUT_X and LAYOUT_Y which must be defined before the user generates the workflow or it will fail (they are defined automatically if a predefined grid is chosen, but can be overwritten by a user who defines them explicitly).

christinaholtNOAA: @JeffBeck-NOAA The `config_default.sh` file, then the `config.sh` file is sourced at the very beginning of `setup.sh`. The settings in those two files are used throughout the remainder of `setup.sh` to compute derived values of other variables, including `PE_MEMBER01`. There is a chicken/egg problem here. If the `RUN_CMD_FCST` set by the user in the config file relies on `PE_MEMBER01`, it is not yet filled in, when the `config.sh` is sourced.

If we follow conventions like we have already talked about (no values should be left un-expanded in `var_defns.sh`), then we must overwrite the `RUN_CMD_FCST` here so that we can include only one instance (for clarity) that has the appropriate value in it.

JeffBeck-NOAA: @christinaholtNOAA, yes, I understand how the config_defaults/config.sh scripts work.  If we don't want to leave un-expanded variables in the var_defns.sh file, then we can't reference undefined variables in the config_defaults.sh or config.sh scripts.

Currently in config_defaults.sh, we see:

```
# RUN_CMD_FCST:
# The run command for the model forecast step. This will be appended to the end
# of the variable definitions file, so it can reference other variables.
```

Therefore, the original goal was to reference other variables and not expand anything.  We should come up with another implementation to avoid having to replace an already defined field.  I'll approve for now, but a goal of explicit variable definitions needs to be implemented to avoid a sed replacement call.

christinaholtNOAA: I'm curious if it's time to move away from bash configuration files....

JeffBeck-NOAA (approved):

christinaholtNOAA:
I am working on addressing the `NCORES_PER_NODE` solution we discussed. Thanks for approving, but it's not ready to merge.

ush/setup.sh
christinaholtNOAA: Should I add this line below to ensure we don't have un-expanded variable references in the definition?

```
RUN_CMD_FCST=$(eval echo ${RUN_CMD_FCST})
```

That will allow the value that shows up in `var_defns.sh` to be:

```
 NCORES_PER_NODE="40"
 PE_MEMBER01="12"
RUN_CMD_FCST=mpirun -np 12
```

JeffBeck-NOAA: Yeah, I was wondering how it was being expanded, if at all, so this "eval" command looks like it'll do the trick.

christinaholtNOAA: Okay, I'll add that here, too.

christinaholtNOAA:
Okay, I _think_ I've made all the requested changes. Please correct me if I'm wrong. If all is good, I'm happy with a merge as I've tested that the new changes result in the appropriate experiment directory changes (all mods were limited to configuration settings).

ush/load_modules_run_task.sh
gsketefian: Sorry, late to the game.  Removing this is ok since it is outdated.  The new build system does its own thing now with modules (I am not very familiar with how/what it does).

ush/load_modules_run_task.sh
gsketefian: I think @mkavulich put in the default module capability.  Not sure when/why it was commented out.  Ok to take it out since we apparently don't need it.

ush/setup.sh
gsketefian: This actually will not have an effect since `setup.sh` just defines the `setup()` function; it does not contain a script.  So you can't call `setup()` directly from a bash shell (unless you source `setup.sh` first); it has to be called from a script that sources `setup.sh` and which includes the `#!/bin/bash` at the top.  I intentionally omit the `#!/bin/bash` from functions since it is not needed and as a reminder that the file defines a function, not a script.

ush/launch_FV3LAM_wflow.sh
gsketefian: Removing is ok.  I guess the best way to deal with situations like this where one needs to remember what was tried is to create an issue and put all the various attempts there.

gsketefian: Created Issue #[631](https://github.com/NOAA-EMC/regional_workflow/issues/631) so this isn't forgotten.

ush/launch_FV3LAM_wflow.sh
gsketefian: Can remove.  Created Issue #[631](https://github.com/NOAA-EMC/regional_workflow/issues/631) so this isn't forgotten.

ush/launch_FV3LAM_wflow.sh
gsketefian: Can remove.  Created Issue #[631](https://github.com/NOAA-EMC/regional_workflow/issues/631) so this isn't forgotten.

ush/load_modules_run_task.sh
```diff
@@ -147,13 +127,15 @@ jjob_fp="$2"
...
+# Sourcing ufs-srweather-app build env file
 #
 #-----------------------------------------------------------------------
 #
+
+module purge
+
 machine=$(echo_lowercase $MACHINE)
-env_fn="build_${machine}_${COMPILER}.env"
+env_fn=${BUILD_ENV_FN:-"build_${machine}_${COMPILER}.env"}
```
gsketefian, line 138: Sorry again for the late input.  Since BUILD_ENV_FN now appears in config_defaults.sh, it will also appear in var_defns.sh.  If it is not explicitly specified by the user in config.sh, it will be set to its default value (a null string) in var_defns.sh because currently it does not get reset to "build_${machine}_${COMPILER}.env" anywhere.  That should happen in setup.sh, i.e. include this in setup.sh:

BUILD_ENV_FN=${BUILD_ENV_FN:-"build_${machine}_${COMPILER}.env"}

This should be placed somewhere before the line 

cp_vrfy $USHDIR/${EXPT_DEFAULT_CONFIG_FN} ${GLOBAL_VAR_DEFNS_FP}

in setup.sh.  This will cause var_defns.sh to contain the correct name of the environment file.  It will also make things easier in this file since now we don't need the env_fn variable; we can just use BUILD_ENV_FN (which will be available since var_defns.sh is sourced at the top of this file).

ush/launch_FV3LAM_wflow.sh
gsketefian: That sounds good.

But the same comment I have below for BUILD_ENV_FN in load_modules_run_task.sh goes here for WFLOW_ENV_FN, i.e. WFLOW_ENV_FN should be (re)set as necessary in setup.sh, i.e. in setup.sh, include this:
`
WFLOW_EN_FN=${WFLOW_ENV_FN:-"wflow_${machine}.env"}
`
so that it shows the correct value in var_defns.sh.  Then it can be used here instead of needing the new variable env_fn (since var_defns.sh is sourced at the top of launch_FV3LAM_wflow.sh).

Issue #623, opened 2021-11-03
cycl_freq is not adjustable in workflow


chan-hoo:
## Description
'cycl_freq' is fixed to '24:00:00' in the workflow generation script.

Issue #392, opened 2021-01-07
develop: Add user-defined interval date option for experiment generation


JeffBeck-NOAA:
In ush/set_cycle_dates.sh, the experiment directories are defined as:

  date_crnt="${date_start}"
  while [ "${date_crnt}" -le "${date_end}" ]; do
    all_cdates+=( $( printf "%s " ${cycle_hrs[@]/#/${date_crnt}} ) )
    date_crnt=$( date -d "${date_crnt} + **1 days**" +%Y%m%d )
  done

There is interest in having a user-defined interval date value which would replace the "1 days" in the date command with whatever the user would like to use.  For example, every three days could be implemented with ${DATE_INTERVAL}=3 in config.sh, and:

  date_crnt="${date_start}"
  while [ "${date_crnt}" -le "${date_end}" ]; do
    all_cdates+=( $( printf "%s " ${cycle_hrs[@]/#/${date_crnt}} ) )
    date_crnt=$( date -d "${date_crnt} + **${DATE_INTERVAL} days**" +%Y%m%d )
  done

The default ${DATE_INTERVAL} value would be set to one in config_defaults and used if undefined in config.sh.

jwolff-ncar:
In addition, generate_FV3LAM_wflow.sh is also set to be a daily increment
‘cycl_freq’: !!str 24:00:00

We would like the ability to set an increment (in hours) to have more flexibility in this setting. And this should become a user-defined setting that can be passed in through the config.sh.

PR #624, opened 2021-11-03
Make cycl_freq calculated from CYCL_HRS

chan-hoo:
## DESCRIPTION OF CHANGES: 
Make 'cycl_freq' calculated from CYCL_HRS for multiple cycle cases.

## TESTS CONDUCTED: 
- Multiple cycle case for RRFS-CMAQ
- Multiple cycle case for RRFS-CMAQ-DA
- Single cycle case: GST test

## ISSUE: 
Fixes issue mentioned in #623

### DIFF
--- a/ush/generate_FV3LAM_wflow.sh
+++ b/ush/generate_FV3LAM_wflow.sh
@@ -376,7 +376,7 @@ settings="\
   'date_last_cycl': ${DATE_LAST_CYCL}
   'cdate_first_cycl': !datetime ${DATE_FIRST_CYCL}${CYCL_HRS[0]}
   'cycl_hrs': [ $( printf "\'%s\', " "${CYCL_HRS[@]}" ) ]
-  'cycl_freq': !!str 24:00:00
+  'cycl_freq': !!str ${CYCL_INC}:00:00
 #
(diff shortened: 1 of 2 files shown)


JeffBeck-NOAA:
@chan-hoo, this PR is related to [this](https://github.com/NOAA-EMC/regional_workflow/issues/392) issue that I created a while ago.  They're related but separate.  Will your PR work if we implement the change shown in that issue?  I think so?

chan-hoo:
@JeffBeck-NOAA , if CYCL_HRS was intended for the arbitrary number of cycle hours for each day, you are correct. I misunderstood CYCL_HRS. In RRFS-CMAQ-DA, we usually set CYCL_HRS=( "00" "06" "12" "18" ) with the same interval of 6 for multiple days. In this case, we should set cycl_freq=6 again for the offset of cycled runs (https://github.com/chan-hoo/regional_workflow/blob/rrfs_cmaq_da/ush/templates/fv3gfs_aqm_wflow.xml). To remove this kind of duplicated inputs, I let cycl_freq calculated from CYCL_HRS. I'll close this PR and create a new PR to address these issues together soon.

PR #634, opened 2021-11-15
Make cycl_freq adjustable in the configuration file

chan-hoo:
## DESCRIPTION OF CHANGES: 
- The cycle frequency, "cycl_freq", can be set with a new parameter "INCR_CYCL_FREQ" in the configuration file. Its default value is 24 which means cycl_freq="24:00:00".
- The other cycle intervals that are larger than 24 hr can be set in increments of 24 (e.g. 24, 48, 72, ...).

## TESTS CONDUCTED: 
- RRFS-CMAQ 2 day 2 cycle run with cycl_freq=24:00:00
- RRFS-CMAQ 5 day 3 cycle run with cycl_freq=48:00:00
- RRFS-CMAQ-DA 1 day 4 cycle run with cycl_freq=06:00:00
- GST test (1 day 1 cycle) with cycl_freq=24:00:00

## ISSUE: 
- Fixes issue mentioned in #623 
- Fixes the issue mentioned in #392


ush/config_defaults.sh
```diff
@@ -436,6 +436,10 @@ WFLOW_LAUNCH_LOG_FN="log.launch_FV3LAM_wflow"
 # two-digit string representing an integer that is less than or equal to
 # 23, e.g. "00", "03", "12", "23".
 #
+# INCR_CYCL_FREQ:
+# Increment in hours for Cycle Frequency (cycl_freq).
+# Default is 24, which means cycle_freq=24:00:00
+#
 # FCST_LEN_HRS:
```
JeffBeck-NOAA, line 443: @chan-hoo, thanks for restructuring this PR.  Since CYCL_HRS is meant to be for a 24-hr period, how would INCR_CYCL_FREQ work for anything other than a 24-hr increment (24, 48, 72, etc)?  I see you ran some tests with INCR_CYCL_FREQ of 6 hours, so I'm trying to understand how that worked.  What were your CYCL_HRS values when you did that?

chan-hoo: @JeffBeck-NOAA, INCR_CYCL_FREQ is used only in the workflow template. To make other increments larger than 24 hrs work, as you know, we should modify "set_cycle_dates.sh" to get the right all_cdates. In RRFS-CMAQ-DA, CYCL_HRS were 00, 06, 12, 18. In RRFS-CMAQ and RRFS-CMAQ-DA, I didn't meet this case (>24) yet.  Do you want me to do this (>24) in this PR?

chan-hoo: @JeffBeck-NOAA, I'll make the other increments (>24) work in this PR. I am testing it now and it looks working well in RRFS-CMAQ. I'll modify this PR and update the code soon.

JeffBeck-NOAA: @chan-hoo, thanks for making this work for multiple day cadences!  So, CYCL_HRS can be any number of hours between 00 and 23, and INCR_CYCL_FREQ can now be anything greater than 24, but in increments of 24 (24, 48, 72, etc.).  Correct?

ush/setup.sh
```diff
@@ -866,11 +899,12 @@ set_cycle_dates \
...
   date_end="${DATE_LAST_CYCL}" \
   cycle_hrs="${CYCL_HRS_str}" \
+  incr_cycl_freq="${INCR_CYCL_FREQ}" \
   output_varname_all_cdates="ALL_CDATES"
 
 NUM_CYCLES="${#ALL_CDATES[@]}"
 
-if [ $NUM_CYCLES -gt 30 ] ; then
+if [ $NUM_CYCLES -gt 365 ] ; then
   unset ALL_CDATES
```
JeffBeck-NOAA, line 947: This was limited to 30 days to keep the size of ALL_CDATES from getting massively large in the output (var_defns.sh, etc.).  Is there a need to increase it to 365 days?

ush/templates/FV3LAM_wflow.xml
```diff
@@ -260,7 +260,7 @@ MODULES_RUN_TASK_FP script.
 ************************************************************************
 ************************************************************************
 -->
-  <task name="&GET_EXTRN_ICS_TN;" maxtries="{{ maxtries_get_extrn_ics }}">
+  <task name="&GET_EXTRN_ICS_TN;" cycledefs="forecast" maxtries="{{ maxtries_get_extrn_ics }}">
```
JeffBeck-NOAA, line 265: I understand what you did here, and it's probably best to explicitly list that these tasks apply to cycledefs="forecast".  The reason it was working before is because... "Rocoto uses the union of all sets of cycles specified by the `<cycledef>` tags to create the overall cycle pool" and "If `<cycledef>` is not set, the task will be run for every cycle in the general pool of cycles."  @gsketefian do you have a preference for whether we define cycledefs explicitly as "forecast" for these tasks or leave things blank as they were?

ush/config_defaults.sh
chan-hoo: @JeffBeck-NOAA, Yes, correct.

ush/setup.sh
chan-hoo: @JeffBeck-NOAA, I was asked to increase this number in RRFS-CMAQ. Someone wanted to use 45 cycles for his runs. Do we need to specify a small number here? If so, what about 60 or 90? ALL_CDATES is necessary for RRFS-CMAQ, too.

ush/templates/FV3LAM_wflow.xml
chan-hoo: @JeffBeck-NOAA, I agree. This change is not necessary for the current SRW App. However, another 'cycledef' is defined in RRFS-CMAQ. So I put this change here for future expandability. Even in RRFS-CMAQ, the cycledef 'forecast' is not necessary because it ('forecast') means all cycles as you said. I put it for a clear distinction. In case that we don't need this change, what about removing the definition part of the cycledef 'forecast' (lines 132-138), too? @JeffBeck-NOAA @gsketefian, let me know what you think.

ush/setup.sh
JeffBeck-NOAA: @chan-hoo, ahh, OK.  Just some background here, there was concern from an external contributor in a previous PR that the ALL_CDATES array could become too large for real-time or long retro configurations.  I think we can increase the limit to 90 without a problem.

ush/templates/FV3LAM_wflow.xml
JeffBeck-NOAA: @chan-hoo, I'm fine with what you've done.  I think listing things explicitly is a good idea, particularly since there are other cycledefs being added.  So, I think we should keep lines 132-138.  If @gsketefian has no objections, I say we go with what you've done.

JeffBeck-NOAA (approved):

gsketefian:
Making cycl_freq adjustable is necessary, but it might be confusing to do it this way since if you specify it to be say 6 hours, then you have to remember to make CYCL_HRS only a single value, e.g. CYCL_HRS=("00"), not CYCL_HRS=("00" "06" "12" "18").  Otherwise, you might get cycles that you didn't really want, I think.  

I am thinking making the new variable (INCR_CYCL_FREQ) an array might make more sense.  It would be the same size as CYCL_HRS (the workflow would check for that), and each element would specify the interval for each corresponding cycle hour in CYCL_HRS.  I'm still thinking about what the best way is...

JeffBeck-NOAA:
> Making cycl_freq adjustable is necessary, but it might be confusing to do it this way since if you specify it to be say 6 hours, then you have to remember to make CYCL_HRS only a single value, e.g. CYCL_HRS=("00"), not CYCL_HRS=("00" "06" "12" "18"). Otherwise, you might get cycles that you didn't really want, I think.
> 
> I am thinking making the new variable (INCR_CYCL_FREQ) an array might make more sense. It would be the same size as CYCL_HRS (the workflow would check for that), and each element would specify the interval for each corresponding cycle hour in CYCL_HRS. I'm still thinking about what the best way is...

My understanding was that cycl_freq must be at least 24, and then intervals of 24 after that.  Maybe we should add a check to fail if the user tries something else?  If a user wants to run a single cycle, then they should use cycl_freq=24 and CYCLE_HRS=("whatever hour they want to run").  If we want to make it more complicated (having an array of cycl_freq/cycl_hrs), maybe that should be a separate PR.

gsketefian:
> > Making cycl_freq adjustable is necessary, but it might be confusing to do it this way since if you specify it to be say 6 hours, then you have to remember to make CYCL_HRS only a single value, e.g. CYCL_HRS=("00"), not CYCL_HRS=("00" "06" "12" "18"). Otherwise, you might get cycles that you didn't really want, I think.
> > I am thinking making the new variable (INCR_CYCL_FREQ) an array might make more sense. It would be the same size as CYCL_HRS (the workflow would check for that), and each element would specify the interval for each corresponding cycle hour in CYCL_HRS. I'm still thinking about what the best way is...
> 
> My understanding was that cycl_freq must be at least 24, and then intervals of 24 after that. Maybe we should add a check to fail if the user tries something else? If a user wants to run a single cycle, then they should use cycl_freq=24 and CYCLE_HRS=("whatever hour they want to run"). If we want to make it more complicated (having an array of cycl_freq/cycl_hrs), maybe that should be a separate PR.

cycl_freq does not have to be at least 24, at least not in rocoto.  You can specify running on the same date at hours 00, 06, 12, and 18 by setting cycl_freq="06", setting CYCL_HRS=("00"), and specifying the start and end dates to be the same.  But you can also do the same thing using CYCL_HRS=("00" "06" "12" "18"), so it becomes redundant and confusing.  The only part of making `INCR_CYCL_FREQ` and `cycl_freq` an array that might require some trial-and-error is how to generalize the following loop in the xml template to make cycl_freq an array:
```
{% for c in cycl_hrs %}
  <cycledef group="forecast">
  {%- set cdate_first=date_first_cycl ~ c ~ "00" -%}
  {%- set cdate_last=date_last_cycl ~ c ~ "00" -%}
  {{- cdate_first ~ " " ~ cdate_last ~ " " ~ cycl_freq -}}
  </cycledef>
{%- endfor %}
```
I think (I haven't tried; just looked up online) this can be generalized by changing the line containing cycl_freq to:
```
  {{- cdate_first ~ " " ~ cdate_last ~ " " ~ cycl_freq[loop.index] -}}
```
assuming cycl_freq has already been set to an array just as cycl_hrs is (and the number of elements has been checked to be the same in setup.sh).

I'll leave it up to @chan-hoo if he wants to do this.  But I prefer we do it now since previous experience indicates that these things will be forgotten after the PR is in, and over time that approach degrades the quality of the workflow.

ush/templates/FV3LAM_wflow.xml
gsketefian: This is ok with me.

JeffBeck-NOAA:
@gsketefian, yeah, I was just suggesting that maybe we specifically limit cycl_freq to be 24-h intervals and do the rest with cycl_hrs.  I tend to think setting cycl_freq="06" and CYCL_HRS=("00") to run hours 00, 06, 12, and 18, is a bit more confusing than doing it with cycl_freq="24" and CYCL_HRS=("00" "06" "12" "18"), but it's a personal preference.

If INCR_CYCL_FREQ were an array, how would you run the same interval for multiple CYCL_HRS?  Would it be a single value or just repeated?

ush/setup.sh
gsketefian: @chan-hoo Is the `ALL_CDATES` variable used anywhere in the RRFS-CMAQ?  It is currently not used anywhere in the workflow, it just gets recorded in `var_defns.sh`.  If it is not needed in the RRFS-CMAQ, I'm thinking to avoid having a long list in var_defns.sh, we can still "calculate" ALL_CDATES via set_cycle_dates.sh, but instead of writing to var_defns.sh, do one of the following:
1. Write ALL_CDATES to a different file, e.g. all_cdates.txt.
2. Don't write it to a file for now.

gsketefian:
> @gsketefian, yeah, I was just suggesting that maybe we specifically limit cycl_freq to be 24-h intervals and do the rest with cycl_hrs. I tend to think setting cycl_freq="06" and CYCL_HRS=("00") to run hours 00, 06, 12, and 18, is a bit more confusing than doing it with cycl_freq="24" and CYCL_HRS=("00" "06" "12" "18"), but it's a personal preference.
> 
> If INCR_CYCL_FREQ were an array, how would you run the same interval for multiple CYCL_HRS? Would it be a single value or just repeated?

I prefer that way too, but that's actually the main way rocoto assumes you're going to run it -- specify initial and final cdates (including minute) and a cycl_freq, which would be something less than 24 if you're running multiple cycles per day.  

To run the same interval for multiple CYCL_HRS, you can do something crazy like this:
```
CYCL_HRS=("00" "06" "18")
INCR_CYCL_FREQ=("24" "48" "72")
```
Then the "00" cycle would run for every day, the "06" cycle for every other day, and the "18" cycle for every third day.

I'm still thinking about the best (least confusing for the user) way to set it up.

JeffBeck-NOAA:
> > @gsketefian, yeah, I was just suggesting that maybe we specifically limit cycl_freq to be 24-h intervals and do the rest with cycl_hrs. I tend to think setting cycl_freq="06" and CYCL_HRS=("00") to run hours 00, 06, 12, and 18, is a bit more confusing than doing it with cycl_freq="24" and CYCL_HRS=("00" "06" "12" "18"), but it's a personal preference.
> > If INCR_CYCL_FREQ were an array, how would you run the same interval for multiple CYCL_HRS? Would it be a single value or just repeated?
> 
> I prefer that way too, but that's actually the main way rocoto assumes you're going to run it -- specify initial and final cdates (including minute) and a cycl_freq, which would be something less than 24 if you're running multiple cycles per day.
> 
> To run the same interval for multiple CYCL_HRS, you can do something crazy like this:
> 
> ```
> CYCL_HRS=("00" "06" "18")
> INCR_CYCL_FREQ=("24" "48" "72")
> ```
> 
> Then the "00" cycle would run for every day, the "06" cycle for every other day, and the "18" cycle for every third day.
> 
> I'm still thinking about the best (least confusing for the user) way to set it up.

Yeah, I can see how this would get complicated quickly.  How would you run 00, 06, and 18 every day, would it be:

```
 CYCL_HRS=("00" "06" "18")
 INCR_CYCL_FREQ=("24" "24" "24")
 ```
or 
```
 CYCL_HRS=("00" "06" "18")
 INCR_CYCL_FREQ=("00" "00" "00")
 ```

?

gsketefian:
> > > @gsketefian, yeah, I was just suggesting that maybe we specifically limit cycl_freq to be 24-h intervals and do the rest with cycl_hrs. I tend to think setting cycl_freq="06" and CYCL_HRS=("00") to run hours 00, 06, 12, and 18, is a bit more confusing than doing it with cycl_freq="24" and CYCL_HRS=("00" "06" "12" "18"), but it's a personal preference.
> > > If INCR_CYCL_FREQ were an array, how would you run the same interval for multiple CYCL_HRS? Would it be a single value or just repeated?
> > 
> > 
> > I prefer that way too, but that's actually the main way rocoto assumes you're going to run it -- specify initial and final cdates (including minute) and a cycl_freq, which would be something less than 24 if you're running multiple cycles per day.
> > To run the same interval for multiple CYCL_HRS, you can do something crazy like this:
> > ```
> > CYCL_HRS=("00" "06" "18")
> > INCR_CYCL_FREQ=("24" "48" "72")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then the "00" cycle would run for every day, the "06" cycle for every other day, and the "18" cycle for every third day.
> > I'm still thinking about the best (least confusing for the user) way to set it up.
> 
> Yeah, I can see how this would get complicated quickly. How would you run 00, 06, and 18 every day? Would it be:
> 
> ```
>  CYCL_HRS=("00" "06" "18")
>  INCR_CYCL_FREQ=("24" "24" "24")
> ```
> 
> ? I think so. This option might be the most flexible (using an INCR_CYCL_FREQ array), although I'm still uncertain how values of less than "24" would impact cycles. What would this do?
> 
> ```
>  CYCL_HRS=("00" "06" "18")
>  INCR_CYCL_FREQ=("06" "06" "06")
> ```

Maybe this approach:
1. Set the default value of INCR_CYCL_FREQ to ("24"), i.e. one element, set to "24".
2. User may override INCR_CYCL_FREQ with something else in config.sh, e.g. INCR_CYCL_FREQ=("24" "48").
3. In setup.sh:
  a. If INCR_CYCL_FREQ is a scalar or an array with one element, reset it to an array of the same size as CYCL_HRS with all elements having the original value.
  b. If INCR_CYCL_FREQ is an array with more than one element, then:
      i. If the number of elements is not the same as that in CYCL_HRS, print out an error message and exit.
      ii. If the number of elements is the same as in CYCL_HRS, then proceed as expected, i.e. each cycle hour will have a corresponding interval in INCR_CYCL_FREQ.

I don't think there is a need to check whether the values in INCR_CYCL_FREQ are less than or greater than 24 since when rocoto forms the cycledef, it keeps only one copy of each cdate (i.e. it won't run the same date more than one time).  But we should still calculate ALL_CDATES in set_cycle_dates and then check for repeated cdates in the resulting array.  Then, if there are repeated cdates (e.g. because the user specified settings like in your last example), print out an informational message but then weed out the repeated values (using the uniq utility).  That way the user will know that there is a simpler way to specify the dates/cycles, but their settings will still work.

ush/set_cycle_dates.sh
```diff
@@ -104,9 +105,23 @@ End date (date_end) must be at or after start date (date_start):
 #
   all_cdates=()
   date_crnt="${date_start}"
+
```
gsketefian, line 108: Why do we restrict INCR_CYCL_FREQ to be integer number of days?  Why not let the user specify it as "hh:mn:ss"?  Rocoto would just accept that string as is, and here, we can use sed to extract the hours, minutes, and seconds. and add them in the DATE_UTIL command below.

ush/setup.sh
```diff
@@ -853,6 +853,39 @@ does not have this form:
 done
 #
 #-----------------------------------------------------------------------
+# Check cycle increment for cycle frequency (cycl_freq).
+# only if INCR_CYCL_FREQ < 24.
+#-----------------------------------------------------------------------
+#
+if [ "${INCR_CYCL_FREQ}" -lt "24" ] && [ "$i" -gt "1" ]; then
+  cycl_intv="$(( 24/$i ))"
```
gsketefian, line 900: It looks like this assumes that the cycle hours in CYCL_HRS have a constant increment, e.g. CYCL_HRS=("00" "06" "12" "18"), in which case the increment is 6 hours.  But one can also have CYCL_HRS=("00" "06"), in which case there is no uniform increment value (first increment is 6, second increment is 18).  If we use the approach I suggested in my other comments about allowing INCR_CYCL_FREQ to be an array, then we won't need this check because any value for the elements of INCR_CYCL_FREQ would be allowed (since rocoto will automatically weed out repeated values in cycledefs, and we can inform/warn the user about repeat values in set_cycle_dates.sh).

ush/setup.sh
chan-hoo: @gsketefian, In RRFS-CMAQ, ALL_CDATES is used to set another cycledef which excludes the 'at first' cycledef tasks. However, I think there should be different ways to do it. So if you plan to remove ALL_CDATES from the scripts, don't worry about this. I'll be able to find an alternative.

ush/set_cycle_dates.sh
chan-hoo: @gsketefian, INCR_CYCL_FREQ is not an integer number of days but that of hours. cycl_freq="{INCR_CYCL_FREQ}:00:00". Theoretically, yes, we can specify minutes and seconds, but practically we set the cycle frequency to hours like 3h, 6h, 24h. I think we don't need to give users unnecessary options if possible. Have you seen any cases using cycle frequencies in minutes and seconds? If so, we should modify this.

ush/setup.sh
chan-hoo: @gsketefian, This check is only for the cycle frequency is less than 24 hours. For example, in DA runs, we run 12 cycles with cycle_freq=06:00:00 (3 day run). To do this, we should set CYCL_HRS=("00" "06" "12" "18") and INCR_CYCL_FREQ="06". This check prevents typos in CYCL_HRS. In your second example, CYCL_HRS=("00" "06"), their increments are "6" and "18" as you said. However, its INCR_CYCL_FREQ is "24" because you intend "00/1st_day" "06/1st_day" "00/2nd_day" "06/2nd_day". So this example is not the case for this check. As shown in the second example, we can generate non-uniform cycle frequencies with combination of INCR_CYCL_FREQ="24" and non-uniform CYCL_HRS. I don't think we need to make cycle frequencies an array at this moment. This will make users much more confused. With this PR, any integer numbers can be selected between 00 and 24 with 24 hour increment (or more) for CYCL_HRS. But only if the increment of cycle frequencies is less than 24, their intervals should be constant. If CYCL_HRS are not uniform, INCR_CYCL_FREQ should be >=24. This approach can cover all cases we can meet in our runs (hopefully). What do you think of this?

chan-hoo: @gsketefian, if you don't have a comment any more, I'd like to merge this PR as it is. As I mentioned above, this PR works even when interval<24h ,and it also works well in RRFS-CMAQ and RRFS-CMAQ-DA (even intervals: 6h or 3h). If you want to make the interval as an array, I suggest you open a new PR for it. It looks so complicated to me. I'd like to merge this into RRFS-CMAQ as early as possible to make the two regional workflows go forward in parallel.

gsketefian: @chan-hoo Go ahead.  I had some more comments but haven't had time to get back to this, and they're not super important.  Thanks.

PR #639, opened 2021-11-19
Add new DEBUG variable

gsketefian:
## DESCRIPTION OF CHANGES: 
1. Add a new experiment configuration variable named `DEBUG` to enable more in-depth debugging output from workflow scripts.  Set default value of `DEBUG` in `config_defaults.sh` to `"FALSE"`.
2. In experiment generation scripts, change circumstances under which different messages are printed to screen (e.g. when `VERBOSE` is `"TRUE"`, when `DEBUG` is `"TRUE"`, or always).
3. In experiment generation scripts, for clarity add new informational messages and modify some existing ones.
4. In various scripts, change "set -x" to "set +x" to reduce output clutter.  This can be changed back as necessary (e.g. for debugging).

Note that if `DEBUG` is set to `"TRUE"`, `VERBOSE` will get reset to `"TRUE"` if necessary in order to also print out all the `VERBOSE` messages.

## TESTS CONDUCTED: 
Ran the WE2E test `grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2` as-is as well as with modifications to the default values of `VERBOSE` and `DEBUG`, as follows:
1. `grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2` as-is, i.e. using default values `VERBOSE="TRUE"` and `DEBUG="FALSE"`.
2. `grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2` modified with `VERBOSE="FALSE"` (and with default of `DEBUG="FALSE"`).
3. `grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2` modified with `DEBUG="TRUE"` (and with default of `VERBOSE="TRUE"`).
4. `grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2` modified with `DEBUG="TRUE"` and `VERBOSE="FALSE"` (which should get reset to `"TRUE"`).

All tests were successful.  The experiment generation log files (`log.generate_FV3LAM_wflow.sh`) were compared and differed in the expected ways.

## DOCUMENTATION:
Necessary documentation of `DEBUG` is in `config_defaults.sh`.  Created Issue #[640 ](https://github.com/NOAA-EMC/regional_workflow/issues/640)to also update rst documentation.


ush/generate_FV3LAM_wflow.sh
```diff
@@ -152,16 +160,26 @@ WFLOW_XML_FP="$EXPTDIR/${WFLOW_XML_FN}"
...
+
+  ensmem_indx_name="\"\""
+  uscore_ensmem_name="\"\""
+  slash_ensmem_subdir="\"\""
+  if [ "${DO_ENSEMBLE}" = "TRUE" ]; then
+    ensmem_indx_name="mem"
+    uscore_ensmem_name="_mem#${ensmem_indx_name}#"
+    slash_ensmem_subdir="/mem#${ensmem_indx_name}#"
+  fi
+
```
JeffBeck-NOAA, line 181: Does the above block of code dealing with ensemble indexing/sub-directories belong within the WORKFLOW_MANAGER if statement?

ush/check_ruc_lsm.sh
```diff
@@ -20,7 +20,7 @@ function check_ruc_lsm() {
 #
 #-----------------------------------------------------------------------
 #
-  { save_shell_opts; set -u -x; } > /dev/null 2>&1
+  { save_shell_opts; set -u +x; } > /dev/null 2>&1
 #
```
JeffBeck-NOAA, line 24: Do all these -x to +x changes get modified back to -x if DEBUG is set to TRUE?

gsketefian: They don't.  I was thinking about doing that but decided to leave it manual.  I'm on the fence about that.

ush/generate_FV3LAM_wflow.sh
gsketefian: I think yes, since those three local variables (`ensmem_indx_name`, `uscore_ensmem_name`, and `slash_ensmem_subdir`) are only used in constructing the rocoto XML.

JeffBeck-NOAA (approved):

Issue #641, opened 2021-11-22
Some template namelist files need to be updated to use newer ufs weather model


chan-hoo:
## Description
Since the namelist parameters were changed in a newer version of the ufs weather model, some template namelist files should be updated.

PR #642, opened 2021-11-22
Update template namelist files for newer ufs weather model

chan-hoo:
## DESCRIPTION OF CHANGES: 
- The template namelist files, 'input.nml' and 'model_configure', are updated for use of a newer version of the UFS Weather Model.
- The namelist parameter, print_diff_pgr, for the pressure tendency diagnostic is added to 'input.nml' (default = .false.).

## TESTS CONDUCTED: 
WE2E tests on WCOSS dell and Hera:
- grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GSD_SAR

## DEPENDENCIES:
- ufs-community/ufs-srweather-app/pull/193

## ISSUE: 
Fixes issue mentioned in #641 

## CONTRIBUTORS: 
@BenjaminBlake-NOAA, @MatthewPyle-NOAA

### DIFF
--- a/ush/templates/input.nml.FV3
+++ b/ush/templates/input.nml.FV3
@@ -70,6 +70,7 @@
     do_schmidt = .true.
     do_vort_damp = .true.
     dwind_2d = .false.
+    dz_min = 2
     external_eta = .true.
     external_ic = .true.
     fill = .true.
@@ -108,6 +109,7 @@
     p_fac = 0.1
     phys_hydrostatic = .false.
     print_freq = 6
+    psm_bc = 1
     range_warn = .true.
     read_increment = .false.
     regional = .true.
@@ -180,6 +182,7 @@
     oz_phys_2015 = .true.
     pdfcld = .false.
     pre_rad = .false.
+    print_diff_pgr = .false.
     prslrd0 = 0.0
     random_clds = .false.
     redrag = .true.
--- a/ush/templates/input.nml.FV3
+++ b/ush/templates/input.nml.FV3
@@ -10,7 +10,6 @@
 &atmos_model_nml
     chksum_debug = .false.
     dycore_only = .false.
-    fdiag = 1
 /
 
 &cires_ugwp_nml
@@ -174,7 +173,6 @@
     ltaerosol = .true.
     lwhtr = .true.
     n_var_lndp = 0
-    ncld = 5
     nsradar_reset = 3600
     nst_anl = .true.
     nstf_name = 2,1,0,0,0
--- a/ush/templates/model_configure
+++ b/ush/templates/model_configure
@@ -1,5 +1,4 @@
-total_member:            1
-PE_MEMBER01:             {{ PE_MEMBER01 }}
+print_esmf:              {{ print_esmf }}
 start_year:              {{ start_year }}
 start_month:             {{ start_month }}
 start_day:               {{ start_day }}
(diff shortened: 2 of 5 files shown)


ush/templates/model_configure
```diff
@@ -14,44 +15,32 @@ cpl:                     {{ cpl }}
...
+# Output frequency in hours.
 #
 # nsout:
 # Output frequency in time steps (positive values override "nfhout" and 
 # "nfhout_hf").
 #
-nfhout:     {{ nfhout }}
-nfhmax_hf:  60
-nfhout_hf:  1
+output_fh:  {{ nfhout }}  -1
```
JeffBeck-NOAA: For "nsout", could you change the comment to say:

... (positive values override "output_fh")

For line 43, {{ nfhout }} should really be changed to {{ output_fh }}.  However, this will require changes to code and comments in ush/create_model_configure_file.sh.  Would you be willing to do that in this PR?  A single find/replace for the old entries should work.

Also, scripts/exregional_run_post.sh has a comment about nfhout and nfhout_hf which should be updated.

ush/templates/input.nml.FV3
```diff
@@ -109,6 +109,7 @@
     p_fac = 0.1
     phys_hydrostatic = .false.
     print_freq = 6
+    psm_bc = 1
     range_warn = .true.
```
JeffBeck-NOAA, line 113: What do the two above changes do?  Just curious.

JeffBeck-NOAA (commented):
Thanks @chan-hoo for getting things up-to-date with ufs-weather-model!  Just a few comments to address.

ush/templates/input.nml.FV3
chan-hoo: @JeffBeck-NOAA, 'dz_min': minimum thickness depth to enforce monotonicity of height to prevent blowup (artificial limiter, default=2), 'psm_bc': option to use original BCs (=0) or zero-gradient BCs (=1) to reconstruct interface u/v with the parabolic spline mehtod for the advection of height (default=0). Do you think it would be better to change "psm_bc = 1" to "0" as a default value in the template? @BenjaminBlake-NOAA, does "psm_bc=1" make results better than "0"?

chan-hoo: https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/issues/64: it is said that "psm_bc=1" is better for stability.

BenjaminBlake-NOAA: @chan-hoo  That is correct, we started using psm_bc = 1 in our model runs on 8/9 to (hopefully!) improve model stability.

ush/templates/model_configure
chan-hoo: @JeffBeck-NOAA, modified as you suggested. and completed the above WE2E tests successfully.

ush/templates/input.nml.FV3
JeffBeck-NOAA: @chan-hoo, @BenjaminBlake-NOAA, does the psm_bc=1 option work similarly to Tom Black's blending?  Is it essentially blending u/v values into the compute domain?

BenjaminBlake-NOAA: @JeffBeck-NOAA  I think so.  Here is some more detailed information (originally from Fanglin) about psm_bc:

Based on the investigation of the model instability issue, a new scheme to reconstruct the interface horizontal wind for the advection of geopotential height was proposed. In the original FV3 dynamic core, the interface horizontal winds are calculated by solving a tridiagonal system of equations which are formulated by 1-D remapping algorithm based in the Parabolic Spline Method (PSM) with high-order boundary conditions (BCs). In the new scheme, the high-order BCs are replaced with zero-gradient BCs. The impact of this new method was investigated for GFSv16, the regional applications, as well as idealized mountain ridge tests. The results are very encouraging as it can solve the model instability issues not only for GFS v16 but also the high-resolution regional applications. In addition, it has little impact on the forecast performances and the numerical solution of idealized mountain right tests (Fig. 2).  This new scheme was approved by both GFDL and EMC developers and merged in the UFS FV3 dynamic core.

JeffBeck-NOAA (approved):

PR #645, opened 2021-12-06
Toggle on/off workflow tasks

gsketefian:
## DESCRIPTION OF CHANGES:
1. Add capability to turn on/off those workflow tasks that currently do not have this capability.  These tasks are:
  a. Getting the external model files for creating initial conditions (`GET_EXTRN_ICS_TN`).
  b. Getting the external model files for creating boundary conditions (`GET_EXTRN_LBCS_TN`).
  c. Regridding/interpolating data from the external model IC files to obtain IC fields on the native grid (`MAKE_ICS_TN`).
  d. Regridding/interpolating data from the external model LBC files to obtain LBC fields on the native grid (`MAKE_LBCS_TN`).
  e. Running the forecast (`RUN_FCST`).
Note that any remaining tasks in the workflow already have the capability to be toggled on/off.

2. Add a WE2E test (named `deactivate_tasks`) to test the ability to turn off tasks.  In this test, all the tasks except `MAKE_GRID_TN`, `MAKE_OROG_TN`, and `MAKE_SFC_CLIMO_TN` are turned off (the latter 3 are not turned off because there is already a separate WE2E for turning those three off).

## TESTS CONDUCTED: 
Ran the new WE2E test (`deactivate_tasks`) successfully on Hera.

## DOCUMENTATION:
Necessary documentation has been included in `config_defaults.sh`.  Created Issue #646 to also update rst documentation.


ush/setup.sh
```diff
@@ -2608,10 +2608,10 @@ read -r -d '' str_to_insert << EOM
 #-----------------------------------------------------------------------
 #-----------------------------------------------------------------------
 # Section 1:
-# This section is a copy of the default workflow/experiment configura-
-# tion file config_defaults.sh in the shell scripts directory USHDIR ex-
-# cept that variable values have been updated to those set by the setup
-# script (setup.sh).
+# This section is a copy of the default experiment configuration file 
+# (${EXPT_DEFAULT_CONFIG_FN}) in the shell scripts directory specified by USHDIR 
+# except that variable values have been updated to those for the experiment
```
JeffBeck-NOAA, line 2613: I like having the descriptive "config_defaults.sh" mentioned here, but if it's a formatting/syntax thing based on other comments, then I understand.

JeffBeck-NOAA (approved):

ush/setup.sh
gsketefian: In the actual variable definitions file that will be created in the experiment directory, it will say "config_defaults.sh", not EXPT_DEFAULT_CONFIG_FN.  This is just to make var_defns.sh contain the correct name of the default configuration file in case we ever decide to change the name of that file.  Let me know if I'm not being clear...

JeffBeck-NOAA (approved):

PR #650, opened 2021-12-13
Enhance ability to use template variables

gsketefian:
## DESCRIPTION OF CHANGES:
1. Enhance ability to use template variables in the experiment configuration file (either in the default configuration file `config_defaults.sh` or the user configuration file `config.sh`).
2. Modify WE2E test system to include test of template variable use.
3. Fix bugs.

### Notes on template variables:
A template variable (or simply a template) is an experiment variable that contains in its definition a reference to another variable(s).  The referenced variable can be another experiment variable (i.e. one that is defined in `var_defns.sh`), or it can be a local variable (i.e. one that is not defined in `var_defns.sh` but in the script or function that sources `var_defns.sh` and uses the template).  For example, a template named `TEMPL_VAR` my be defined in `config_defaults.sh` or `config.sh` as

`TEMPL_VAR='cd ${some_dir}'`

where `some_dir` may be an experiment variable or a local variable.  `TEMPL_VAR` can then be evaluated using bash's `eval` built-in command in a script or function that first sources `var_defns.sh` and, if necessary, defines `some_dir`.  Note that single quotes must be used on the right-hand side to avoid expansion of `${some_dir}` before run time (i.e. when `eval` is called on `TEMPL_VAR`).  For details, see the documentation added in PR #[198](https://github.com/ufs-community/ufs-srweather-app/pull/198).

### Changes to WE2E tests:
* Modify the WE2E test configuration file `config.deactivate_tasks.sh` to include template variables.  `deactivate_tasks` now serves as a test of both deactivating tasks and of using template variables.
* Add `template_vars` as an alternate test name for `deactivate_tasks` (by creating a symlink named `config.template_vars.sh` that points to `config.deactivate_tasks.sh`).

### Bug fixes:
* In `get_WE2Etest_names_subdirs_descs.sh`, change the variable `alt_test_subdirs` to `alt_test_names` at a single location.
* In `setup.sh`, set `BUILD_ENV_FN` and `WFLOW_ENV_FN` (instead of in `load_modules_run_task.sh` and `launch_FV3LAM_wflow.sh`, respectively).  This way, these variables will have the correct values in `var_defns.sh`.
* In `get_expts_status.sh`, fix the way `homerrfs` is calculated.

## TESTS CONDUCTED: 
The WE2E tests `grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2` and `template_vars` were run on Hera.  Both completed successfully.

## DOCUMENTATION:
Documentation is added to the User's Guide via PR #[198](https://github.com/ufs-community/ufs-srweather-app/pull/198) into the ufs-srweather-app repo.

## Dependencies:
PR #[198](https://github.com/ufs-community/ufs-srweather-app/pull/198) for the documentation.

## CONTRIBUTORS:
@christinaholtNOAA and @mkavulich brought up the issue of templates as part of PR #[617](https://github.com/NOAA-EMC/regional_workflow/pull/617).


ush/bash_utils/get_bash_file_contents.sh
```diff
@@ -0,0 +1,119 @@
+#
```
christinaholtNOAA, line 1: Please consider using appropriate indentation level for all comments in the code. It's impossible to follow what the code is doing and the level of indentation when all comments start from the first column.

This request applies to all the new functions.

ush/config_defaults.sh
```diff
@@ -1717,9 +1719,9 @@ NUM_ENS_MEMBERS="1"
 #
 #-----------------------------------------------------------------------
 #
-DO_SHUM="false"
-DO_SPPT="false"
-DO_SKEB="false"
+DO_SHUM="FALSE"
```
christinaholtNOAA, line 1759: Why are these case changes necessary?

ush/constants.sh
```diff
@@ -15,3 +15,18 @@ degs_per_radian=$( bc -l <<< "360.0/(2.0*$pi_geom)" )
...
+# a variable.  This is the dollar sign.  For example, if the string is
+#
+#   "Hello, my name is $NAME."
+#
+# then the dollar sign indicates that $NAME is a reference to the contents
+# of the variable NAME.
+#
+#-----------------------------------------------------------------------
+#
+VARVALUE_REF_CHAR="\$"
```
christinaholtNOAA: Is this level of engineering needed? Is it not clearer to just use these two characters where we want to use a template? I can't think of any reasons why this level of indirection adds value.

ush/launch_FV3LAM_wflow.sh
```diff
@@ -106,14 +106,12 @@ expt_name="${EXPT_SUBDIR}"
 #
 #-----------------------------------------------------------------------
 #
-machine=$(echo_lowercase $MACHINE)
-env_fn=${WFLOW_ENV_FN:-"wflow_${machine}.env"}
-env_fp="${SR_WX_APP_TOP_DIR}/env/${env_fn}"
+env_fp="${SR_WX_APP_TOP_DIR}/env/${WFLOW_ENV_FN}"
```
christinaholtNOAA, line 119: Does this not change the behavior of launching the workflow? It hasn't been set in the config_defaults file, unless I missed something.

christinaholtNOAA (changes requested):
I've left a couple of questions below as I was looking through to see where the overlap might be between our work to move toward templated config options.

tests/WE2E/test_configs/wflow_features/config.deactivate_tasks.sh
```diff
@@ -14,13 +30,33 @@ PREEXISTING_DIR_METHOD="rename"
...
+# variables.  Note that during this test, these templates aren't actually 
+# expanded/used (using bash's "eval" built-in command) anywhere in the 
+# scripts.  They are included here simply included for demonstration 
+# purposes.  If all goes well, the experiment generation system should 
+# place these templates at the end of the experiment's variable definitions 
+# file.
+#
+RUN_CMD_UTILS='cd \$yyyymmdd'
+RUN_CMD_FCST='mpirun -np \${PE_MEMBER01}'
+RUN_CMD_POST='\$( echo hello \$yyyymmdd )'
```
christinaholtNOAA: It's unclear to me how this test will pass/fail. It certainly wouldn't run for a WE2E test where commands are actually submitted. Is it a smoke test to make sure a var_defns.sh file comes out? Are the contents of that file checked for the right answer?

ush/bash_utils/get_bash_file_contents.sh
gsketefian: Ok, will do.

ush/config_defaults.sh
gsketefian: They're not necessary, I just wanted the settings to be consistent throughout this file.

ush/constants.sh
gsketefian: Well when I first started the PR I wasn't sure how many places I was going to need this character.  Currently, it shows up in a couple of places -- in setup.sh and check_var_valid_value.sh.  I'm ok going either way with this.

ush/launch_FV3LAM_wflow.sh
gsketefian: As part of this PR, `WFLOW_ENV_FN` (as well as `BUILD_ENV_FN`) gets set in `setup.sh`.  Since it's a variable that appears in `config_defaults.sh`, it will automatically be recorded in `var_defns.sh`.  In order for it to have the correct value in `var_defns.sh`, it has to get set in `setup.sh` before the generation of `var_defns.sh`.  This is the same approach as with several other experiment variables.  It shouldn't change the launching of the workflow (I did some tests already and all seems ok).

ush/bash_utils/get_bash_file_contents.sh
gsketefian: Ok, done.  Also did `check_expt_config_vars.sh`.  Feedback welcome!

ush/bash_utils/get_bash_file_contents.sh
```diff
@@ -0,0 +1,71 @@
...
+  #    (a backslash) with the following line.
+  # 2) Remove any leading and trailing whitespace.
+  #
+  # Note that these two actions are automatically performed by the "read"
+  # utility in the while-loop below.
+  #
+  contents=""
+  while read crnt_line; do
+    contents="${contents}${crnt_line}
+"
```
christinaholtNOAA, line 54: This line seems like it needs a line continuation, maybe?

ush/constants.sh
christinaholtNOAA: I think it's much clearer just to use the standard bash syntax where it's needed, instead of building this level of re-direction around something that most folks understand in bash.

ush/launch_FV3LAM_wflow.sh
christinaholtNOAA: Ahhh. I couldn't find where it was set in my last pass. I needed to expand the diffs in setup.sh to find it. Thanks for the clarification! :)

ush/setup.sh
```diff
@@ -1566,20 +1581,20 @@ fi
...
-# Note that the CCPP physics suite defintion file does not have a cor-
-# responding template file because it does not contain any values that
-# need to be replaced according to the experiment/workflow configura-
-# tion.  If using CCPP, this file simply needs to be copied over from 
-# its location in the forecast model's directory structure to the ex-
-# periment directory.
+# (i.e. dummy) values for various parameters.  The experiment generation 
+# scripts copy these templates to appropriate locations in the experiment 
+# directory (either the top of the experiment directory or one of the 
+# cycle subdirectories) and replace the placeholders in these copies by
```
christinaholtNOAA: Experiment generation no longer creates cycle subdirectories, if I remember correctly. This comment should be updated to reflect that.

ush/setup.sh
```diff
@@ -1566,20 +1581,20 @@ fi
...
-# tion.  If using CCPP, this file simply needs to be copied over from 
-# its location in the forecast model's directory structure to the ex-
-# periment directory.
+# (i.e. dummy) values for various parameters.  The experiment generation 
+# scripts copy these templates to appropriate locations in the experiment 
+# directory (either the top of the experiment directory or one of the 
+# cycle subdirectories) and replace the placeholders in these copies by 
+# actual values specified in the experiment configuration file (or derived 
+# from such values).  The scripts then use the resulting "actual" files 
+# as inputs to the forecast model.
```
christinaholtNOAA: All of them are "actual" files -- templates and run-time ready ones. Maybe "The scripts move the configured files from the experiment directory to the run directory at run time."

ush/setup.sh
```diff
@@ -1566,20 +1581,20 @@ fi
...
+#
+# Note that the CCPP physics suite defintion file does not have a 
+# corresponding template file because it does not contain any values 
+# that need to be replaced according to the experiment configuration.  
+# If using CCPP, this file simply needs to be copied over from its 
+# location in the forecast model's directory structure to the experiment 
+# directory.
```
christinaholtNOAA: There's no longer an option NOT to use CCPP, so this logic comment is no longer clear.

ush/setup.sh
```diff
@@ -2560,179 +2573,135 @@ set_thompson_mp_fix_files \
...
+# because variable references will not be expanded when the variable 
+# definitions file is sourced; they will be expanded only when the "eval" 
+# built-in command is used to evaluate the contents of the template.
+# For this reason, use of single quotes is the recommended way of defining 
+# a template in the experiment configuration file.  However, in case 
+# a template is enclosed in double quotes, any varaible references will
+# be expanded/evaluated when the variable definitions file is sourced.
+# In this case, unexpected behavior or failure may be encountered if the
+# referenced variables are not already defined.  To cover this scenario,
+# we place the templates all at the end of the variable definitions file.
```
christinaholtNOAA: This seems like too much logic for what functionality we need. Currently there are very few variable paths that need to be expanded at run time instead of at configuration time. By that, I mean that to date, we (rightly) don't propagate variable references that should be expanded to the var_defns.sh file.

If we limit users to only variables we _expect_ to contain templates (run-time specific ones), then we don't have to handle the situation where ALL variables could potentially contain a template. This logic becomes unnecessary and a simplified understanding and implementation.

For example, we document that templates are only supported for RUN_COMMAND_* and EXTRN_MODEL_PATH (or whatever) directory variables. Then, we know that we only have the expectations to expand THOSE variables at run-time. Nothing else needs to change.

ush/setup.sh
```diff
@@ -2804,272 +2763,301 @@ Setting its value in the variable definitions file to an empty string."
...
-LOGDIR="$LOGDIR"
-CYCLE_BASEDIR="${CYCLE_BASEDIR}"
-GRID_DIR="${GRID_DIR}"
-OROG_DIR="${OROG_DIR}"
-SFC_CLIMO_DIR="${SFC_CLIMO_DIR}"
-
-NDIGITS_ENSMEM_NAMES="${NDIGITS_ENSMEM_NAMES}"
-ENSMEM_NAMES=( $( printf "\"%s\" " "${ENSMEM_NAMES[@]}" ))
-FV3_NML_ENSMEM_FPS=( $( printf "\"%s\" " "${FV3_NML_ENSMEM_FPS[@]}" ))
+SR_WX_APP_TOP_DIR=\"${SR_WX_APP_TOP_DIR}\"
```
christinaholtNOAA: Do we _need_ the quotes to show up in config file here? All of these variables would be expanded to include their final results in at this point, if we consider limiting the ability to use templates to a handful of variables where they are needed.

christinaholtNOAA (changes requested):
Gerard, I've left several additional comments and questions. I am also now considering that it could be more useful to hold off on introducing support for templating _all_ config variables. The implementation of this feature seems to add a level of complexity that may hinder the configuration of the system more than help us. When I'd mentioned needing to use a few templated configuration variables in a previous PR, I was already working with a solution that had MUCH less overhead for the problem at hand when separating out machine-dependent files. In general, we do not need all config variables to be templated, but it could be helpful for a select few to work that way so that we can, for instance, supply the templated name of a cycle-dependent input file. 

I think it could be helpful to consider a transition to python and YAML sooner rather than later for managing configuration, and we could mitigate the need for jumping through so many hoops to make bash work in a way python is already great with.

ush/bash_utils/get_bash_file_contents.sh
gsketefian: It is actually correct; I tested it.  It is just adding a newline after appending `${crnt_line}` to `contents`.  Adding a line continuation will cause the newline to be skipped, and all newlines will be missing from `contents` (making it one super long line, which isn't what we want).  An alternate and maybe clearer way to append a new line to `contents` is 

`printf -v "contents" "${contents}${crnt_line}\n"`

This is a bit slower since it calls `printf` for every line that's read in.  Either way is ok with me.

ush/launch_FV3LAM_wflow.sh
gsketefian: Sure!

ush/constants.sh
gsketefian: Removed variable `VARVALUE_REF_CHAR`.

ush/setup.sh
gsketefian: Modified this paragraph.  Hopefully addresses all three concerns here.

tests/WE2E/test_configs/wflow_features/config.deactivate_tasks.sh
gsketefian: Yes, it is a test to make sure the templated variables are placed towards the end of var_defns.sh.  I actually wrote some other code to make sure that the commands in these variables can be evaluated and executed properly (in as much as it makes sense to evaluate them), and they do.  I did not have a good case of how these would be used in an actual application, so I left it at that.  If you have a use case we can include as a WE2E test, that would be great.

ush/setup.sh
gsketefian: I generally just double-quote the value (note: double-quoting doesn't prevent expansion of the value; single-quotes do) since it seems to be the recommended approach in online forums.  If a variable contains whitespace (I know most of these don't), the double quotes will keep the whitespace (and anything following) as part of the value (otherwise, will likely get an error).  Also, it looks nicer with syntax coloring, at least in vi/vim.

ush/setup.sh
gsketefian: The reason for writing templated variables at the end of the var_defns.sh file is to make sure that any variables that they reference are already defined above in that file.  This was the problem @mkavulich was running into when templating RUN_CMD_FCST, so he had to put in code to specifically remove the line originally defining RUN_CMD_FCST (at which point PE_MEMBER01 was not yet defined) and then insert it later in var_defns.sh.  

This code is just identifying which variables are templated.  It would be pretty straightforward to add an if-statement in the loop below to allow only a select few variables to be templated.

gsketefian:
> Gerard, I've left several additional comments and questions. I am also now considering that it could be more useful to hold off on introducing support for templating _all_ config variables. The implementation of this feature seems to add a level of complexity that may hinder the configuration of the system more than help us. When I'd mentioned needing to use a few templated configuration variables in a previous PR, I was already working with a solution that had MUCH less overhead for the problem at hand when separating out machine-dependent files. In general, we do not need all config variables to be templated, but it could be helpful for a select few to work that way so that we can, for instance, supply the templated name of a cycle-dependent input file.
> 
> I think it could be helpful to consider a transition to python and YAML sooner rather than later for managing configuration, and we could mitigate the need for jumping through so many hoops to make bash work in a way python is already great with.

As I mentioned in one of my replies, it would at this point be quite straightforward to add an if-statement in the loop in setup.sh where templated variables are identified to allow only a certain subset of variables to be templated.  And yes, I'm ok with using a python solution for configuration, although it would be good to talk about that with the rest of the DTC team so they're all on-board.  But if we could get this PR in, that would be great.

JeffBeck-NOAA (approved):

tests/WE2E/test_configs/wflow_features/config.deactivate_tasks.sh
christinaholtNOAA: How is this test run?

ush/setup.sh
christinaholtNOAA: The templated variables that we need may not have their values defined in the var_defns.sh file, though. They may need to be expanded at run time, another reason that it's important to limit specific ones to be templated, and handle those specifically in the run scripts. Its irrelevant to the generation of the var_defns.sh file whether the variable is a template or not when we can assume that only specific variables are templates, and those are handled appropriately at run time.

The ones I am thinking of currently are templates for location of cycle-specific files on disk. Also, run commands for the platforms that use "nprocs" for their run command. In those cases, nprocs is computed in the run script, and could be different for each task, so should only be filled in there. PE_MEMBER01 is another one that could be filled in in var_defns.sh for the RUN_CMD, but wouldn't necessarily need to be if we expand all RUN_CMDS at run time (because of the aforementioned treatment of nprocs). 

Are there others that need to be templated that you can think of given that the current system strictly does not allow template variables?

ush/setup.sh
```diff
@@ -2804,272 +2760,301 @@ Setting its value in the variable definitions file to an empty string."
...
+# the variable that it refers to has already been defined (assuming the
+# referenced variable is one of the experiment variables).
+#
+    dollar_or_null=$( printf "%s" "${var_value}" | \
+                      $SED -n -r -e "s/[^\$]*(\$).*/\1/p" )
+    if [ -z "${dollar_or_null}" ]; then
+      printf -v "var_defns_notempl" "${var_defns_notempl}${var_defn}\n"
+    else
+      printf -v "var_defns_templ" "${var_defns_templ}${var_defn}\n"
+    fi
```
christinaholtNOAA: This is not necessary because the whole point is that we don't want to fill in templates in this file!

ush/setup.sh
christinaholtNOAA: By doing this explicitly, none of these values CAN be templates. They WILL be expanded once this file is sourced, even if they aren't expanded when the file is written, and we definitely want to preserve templates until run time when they should be filled in.


For example, if you wanted log files to be written to a date-specific location, which is something that is useful for real-time runs, we would want to preserve that template and would need to track down where to remove double quotes for expansion. 

Instead, we probably should remove these double quotes completely unless a given argument is allowed to contain white space. It looks like the case where there are multiple items with white space already DO NOT have double quotes and are handled appropriately as strings aside from double quotes.

Please consider removing all the double quotes here for clarity in how we are able to introduce templated variable values in the future.

JeffBeck-NOAA:
> > Gerard, I've left several additional comments and questions. I am also now considering that it could be more useful to hold off on introducing support for templating _all_ config variables. The implementation of this feature seems to add a level of complexity that may hinder the configuration of the system more than help us. When I'd mentioned needing to use a few templated configuration variables in a previous PR, I was already working with a solution that had MUCH less overhead for the problem at hand when separating out machine-dependent files. In general, we do not need all config variables to be templated, but it could be helpful for a select few to work that way so that we can, for instance, supply the templated name of a cycle-dependent input file.
> > I think it could be helpful to consider a transition to python and YAML sooner rather than later for managing configuration, and we could mitigate the need for jumping through so many hoops to make bash work in a way python is already great with.
> 
> As I mentioned in one of my replies, it would at this point be quite straightforward to add an if-statement in the loop in setup.sh where templated variables are identified to allow only a certain subset of variables to be templated. And yes, I'm ok with using a python solution for configuration, although it would be good to talk about that with the rest of the DTC team so they're all on-board. But if we could get this PR in, that would be great.

@gsketefian @christinaholtNOAA, I reviewed and approved this PR, since it looks like Christina's concerns have been resolved.  Transitioning to a purely Python and YAML-based configuration system would make sense, since we've already set that up for other components of the workflow.  However, given the rather large footprint of that work, a future PR makes sense.  Since variable templating is desired soon, I think we should get this PR merged and then work on a larger migration to Python afterward.

christinaholtNOAA (changes requested):
Gerard, I feel as though this solution is still not quite capturing what we need for being able to use a template as a value in a variable. 

It will be helpful if we can specify a variable that contains more generic templated information that is not expanded until _run time_, and to limit those variables to a small handful that we know might need to contain processor counts or date-time information, e.g.. This can be as simple as not using double quotes to expand variables by default, and to just have them show up in var_defns.sh as they are set by the scripts and configure files.

It seems that the solution here is attempting to engineer the capability to specify a variable that reference any other known variable and expand it in the var_defns.sh file. Bash already does this for us by default if we are careful about the order in which variables are specified, and we are. If we aren't careful, we have the `set -u` feature of bash that will readily call us out to fix our mistake.

To date, the setup.sh file expands everything before writing any of it to the var_defns.sh file, leaving no references to other variables, which is great and as it should be.  I was proposing in a comment of a previous PR that If we assume that everything will be expanded before writing the variables to var_defns.sh file, then we can write those values with single quotes and nothing changes, _except_ our ability to write a template value to that file and preserve it until run time when we want to fill it in with task- and configuration-specific information.

Here's an example of what I mean:
```
HOMErrfs="/some/path/on/disk"
EXTRN_MDL_SYSBASEDIR_ICS="/some/path/to/${YYYYMMDD}/${HH}"
```
Gets expanded to an unuseful path when sourced:
```
HOMErrfs="/some/path/on/disk"
EXTRN_MDL_SYSBASEDIR_ICS="/some/path/to//"
```
However, if we use single quotes for everything, we lose nothing from how we were already doing things and we gain the ability to template some fields
```
HOMErrfs='/some/path/on/disk
EXTRN_MDL_SYSBASEDIR_ICS='/some/path/to/${YYYYMMDD}/${HH}'
```

We can then use `$(eval echo $EXTRN_MDL_SYSBASEDIR_ICS)` once we've pinned down what YYYYMMDD and HH should be and have a nice path like `/some/path/to/20210103/12`

gsketefian:
@christinaholtNOAA After skimming through your comments, I think we're close to a solution.  The single vs. double quotes was something I was grappling with as well when making this PR.  Let me read through them more carefully and get back to you.  Thanks.

christinaholtNOAA:
@JeffBeck-NOAA We are working at NOAA GSL toward a plan for merging capabilities for RRFS DA into develop. That will inevitably include a transition to Python + YAML and we need to be in regular coordination about that effort. Please do not move forward on that without first discussing details with our group.

JeffBeck-NOAA:
> @JeffBeck-NOAA We are working at NOAA GSL toward a plan for merging capabilities for RRFS DA into develop. That will inevitably include a transition to Python + YAML and we need to be in regular coordination about that effort. Please do not move forward on that without first discussing details with our group.

@christinaholtNOAA, since you've contributed similar features in the past, I was going to refer to you for that development, so no worries.

gsketefian:
@christinaholtNOAA Since the comments/replies above are getting convoluted, I'll reply to you here.  Let me know if I miss anything.

> It seems that the solution here is attempting to engineer the capability to specify a variable that reference any other known variable and expand it in the var_defns.sh file. 

Yes, that was my intent.  I was trying to allow for the case @mkavulich was struggling with that used double-quotes, i.e.

`RUN_CMD_FCST="mpirun -n \${PE_MEMBER01}"`

and also to allow the referenced variable (`PE_MEMBER01` here) to be expanded either in `var_defns.sh` or at run time.  That's just to cover as many use cases as possible.  But I'm ok with narrowing things down so that variable expansion happens only at runtime (by requiring that templates always use single quotes).  Then it would be unnecessary to move the template variables to the end of `var_defns.sh`.

The other thing I was assuming that differs from your assumption is that the template is evaluated just using `eval` without `echo`, e.g. `eval $RUN_CMD_FCST`.  In this case, the `$` in front of `PE_MEMBER01` needs to be escaped even if using single-quotes, e.g.

`RUN_CMD_FCST='mpirun -n \${PE_MEMBER01}'`

That is why the double-quoting on line ~2879 of `setup.sh` still worked with templates for me (i.e. variables did not get expanded).  But if we use `eval echo $RUN_CMD_FCST`, then there's no need to escape the `$`, but then we do have to remove the double quotes from line ~2879.  I think not escaping the `$` is simpler for the user, so let's go with that.  I'll remove those double-quotes.

> Are there others that need to be templated that you can think of given that the current system strictly does not allow template variables?

Nope!

Finally, it would be good to have a WE2E test that properly tests the templating capability.  If you have a simple case we can use, that would be great.  I cannot think of one without artificially changing one of the J-jobs or ex-scripts.

I do have a couple of additional questions:
1. How do you propose to determine whether a templated variable is one of the ones allowed to be a template?
2. Should we be concerned that a (malicious) 3rd party may give a user a `config.sh` file to run that contains something like `RUN_CMD_FCST=' rm -rf *`?  Or do we just leave that up to the user to catch?

I'll work on the code adjustments to this PR early next week.  Please let me know if I missed anything.  Thanks.

ush/bash_utils/check_var_valid_value.sh
```diff
@@ -108,24 +108,34 @@ The value specified in ${var_name} is not supported:
...
+# contain a literal string.  In this case, we check whether it is equal 
+# to one of the elements of the array valid_var_values.  If not, we 
+# print out an error message and exit the calling script.
 #
 #-----------------------------------------------------------------------
 #
-  is_element_of "valid_var_values" "${var_value}" || { \
-    valid_var_values_str=$(printf "\"%s\" " "${valid_var_values[@]}");
-    print_err_msg_exit "\
+  if [[ "${var_value}" != *"\$"* ]]; then
```
christinaholtNOAA: Would the `\` be required in a template variable? I don't think you need it when using single quotes.

christinaholtNOAA (approved):
@gsketefian Thanks for making those changes. I think this is a super useful feature that will help out with how we can achieve machine independence. I appreciate the willingness to help and make it what we need for this project.

I wanted to respond to your earlier questions.

As for tests with this, I think they can be covered by the existing tests in the next couple of iterations of PRs since we haven't actually implemented a use case for one of these templates just yet.

> How do you propose to determine whether a templated variable is one of the ones allowed to be a template? Have an "allowed" list to compare against?

I think it would be acceptable to just document them for now. It's a relatively obscure change that was made to existing behavior and since we're treating all variables the same, I don't foresee any issues that might arise from compatibility with older config files and new scripts. If it's a feature that starts causing us trouble in the future, we could build a check to make sure that templates aren't specified where we don't expect them. 

> Should we be concerned that a (malicious) 3rd party may give a user a config.sh file to run that contains something like RUN_CMD_FCST=' rm -rf *? Or do we just leave that up to the user to catch?

We should always be concerned with malicious settings for config items, but as it is, whether we set a template and use "eval" to run it, or if we just run it as a command, we will have the same result. We do not currently have check in place for run commands that could contain malicious attacks, so I can't see how we'd need to require that for this particular feature.

ush/bash_utils/check_var_valid_value.sh
gsketefian: I want to test this but Hera is down today, and I'm reluctant to rebuilt on another platform...  I will if I have time, otherwise tomorrow.

gsketefian: Per your suggestion, we're not using the `\` before the `$` in the templates any more (and requiring that templates always be in single quotes).  Then the question is what needs to change here.  Turns out all of the following work to detect a `$` in a single-quoted template:

```
if [[ "${var_value}" != *"\$"* ]]; then ...
if [[ "${var_value}" != *"$"* ]]; then ...
if [[ "${var_value}" != *'$'* ]]; then ...
```

I think the last one is the clearest, so I'm going with that.

gsketefian:
I am ready to merge this one (along with PR #[198](https://github.com/ufs-community/ufs-srweather-app/pull/198)) if there are no more comments.

christinaholtNOAA:
I can take another look.

christinaholtNOAA:
LGTM!

Issue #656, opened 2021-12-20
Update divergence damping parameters


chan-hoo:
## Description
Since the precipitation rate issue is reduced with the new values of the divergence damping 'nord' and 'd4_bg', they should be updated in the template 'input.nml.FV3'.

PR #657, opened 2021-12-20
Update divergence damping parameters

chan-hoo:
## DESCRIPTION OF CHANGES: 
The values of the divergence damping parameters 'nord' and 'd4_bg' are updated in the template 'input.nml.FV3' to reduce the precipitation rate issue:
- nord: 2 -> 3
- d4_bg: 0.15 -> 0.12
- In the field tables, 'ice_nc', 'rain_nc', and 'sgs_tke' are changed from '1.e30' to '0.0'.

## TESTS CONDUCTED: 
1. WE2E tests on WCOSS Dell:
- grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GSD_SAR

2. GST test on WCOSS Dell and Cray, Hera, and Orion

## DEPENDENCIES:
- ufs-community/ufs-srweather-app/pull/200

## ISSUE: 
Fixes issue mentioned in #656 

## CONTRIBUTORS: 
@BenjaminBlake-NOAA


JeffBeck-NOAA:
@BenjaminBlake-NOAA, @chan-hoo, are these changes going to go into EMC's version of the RRFS?  Has there been discussion with GSL on this topic?

chan-hoo:
@BenjaminBlake-NOAA, @JacobCarley-NOAA, can you answer @JeffBeck-NOAA's questions?

BenjaminBlake-NOAA:
@JeffBeck-NOAA @chan-hoo  Yes, these changes were added to our real-time RRFS parallels on 11/30/2021.  The new nord/d4_bg values yield improvements in the precip/reflectivity stats.  I was not included in any discussions with GSL, but @JacobCarley-NOAA or @ericaligo-NOAA will hopefully know more!

ericaligo-NOAA:
We also see improved storm structure with these new nord/d4_bg values.

JeffBeck-NOAA (approved):

modulefiles/tasks/hera/make_grid.local
```diff
@@ -1,5 +1,6 @@
 #%Module
 module use -a /contrib/miniconda3/modulefiles
 module load miniconda3
+module load netcdf
```
JeffBeck-NOAA: @chan-hoo, what was the reason for adding the netcdf module load to all the local module files?  This should be handled when the workflow sources the build-based environment file in 'ush/load_modules_run_task.sh':

```
machine=$(echo_lowercase $MACHINE)
env_fn=${BUILD_ENV_FN:-"build_${machine}_${COMPILER}.env"}
env_fp="${SR_WX_APP_TOP_DIR}/env/${env_fn}"
source "${env_fp}" || print_err_msg_exit "\
Sourcing platform- and compiler-specific environment file (env_fp) for the 
workflow task specified by task_name failed:
  task_name = \"${task_name}\"
  env_fp = \"${env_fp}\"" 
```
For example, for Hera, this will load the build_hera_intel.env file in ufs-srweather-app which contains the "module load netcdf/4.7.4" line.

chan-hoo: @JeffBeck-NOAA, you're right! In my test run, make_grid, make_orog, run_fcst failed with the error "ncdump was not found". However, based on your comment, I found out that it was caused by a missing command line. Since I introduced 'srw_common' for this PR (https://github.com/ufs-community/ufs-srweather-app/pull/200), I should add "module use "${SR_WX_APP_TOP_DIR}/env"" before the source command. I'll modify the PR soon. Thank you for your comment!!!

ush/load_modules_run_task.sh
```diff
@@ -137,6 +137,7 @@ module purge
 machine=$(echo_lowercase $MACHINE)
 env_fn=${BUILD_ENV_FN:-"build_${machine}_${COMPILER}.env"}
 env_fp="${SR_WX_APP_TOP_DIR}/env/${env_fn}"
+module use "${SR_WX_APP_TOP_DIR}/env"
```
christinaholtNOAA, line 139: I'm  not sure this line should go here.

chan-hoo: @christinaholtNOAA, this PR is linked to https://github.com/ufs-community/ufs-srweather-app/pull/200 where srw_common is introduced. Without this line, srw_common is not loaded properly.

christinaholtNOAA: I was saying that this line should go into the environment files themselves since a user would need it to load the build environment manually. Putting this line here instead of in the env files breaks that capability.

chan-hoo: @christinaholtNOAA, I agree. I tried to do it, but it was hard to put the path to 'env' into the environment file. Do you have an idea?

christinaholtNOAA: There are a couple of options, I think. 

1) Since these env files are bash files (not module files), then you could get the path using something like 
`$( readlink -f "${BASH_SOURCE[0]}" )` for doing a "module use" in the env files themselves. That won't work for macos because it doesn't support readlink, but I don't think macos is supporting module file use anyway. Maybe @mkavulich could weigh in on that requirement.

2) Transition all the env files to actual module files so that it behaves the same way as it does in ufs-weather-model. Users will need to do a `module use...module load` to load the env module files instead of sourcing the bash files. Those are standard requirements for loading actual modules, instead of `moudle use...source`, which is non-standard.

chan-hoo: @christinaholtNOAA, Thank you for your suggestions! The second one looks more reasonable to me. I'll try it soon.

PR #660, opened 2021-12-22
Updates to verification configs for RRFS_baseline branch 

jwolff-ncar:
## DESCRIPTION OF CHANGES: 
Updates for changes that were tested on disk but not yet committed to the RRFS_baseline branch
Changes to several METplus conf files:
1. Derive 10-m instantaneous wind from 10-m U and V components
2. Include several thresholds for categorical wind verification
3. Add censor threshold to visibility to set everything greater than 10 miles to 10 miles to have consistent behavior between the model and observations.
4. For REFC, use censor threshold to set bad reflectivity data to a consistent -999 and data <-20 dBZ to a minimum of -20 dBZ to have consistent behavior between the model and observations.
5. For RETOP:
- convert MRMS data from km to kft; set -3(km)=-9.84252kft->-9999 (no coverage - what we call missing); -1(km)=-3.28084kft->-16.4042kft (missing - what we call no data):to match fcst
- convert fcst data from m to kft; set -5000m=-16.4042 (no data)
- change cnt_thresh >0 to compute for anything above ground.
6. Discussed the cnt_logic for REFC and RETOP and decided to proceed with UNION.

## TESTS CONDUCTED: 
These changes were used for all cases included in the RRFS baseline testing and resulting verification plots were examined for appropriateness.

## DOCUMENTATION:
Still need to add verification documentation!

## CONTRIBUTORS (optional): 
@michelleharrold


ush/templates/parm/metplus/PointStat_conus_sfc.conf
```diff
@@ -135,26 +135,31 @@ BOTH_VAR5_NAME = VGRD
 BOTH_VAR5_LEVELS = Z10
 BOTH_VAR5_THRESH = >=2.572 ;; m/s or 5kts
 
-BOTH_VAR6_NAME = PRMSL
-BOTH_VAR6_LEVELS = Z0
+BOTH_VAR6_NAME = WIND
+BOTH_VAR6_LEVELS = Z10
+BOTH_VAR6_THRESH = >=2.572, <5.144, >=5.144, >=10.288, >=15.433 ;; m/s or 5, 10, 20, 30kts
+BOTH_VAR6_OPTIONS = GRIB2_pdt = 0; ;; derive instantaneous 10-m wind from U/V components, overriding max 10-m wind
```
michelleharrold, line 141: This isn't a problem, but I just realized that we typically threshold >=2.572 to avoid the lowest winds -- for the <5.144 threshold, do we want >=2.572 && <5.144?

jwolff-ncar: Good suggestion. We didn't run that way for our baseline. Should we still check it in that way to the branch or just do that when we push it to develop?

michelleharrold: I think doing it when it's push to develop is totally okay, especially since it wasn't run like that for the baseline.

michelleharrold (approved):
I think everything looks in order!

ush/templates/parm/metplus/PointStat_conus_sfc.conf
JeffBeck-NOAA: Making an explicit note here to remind myself of this change:

When opening this PR to develop, change "<5.144" to " >=2.572 && <5.144"

Approving this PR now.

JeffBeck-NOAA (approved):

Issue #664, opened 2022-01-11
The workflow of RRFS-CMAQ needs to be merged into the regional workflow


chan-hoo:
## Description
Since RRFS-CMAQ, based on the UFS SRW App, has been established for air quality modeling successfully, it should be merged into the authoritative repository.

PR #665, opened 2022-01-11
Merge RRFS-CMAQ into authoritative repo

chan-hoo:
## DESCRIPTION OF CHANGES: 
RRFS-CMAQ for air quality modeling is ready to be merged into the authoritative repo.

## QUICK START GUIDE:
1. For air quality modeling (RRFS-CMAQ):
- https://github.com/ufs-community/ufs-srweather-app/wiki/Air-Quality-Modeling

2. For FV3-LAM: same as before

## TESTS CONDUCTED: 
1. Air quality modeling:
- RRFS-CMAQ: non-DA (free forecast), warm start, 2 day 2 cycles, RRFS-CONUS 13km on hera and wcoss_dell_p3
- RRFS-CMAQ-DA: DA with GSI and JEDI, warm start, 1 day 4 cycles, RRFS-CONUS 13km on hera

2. WE2E test for FV3-LAM on hera and wcoss_dell_p3:
- MET_ensemble_verification

## DEPENDENCIES:
- ufs-community/ufs-srweather-app/pull/203

## ISSUE: 
Fixes issue mentioned in #664


jobs/JREGIONAL_ADD_AQM_ICS
```diff
@@ -0,0 +1,116 @@
...
+#
+#-----------------------------------------------------------------------
+#
+EXTRN_MDL_FILES_DIR="${CYCLE_DIR}/AQM/tmp_AQICS"
+#
+#-----------------------------------------------------------------------
+#
+# Create the directory specific to the current forecast (whose starting
+# date and time is specified in CDATE) in which to store the external 
+# model files.  Then change location to that directory.
```
christinaholtNOAA: Is this comment correct and placed correctly? Maybe it should go before the definition of `EXTRN_MDL_FILES_DIR` since no additional date info shows up below?

jobs/JREGIONAL_RUN_CHEM_ANAL
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
```
christinaholtNOAA: Could we please adopt a standard for SRW that the abbreviation of "analysis" can only be "ANL" or "anl" (depending on the standard for capitalization when used)? It would be nice to standardize such that we can grep for a single string.

jobs/JREGIONAL_RUN_GSI_ANAL
```diff
@@ -0,0 +1,101 @@
+#!/bin/bash
+
+#
+#-----------------------------------------------------------------------
+#
+# This J-JOB script runs the GSI analysis 
+#      for the FV3-LAM model
```
christinaholtNOAA: Can this become a single line consistent with the style of the other comments?

jobs/JREGIONAL_RUN_GSI_ANAL
```diff
@@ -0,0 +1,101 @@
...
+#
+# Call the ex-script for this J-job and pass to it the necessary varia-
+# bles.
+#
+#-----------------------------------------------------------------------
+#
+$SCRIPTSDIR/exregional_run_gsi_anal.sh \
+             CYCLE_DIR="${CYCLE_DIR}" \
+             ANALWORKDIR="${ANALWORKDIR}" \
+             CYCLE_ROOT="${CYCLE_ROOT}" || \
```
christinaholtNOAA: A question for @gsketefian, do these variables need to be passed since they are already globals?

And for @chan-hoo, you are passing many duplicates of information here. Can we reduce the duplicate information that is need? For example, you pass CYCLE_DIR, CYCLE, and CYCLE_ROOT, but CYCLE_DIR is a combination of CYCLE_ROOT and CYCLE_DIR. This question applies to both the set of env variables provided to the job script, and those passed along to the ex-script.

jobs/JREGIONAL_RUN_NEXUS
```diff
@@ -0,0 +1,103 @@
...
+# Call the ex-script for this J-job and pass to it the necessary varia-
+# bles.
+#
+#-----------------------------------------------------------------------
+#
+#EXTRN_MDL_FNS_str="( "$( printf "\"%s\" " "${EXTRN_MDL_FNS[@]}" )")"
+
+$SCRIPTSDIR/exregional_run_nexus.sh \
+  CYCLE_DATE="${CDATE}" \
+  CYCLE_DIR="${CYCLE_DIR}" || \
```
christinaholtNOAA: NEXUS_DIR is created above, but it is not passed to the ex-script. There is duplicate information in the ex-script to define, create, and cd to that directory. Please move all that to the this job script and pass the resulting path to the ex-script to reduce the duplication.

jobs/JREGIONAL_RUN_POST_STAT
```diff
@@ -0,0 +1,100 @@
+#!/bin/bash
```
christinaholtNOAA, line 1: The name of this job (and likely ex-script and rocoto task) is unclear in its distinction from RUN_POST. How does it differ from JREGIONAL_RUN_POST and can that job script be utilized to do the function of this one without the duplication of an entire script?

modulefiles/tasks/hera/add_aqm_lbcs.local
```diff
@@ -0,0 +1,10 @@
+#%Module#####################################################
+## Module file for add_aqm_lbcs task.
+#############################################################
+
+module purge
+
+module load intel/18.0.5.274  impi/2018.0.4
+module load netcdf/4.7.0
+module load nco
+
```
christinaholtNOAA, line 6: Why are the two module files different for the ics and lbcs tasks?

modulefiles/tasks/hera/run_chem_anal.local
```diff
@@ -0,0 +1,22 @@
...
+set JEDI_OPT /scratch1/NCEPDEV/jcsda/jedipara/opt/modules
+setenv JEDI_OPT /scratch1/NCEPDEV/jcsda/jedipara/opt/modules
+module use -a $JEDI_OPT/modulefiles/core
+module load jedi/intel-impi/2020.2
+
+
+module load nco
+
+module use -a /scratch2/NCEPDEV/marineda/Jong.Kim/save/modulefiles/
+module load anaconda/3.15.1
```
christinaholtNOAA: How is anaconda used here? This is probably a dependency on a personal installation that is high risk. It might be a good idea to make this a dependency on a small subset of packages, if possible. I can help with that if you can send me a link to the external repo where most of the python scripts live.

The single python script that is included in this PR has no need for any non-standard packages.

modulefiles/tasks/hera/run_chem_anal.local
```diff
@@ -0,0 +1,22 @@
...
+
+set COMP ifort
+set COMP_MP mpfort
+set COMP_MPI mpiifort
+
+set C_COMP icc
+set C_COMP_MP mpcc
+
+set JEDI_OPT /scratch1/NCEPDEV/jcsda/jedipara/opt/modules
+setenv JEDI_OPT /scratch1/NCEPDEV/jcsda/jedipara/opt/modules
```
christinaholtNOAA: Should this not use the $JEDI_OPT variable to reduce duplication?

modulefiles/tasks/wcoss_dell_p3/add_aqm_lbcs.local
```diff
@@ -0,0 +1,9 @@
+#%Module#####################################################
+## Module file for add_aqm_lbcs task.
+#############################################################
+
+module purge
+
+module load ips/18.0.1.163
+module load HPSS/5.0.2.5
+module load NCO/4.7.0
```
christinaholtNOAA, line 8: This file is different from its ICS counterpart. Is that intended?

scripts/exregional_add_aqm_ics.sh
```diff
@@ -0,0 +1,215 @@
...
+#
+#-----------------------------------------------------------------------
+#
+# For debugging purposes, print out values of arguments passed to this
+# script.  Note that these will be printed out only if VERBOSE is set to
+# TRUE.
+#
+#-----------------------------------------------------------------------
+#
+#print_input_args valid_args
```
christinaholtNOAA: Please remove commented code.

ush/templates/fv3gfs_aqm_wflow.xml
```diff
@@ -0,0 +1,1260 @@
...
+    &RSRV_DEFAULT;
+    <command>&LOAD_MODULES_RUN_TASK_FP; "&ADD_AQM_ICS_TN;" "&JOBSDIR;/JREGIONAL_ADD_AQM_ICS"</command>
+    <nodes>{{ nnodes_add_aqm_ics }}:ppn={{ ppn_add_aqm_ics }}</nodes>
+    <walltime>{{ wtime_add_aqm_ics }}</walltime>
+    <nodesize>&NCORES_PER_NODE;</nodesize>
+    <jobname>&ADD_AQM_EXTRN_ICS_TN;</jobname>
+    <join><cyclestr>&LOGDIR;/&ADD_AQM_EXTRN_ICS_TN;_@Y@m@d@H.log</cyclestr></join>
+
+    <envar><name>GLOBAL_VAR_DEFNS_FP</name><value>&GLOBAL_VAR_DEFNS_FP;</value></envar>
+    <envar><name>PREV_CYCLE_DIR</name><value>&RESTART_CYCLE_DIR;</value></envar>
```
christinaholtNOAA: Given the name PREV_CYCLE_DIR, would this not need a cyclestr with an offset?

scripts/exregional_add_aqm_ics.sh
```diff
@@ -0,0 +1,215 @@
...
+#
+#-----------------------------------------------------------------------
+#
+# Create work directory
+#
+#-----------------------------------------------------------------------
+#
+workdir="${CYCLE_DIR}/AQM/tmp_AQICS"
+mkdir_vrfy -p "${workdir}"
+cd_vrfy ${workdir}
```
christinaholtNOAA, line 100: Defining and creating working directories should be done at the jobs list.

modulefiles/tasks/hera/add_aqm_ics.local
```diff
@@ -0,0 +1,8 @@
+#%Module#####################################################
+## Module file for get_extrn_aqm task.
+#############################################################
+
+module purge
+
+module load hpss
+module load intel netcdf nco
```
christinaholtNOAA: Each module should go on a separate line here and elsewhere (if that exists elsewhere)

scripts/exregional_add_aqm_ics.sh
```diff
@@ -0,0 +1,215 @@
...
+ncks -O -d lev,0,0 tmp1.nc tmp1_ptop.nc
+
+${ncap_cmd} -s 'lev=lev+1' tmp1.nc tmp1_pfull.nc
+
+ncrcat -O -o tmp1.nc tmp1_ptop.nc tmp1_pfull.nc
+
+ncks --fix_rec_dmn lev -C -O -o tmp2.nc tmp1.nc
+ 
+ncks -A -C -x -v lon,lat,lev tmp2.nc ${wrk_ic_file}
+
```
christinaholtNOAA, line 184: Will ncks failures exit this script with non-zero exit status? There is not much error handling through this section.

scripts/exregional_add_aqm_lbcs.sh
```diff
@@ -0,0 +1,278 @@
...
+    fhr=$( printf "%03d" "${hr}" )
+    if [ -r ${CYCLE_DIR}/INPUT/gfs_bndy.tile7.${fhr}.nc ]; then
+        ncks -A ${CHEM_BOUNDARY_CONDITION_FILE} ${CYCLE_DIR}/INPUT/gfs_bndy.tile7.${fhr}.nc
+    fi
+  done
+
+  print_info_msg "
+========================================================================
+Successfully added chemical LBCs !!!
+========================================================================"
```
christinaholtNOAA, line 143: Is this guaranteed if we made it here? No checks are done in the above logic.

scripts/exregional_add_aqm_lbcs.sh
```diff
@@ -0,0 +1,278 @@
...
+
+  ext_lbcs_file=${AQM_LBCS_FILES}
+  ext_lbcs_file=${ext_lbcs_file//<MM>/${mm}}
+
+  CHEM_BOUNDARY_CONDITION_FILE=${ext_lbcs_file}
+
+  FULL_CHEMICAL_BOUNDARY_FILE=${AQM_LBCS_DIR}/${CHEM_BOUNDARY_CONDITION_FILE}
+  if [ -f ${FULL_CHEMICAL_BOUNDARY_FILE} ]; then
+    #Copy the boundary condition file to the current location
+    cp ${FULL_CHEMICAL_BOUNDARY_FILE} .
```
christinaholtNOAA: cp_vrfy is standard when copying files.

scripts/exregional_add_aqm_lbcs.sh
```diff
@@ -0,0 +1,278 @@
...
+#
+#-----------------------------------------------------------------------
+#
+# Add chemical LBCS
+#
+#-----------------------------------------------------------------------
+#
+
+yyyymmdd="${PDY:0:8}"
+mm="${PDY:4:2}"
```
christinaholtNOAA: Most other scripts use CDATE to get this information. Can we use that variable for consistency with other tasks?

scripts/exregional_add_aqm_lbcs.sh
```diff
@@ -0,0 +1,278 @@
...
+  NUMTS="$(( ${FCST_LEN_HRS} / ${LBC_SPEC_INTVL_HRS} + 1 ))"
+
+cat > gefs2lbc-nemsio.ini <<EOF
+&control
+ tstepdiff=${GEFS_CYC_DIFF}
+ dtstep=${LBC_SPEC_INTVL_HRS}
+ bndname='aothrj','aecj','aorgcj','asoil','numacc','numcor'
+ mofile='${AQM_GEFS_DIR}/$yyyymmdd/${AQM_GEFS_CYC}/gfs.t00z.atmf','.nemsio'
+ lbcfile='${CYCLE_DIR}/INPUT/gfs_bndy.tile7.','.nc'
+ topofile='${OROG_DIR}/${CRES}_oro_data.tile7.halo4.nc'
```
christinaholtNOAA, line 169: Are these environment variables meant to be expanded here? Should double quotes be used for clarity if that is the  case? (I'm assuming that if they are expanded here then the script does what it needs to do). 

If they aren't expanded here (standard for the single quotes), then where do they get expanded and what are the requirements on setting these templates? Including that documentation and caveat would be very helpful.

scripts/exregional_run_chem_anal.sh
```diff
@@ -0,0 +1,379 @@
...
+#-----------------------------------------------------------------------
+#
+# Create JEDI subdirectory
+#
+#-----------------------------------------------------------------------
+#
+print_info_msg "$VERBOSE" "
+Creating JEDI subdirectory..."
+
+mkdir_vrfy -p ${CYCLE_DIR}/JEDI/Data
```
christinaholtNOAA: This directory structure (at lest through the JEDI directory) should have been defined at the Jobs level.

scripts/exregional_run_chem_anal.sh
```diff
@@ -0,0 +1,379 @@
...
+ln_vrfy -sf $JEDI_DIR/build/fv3-jedi/test/Data/fieldsets ${CYCLE_DIR}/JEDI/Data/fieldsets
+ln_vrfy -sf $JEDI_DIR/build/fv3-jedi/test/Data/fv3files ${CYCLE_DIR}/JEDI/Data/fv3files
+# FV3 namelist
+ln_vrfy -sf $FV3_NML_FP ${CYCLE_DIR}/JEDI/Data/input.nml
+# FV3 input dir (grid info, etc.)
+mkdir_vrfy -p ${CYCLE_DIR}/JEDI/INPUT
+
+print_info_msg "$VERBOSE" "
+Creating links in the JEDI/INPUT subdirectory of the current cycle's run di-
+rectory to the grid and (filtered) orography files ..."
```
christinaholtNOAA: Including what that directory is here would be a helpful log message.

scripts/exregional_run_chem_anal.sh
```diff
@@ -0,0 +1,379 @@
...
+
+cd_vrfy ${CYCLE_DIR}/JEDI/INPUT
```
christinaholtNOAA: This path contains duplicated information from line 121.

scripts/exregional_run_chem_anal.sh
```diff
@@ -0,0 +1,379 @@
...
+#
+#-----------------------------------------------------------------------
+#
+# create output directories for the bump, analysis and hofx files
+#
+#-----------------------------------------------------------------------
+#
+mkdir_vrfy -p ${CYCLE_DIR}/JEDI/Data/hofx
+mkdir_vrfy -p ${CYCLE_DIR}/JEDI/Data/analysis
+mkdir_vrfy -p ${CYCLE_DIR}/JEDI/Data/bump
```
christinaholtNOAA: Again, this directory structure should be relative -- defined by a variable passed from the jobs level.

Also, it would be helpful to set up the entire directory structure at the same time. It seems like the structure is pretty complex compared to other tasks in the regional_workflow, so keeping it together might clarify what the structure is in an easier to follow way.

scripts/exregional_run_chem_anal.sh
```diff
@@ -0,0 +1,379 @@
...
+      if [ -r ${fv_tracer_file} ]; then
+        print_info_msg "
+  Tracer file found: \"${fv_tracer_file}\""
+      else
+        print_err_msg_exit "\
+  No suitable tracer restart file found."
+      fi
+    fi
+  fi
+fi
```
christinaholtNOAA: Is this a duplicate from a script run earlier in the workflow? Can it come up with a different answer at this point, and is that a desired outcome? Should the answer from the previous execution be propagated to this task?

scripts/exregional_run_chem_anal.sh
```diff
@@ -0,0 +1,379 @@
...
+${USHDIR}/gen_JEDI_yaml.py -i $TEMPLATEDIR -o ${CYCLE_DIR}/JEDI/ -c ${CDATE} -y $YAMLS
+#
+#-----------------------------------------------------------------------
+#
+# Set and export variables.
+#
+#-----------------------------------------------------------------------
+#
+export OMP_NUM_THREADS=1
+export OMP_STACKSIZE=1024m
```
christinaholtNOAA: This sort of task should be handled near the top of the script with the machine case statement.

scripts/exregional_run_fcst.sh
```diff
@@ -502,6 +508,53 @@ if [ ${WRITE_DOPOST} = "TRUE" ]; then
...
+  AQM_RC_IN_FN="aqm.rc"
+  AQM_RC_IN_FP="${run_dir}/${AQM_RC_IN_FN}"
+  cp_vrfy ${AQM_RC_FP} ${AQM_RC_IN_FP}
+
+  YYYY=${CDATE:0:4}
+  MM=${CDATE:4:2}
+  DD=${CDATE:6:2}
+  HH=${CDATE:8:2}
+  YYYYMMDD=${CDATE:0:8}
+  YYMMDD=${CDATE:2:6}
```
christinaholtNOAA: These types of variables are typically lower case in all other tasks. Also, can these definitions be moved closer to the top and out of the if statement to simplify logic? There is another definition of these same variables (lower case) in the post processing section below. Also, can you remove the unused variables? It seems like only YYMMDD and YYYYMMDD are used.

scripts/exregional_run_fcst.sh
```diff
@@ -502,6 +508,53 @@ if [ ${WRITE_DOPOST} = "TRUE" ]; then
...
+  AQM_RC_IN_FP = \"${AQM_RC_IN_FP}\""
+
+    set_file_param "${AQM_RC_IN_FP}" "init_concentrations" "${init_concentrations}"
+    set_file_param "${AQM_RC_IN_FP}" "aqm_config_dir" "${AQM_CONFIG_DIR%/}"
+    set_file_param "${AQM_RC_IN_FP}" "aqm_bio_dir" "${AQM_BIO_DIR%/}"
+    set_file_param "${AQM_RC_IN_FP}" "aqm_bio_file" "${AQM_BIO_FILE}"
+    set_file_param "${AQM_RC_IN_FP}" "aqm_fire_dir" "${AQM_FIRE_DIR%/}"
+    set_file_param "${AQM_RC_IN_FP}" "YYYYMMDD" "${YYYYMMDD}"
+    set_file_param "${AQM_RC_IN_FP}" "aqm_fire_file" "${AQM_FIRE_FILE}"
+    set_file_param "${AQM_RC_IN_FP}" "YYMMDD" "${YYMMDD}"
```
christinaholtNOAA: Why not use a jinja template here, too, for consistency?

scripts/exregional_run_gsi_anal.sh
```diff
@@ -0,0 +1,673 @@
...
+    module use $( pwd -P )
+    module load modules.fv3
+    module list
+    set -x
+  else
+    . /apps/lmod/lmod/init/sh
+    module purge
+    module use /scratch4/NCEPDEV/nems/noscrub/emc.nemspara/soft/modulefiles
+    module load intel/16.1.150 impi/5.1.1.109 netcdf/4.3.0 
+    module list
```
christinaholtNOAA: This should all be handled by module files, not in MACHINE case statements, also this logic is inside the WCOSS_DELL_P3 case. Does that platform have a /scratch4? Is this right?

scripts/exregional_run_gsi_anal.sh
```diff
@@ -0,0 +1,673 @@
...
+  ulimit -a
+  APRUN="mpirun -l -np ${PE_MEMBER01}"
+  ;;
+#
+"HERA")
+  ulimit -s unlimited
+  ulimit -a
+  APRUN="srun"
+  LD_LIBRARY_PATH="${UFS_WTHR_MDL_DIR}/FV3/ccpp/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"
+  NCKS=/apps/nco/4.9.1/intel/18.0.5.274/bin/ncks
```
christinaholtNOAA: ncks should be in the environment from loading the module, so should not be redefined here. It overrides what is defined in the module file, and may not have access to the appropriate env variables set by the module.

scripts/exregional_run_gsi_anal.sh
```diff
@@ -0,0 +1,673 @@
...
+  ulimit -s unlimited
+  ulimit -a
+  APRUN="mpirun -l -np ${PE_MEMBER01}"
+  ;;
+#
+"HERA")
+  ulimit -s unlimited
+  ulimit -a
+  APRUN="srun"
+  LD_LIBRARY_PATH="${UFS_WTHR_MDL_DIR}/FV3/ccpp/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"
```
christinaholtNOAA: Is LD_LIBRARY_PATH needed here? It's not very standard on other tasks.

scripts/exregional_run_gsi_anal.sh
```diff
@@ -0,0 +1,673 @@
...
+#
+"ODIN")
+#
+  module list
+
+  ulimit -s unlimited
+  ulimit -a
+  APRUN="srun -n ${PE_MEMBER01}"
+  ;;
+#
```
christinaholtNOAA: Are Jet and Odin actually supported? It seemed that only Hera and WCOSS were supported in other tasks. It may be a good idea to remove these two options.

scripts/exregional_run_gsi_anal.sh
```diff
@@ -0,0 +1,673 @@
...
+#-----------------------------------------------------------------------
+
+cd_vrfy ${ANALWORKDIR}
+fixdir=$FIXgsi
+fixgriddir=$FIXgsi/${PREDEF_GRID_NAME}
+DA_CYCLE_INTERV=${INCR_CYCL_FREQ}
+if [ ${BKTYPE} -eq 1 ]; then  # cold start, use background from INPUT
+  bkpath=${CYCLE_DIR}/INPUT
+else
+  YYYYMMDDHHmInterv=`date +%Y%m%d%H -d "${START_DATE} ${DA_CYCLE_INTERV} hours ago"`
```
christinaholtNOAA: Please consider using the `$()` syntax since the back ticks are difficult to support. This applies throughout.

scripts/exregional_run_nexus.sh
```diff
@@ -0,0 +1,225 @@
...
+    ;;
+
+esac
+
+#
+#-----------------------------------------------------------------------
+#
+workdir="${CYCLE_DIR}/NEXUS"
+mkdir_vrfy -p "$workdir"
+cd_vrfy $workdir
```
christinaholtNOAA: This should be done in the job script for this task.

scripts/exregional_run_gsi_anal.sh
```diff
@@ -0,0 +1,673 @@
...
+START_DATE=`echo "${CDATE}" | sed 's/\([[:digit:]]\{2\}\)$/ \1/'`
+
+YYYYMMDDHH=`date +%Y%m%d%H -d "${START_DATE}"`
+JJJ=`date +%j -d "${START_DATE}"`
+
+YYYY=${YYYYMMDDHH:0:4}
+MM=${YYYYMMDDHH:4:2}
+DD=${YYYYMMDDHH:6:2}
+HH=${YYYYMMDDHH:8:2}
+YYYYMMDD=${YYYYMMDDHH:0:8}
```
christinaholtNOAA: Please user lower case here for consistency.

scripts/exregional_run_gsi_anal.sh
```diff
@@ -0,0 +1,673 @@
...
+fi
+#
+#-----------------------------------------------------------------------
+#
+# Set and export variables.
+#
+#-----------------------------------------------------------------------
+#
+export OMP_NUM_THREADS=1
+export OMP_STACKSIZE=2056M
```
christinaholtNOAA: Please move this to the top near the MACHINE case statement if possible

scripts/exregional_run_nexus.sh
```diff
@@ -0,0 +1,225 @@
...
+########################################################################
+
+#
+#----------------------------------------------------------------------
+# 
+# modify time configuration file
+#
+cp_vrfy ${EXECDIR}/../src/arl_nexus/utils/python/nexus_time_parser.py .
+echo ${start_date} ${end_date} # ${cyc}
+./nexus_time_parser.py -f ${workdir}/HEMCO_sa_Time.rc -s $start_date -e $end_date
```
christinaholtNOAA: It may be better here create a more robust log message and to allow for failure if this call to python fails. Also, does the script need to be copied to the run directory? Can it just be run from its original location? Same for other python calls.

scripts/exregional_run_nexus.sh
```diff
@@ -0,0 +1,225 @@
...
+mkdir_vrfy -p ${workdir}/inputs
+if [ ${NEI2016} == "TRUE" ]; then #NEI2016
+    cp_vrfy ${EXECDIR}/../src/arl_nexus/utils/python/nexus_nei2016_linker.py .
+    cp_vrfy ${EXECDIR}/../src/arl_nexus/utils/python/nexus_nei2016_control_tilefix.py .
+    mkdir_vrfy -p ${workdir}/inputs/NEI2016v1
+    mkdir_vrfy -p ${workdir}/inputs/NEI2016v1/v2020-07
+    mkdir_vrfy -p ${workdir}/inputs/NEI2016v1/v2020-07/${mm}
+    ./nexus_nei2016_linker.py --src_dir ${NEXUS_INPUT_BASE_DIR} --date ${yyyymmdd} --work_dir ${workdir}/inputs
+    ./nexus_nei2016_control_tilefix.py -f NEXUS_Config.rc -d ${yyyymmdd}
+fi
```
christinaholtNOAA, line 187: There's no logging or error handling here.

ush/bash_utils/set_file_param.sh
```diff
@@ -143,6 +143,11 @@ Setting parameter \"$param\" in file \"$file\" to \"$value\" ..."
     regex_replace="\1$value\5"
 #    set_bash_param "${file_fp}" "$param" "$value"
     ;;
+#
+  "${AQM_RC_IN_FN}")
+    regex_search="(<$param>)"
+    regex_replace="$value"
+    ;;
```
christinaholtNOAA: If we use the jinja template, this likely wouldn't be needed.

ush/config_defaults.sh
```diff
@@ -1882,6 +1882,168 @@ OMP_STACKSIZE_RUN_FCST="1024m"
...
 #
+# Parameters for RRFS-CMAQ
+#
+# AQM_RC_FN:
+# Name of resource file for NOAA Air Quality Model (AQM)
+#
+# AQM_CONFIG_DIR:
+# Configuration directory for AQM
+#
+# AQM_BIO_DiR:
```
christinaholtNOAA: Typo: i -> I

ush/config_defaults.sh
```diff
@@ -1882,6 +1882,168 @@ OMP_STACKSIZE_RUN_FCST="1024m"
...
+#
+# RUN_TASK_DACYC:
+# GSI analysis for AOD PM2.5
+# Set to "TRUE" to generate PM2.5/AOD DA tasks
+#
+# ENKF_FCST:
+# Path to ensemble members for hybrid GSI
+#
+#-----------------------------------------------------------------------
+#
```
christinaholtNOAA, line 2040: I think it would be helpful for these default values to be moved to the sections where other similar variables are stored. RUN_TASK* flags go with other RUN_TASK* flags, and so on.

scripts/exregional_run_chem_anal.sh
```diff
@@ -0,0 +1,379 @@
...
+ln_vrfy -sf ${DA_OBS_DIR}/${CDATE} ${CYCLE_DIR}/JEDI/Data/obs
+#
+#-----------------------------------------------------------------------
+#
+# Run Python script to create YAML
+#-----------------------------------------------------------------------
+#
+YAMLS='jedi_no2_3dvar.yaml jedi_no2_bump.yaml'
+TEMPLATEDIR=${USHDIR}/templates
+${USHDIR}/gen_JEDI_yaml.py -i $TEMPLATEDIR -o ${CYCLE_DIR}/JEDI/ -c ${CDATE} -y $YAMLS
```
christinaholtNOAA: More logging and error handling here would be helpful

ush/load_modules_run_task.sh
```diff
@@ -132,16 +132,26 @@ jjob_fp="$2"
...
 
+machine=$(echo_lowercase $MACHINE)
+
+if [[ "${FCST_MODEL}" = "fv3gfs_aqm" && ( "${task_name}" = "${RUN_FCST_TN}" || \
+   "${task_name}" = "${RUN_CHEM_ANAL_TN}" || "${task_name}" = "${RUN_GSI_ANAL_TN}" ) ]]; then
+
+module purge
+
+else
+
```
christinaholtNOAA: What is the need for this logic? We should be able to handle all tasks the same way.

ush/setup.sh
```diff
@@ -1033,6 +1240,10 @@ VX_CONFIG_DIR="$TEMPLATE_DIR/parm"
 METPLUS_CONF="$TEMPLATE_DIR/parm/metplus"
 MET_CONFIG="$TEMPLATE_DIR/parm/met"
 
+ARL_NEXUS_DIR="src/arl_nexus"
+JEDI_DIR="src/jedi"
+GSI_DIR="src/gsi"
```
christinaholtNOAA: These should have full paths, as the others do.

ush/setup.sh
```diff
@@ -1195,6 +1484,77 @@ The base directory in which the UPP source code should be located
 Please clone the external repository containing the code in this directory,
 build the executable, and then rerun the workflow."
 fi
+
+#
+##### RRFS-CMAQ ########## start #####
+#
+if [ "${FCST_MODEL}" = "fv3gfs_aqm" ]; then
+#
```
christinaholtNOAA, line 683: Please use appropriate indentation throughout this file for readability.

ush/templates/fv3gfs_aqm.input.yml
```diff
@@ -0,0 +1,30 @@
...
+FV3_HRRR:
+
+FV3_GFS_2017_gfdlmp:
+
+FV3_GFS_v15p2:
+
+FV3_GFS_v16beta:
+
+FV3_CPT_v0:
+
```
christinaholtNOAA, line 30: I'm a bit confused about what this file is meant to do.

ush/templates/fv3gfs_aqm_wflow.xml
```diff
@@ -0,0 +1,1260 @@
+{#
```
christinaholtNOAA: In general, this method of handling different XMLs is going to be very difficult to support given how much duplication exists between the two templates.

ush/templates/model_configure.fv3gfs_aqm
```diff
@@ -0,0 +1,129 @@
+total_member:            1
```
christinaholtNOAA, line 1: How does this file differ from the existing model_configure template?

christinaholtNOAA (changes requested):
This is quite the large PR with the addition of many new capabilities. I have added comments throughout related to style and duplication of information, but I have a couple of larger concerns:

1) How are we going to manage XMLs for different workflows. Even having two that are so complex and mostly duplicated is going to be a huge undertaking to keep in sync that we likely do not want. 

2) Is there a true requirement for Anaconda, or does the external repo provide guidance on dependencies so that we can maintain a supported conda environment?

jobs/JREGIONAL_ADD_AQM_ICS
chan-hoo: You're correct. I'll remove it.

jobs/JREGIONAL_RUN_CHEM_ANAL
chan-hoo: I agree. I'll modify it.

jobs/JREGIONAL_RUN_GSI_ANAL
chan-hoo: Yes. I'll modify it.

jobs/JREGIONAL_ADD_AQM_ICS
chan-hoo: done.

jobs/JREGIONAL_RUN_CHEM_ANAL
chan-hoo: done

jobs/JREGIONAL_RUN_GSI_ANAL
chan-hoo: done.

jobs/JREGIONAL_RUN_GSI_ANAL
chan-hoo: I agree. I'll modify it.

ush/setup.sh
chan-hoo: done

ush/templates/fv3gfs_aqm.input.yml
chan-hoo: Since FV3GFS_AQM is currently coupled with GFS v15, CCPP options are not available in RRFS-CMAQ. It will be updated with the higher version of GFS soon (hopefully). This is the reason why this file has nothing at this moment.

ush/templates/model_configure.fv3gfs_aqm
chan-hoo: As I mentioned above, the current version of FV3GFS_AQM is coupled with GFS v15. So, its configuration is a quite old version. For example, the write-component output parameters are different from the ufs weather model in the UFS SRW App (FV3-LAM).

ush/templates/fv3gfs_aqm_wflow.xml
chan-hoo: I agree. The workflow template for RRFS-CMAQ includes that for FV3-LAM. I tried to make one XML file for both, but I concluded that separate XML files would be more efficient for developers to manage both FV3-LAM and RRFS-CMAQ because many FV3-LAM developers are not familiar with RRFS-CMAQ and the verification part for RRFS-CMAQ, which is a future plan, might be different from that for FV3-LAM.

jobs/JREGIONAL_RUN_GSI_ANAL
chan-hoo: As you pointed out, CYCLE_DIR = CYCLE_BASEDIR+CDATE and it can be removed from the script. However, this is how other tasks do in the template. For consistency, I think it would be better to modify them together in a separate PR. I'll remove 'CYCLE_ROOT' from this script because it is a duplicate as you said.

jobs/JREGIONAL_RUN_NEXUS
chan-hoo: You're right! I'll remove the duplicate.

jobs/JREGIONAL_RUN_GSI_ANAL
chan-hoo: done

jobs/JREGIONAL_RUN_NEXUS
chan-hoo: done

ush/setup.sh
chan-hoo: done

jobs/JREGIONAL_RUN_POST_STAT
chan-hoo: This task evaluates the statistics of the specific fields 'O3' and 'PM2.5' from the result files of 'RUN_POST' only for RRFS-CMAQ. This is the reason why its name has the suffix '-STAT'. For this task, all the 'RUN_POST' tasks must be completed. Therefore, this task should have a separate task and script from RUN_POST. This was also a request from the air quality modeling group.

modulefiles/tasks/hera/add_aqm_lbcs.local
chan-hoo: The 'add_aqm_lbcs' has another sub-task using separately compiled executable 'gefs2lbc_para' to add GEFS aerosol data to LBCS.

modulefiles/tasks/wcoss_dell_p3/add_aqm_lbcs.local
chan-hoo: Yes. This task runs additional sub-task 'gefs2lbc_para'.

scripts/exregional_add_aqm_ics.sh
chan-hoo: done

modulefiles/tasks/hera/run_chem_anal.local
chan-hoo: I agree. It would be great if you can help us. @CoryMartin-NOAA, what do you think about this? Can you answer @christinaholtNOAA's comment?

modulefiles/tasks/hera/run_chem_anal.local
chan-hoo: @CoryMartin-NOAA, can you answer @christinaholtNOAA's comment?

modulefiles/tasks/hera/add_aqm_ics.local
chan-hoo: updated.

modulefiles/tasks/hera/run_chem_anal.local
CoryMartin-NOAA: Agreed, this is not ideal. For this PR, you're right, I believe all that is needed is standard to python3 and really just creates a YAML file from a template. As the JEDI capabilities in these workflows mature, we will likely need more python packages, but I think for now, the standard 'regional_workflow' conda env is probably sufficient here.

modulefiles/tasks/hera/run_chem_anal.local
CoryMartin-NOAA: Are you saying replace line 13 with the variable set in line 12? if so, then sure, I'm not very familiar with writing these sort of modulefiles, so I wasn't sure of the syntax. Basically, all we need is `$JEDI_OPT` to be set for the subsequent lua modules to be loaded correctly.

ush/templates/fv3gfs_aqm_wflow.xml
chan-hoo: There are two different add_aqm_ics tasks. One is for the first cycle 'add_aqm_extrn_ics' and the other is for the other cycles 'add_aqm_ics'. For 'add_aqm_extrn_ics', since this is the first cycle, the directory containing the restart data for the first cycle (RESTART_CYCLE_DIR) should be specified in the configuration file. Therefore, a cyclestr with an offset is not necessary for 'add_aqm_extrn_ics'.

scripts/exregional_add_aqm_ics.sh
chan-hoo: Some scripts such as make_ic/lbcs do not define the working or tmp directories under 'jobs' but under 'scripts. I've checked the NCO standards. You're right!! I'll modify the scripts for RRFS-CMAQ. I think it would be better to modify the other scripts for FV3-LAM later in a separate PR to avoid any conflicts. What do you think?

chan-hoo: done

scripts/exregional_add_aqm_ics.sh
chan-hoo: @rmontuoro, can you answer @christinaholtNOAA's comment?

scripts/exregional_add_aqm_lbcs.sh
chan-hoo: updated.

scripts/exregional_add_aqm_lbcs.sh
chan-hoo: @rmontuoro, can you answer @christinaholtNOAA's comment?

scripts/exregional_run_chem_anal.sh
chan-hoo: done

scripts/exregional_add_aqm_lbcs.sh
chan-hoo: done

scripts/exregional_run_chem_anal.sh
chan-hoo: done

scripts/exregional_add_aqm_lbcs.sh
chan-hoo: @ytangnoaa, can you answer @christinaholtNOAA's comment? Can you send me a short description for the parameters? I'll add it to the script.

modulefiles/tasks/hera/add_aqm_lbcs.local
ytangnoaa: The NEMSIO and associated NCEP libraries were compiled with this setting, and we need to have them
consist with 
module use -a /scratch2/NCEPDEV/nwprod/NCEPLIBS/modulefiles
module load bufr bufr_dumplist dumpjb w3nco w3emc g2c bacio ip ip2 jasper prod_util nemsio nemsiogfs

scripts/exregional_run_chem_anal.sh
chan-hoo: updated.

scripts/exregional_add_aqm_lbcs.sh
ytangnoaa: They (bndname, mofile, lbcfile, topofile) are not the shell environmental variables, but the FORTRAN namelist variables.

cat > gefs2lbc-nemsio.ini <<EOF
 &control
 ....
  bndname
 ...
&end
...
EOF

scripts/exregional_run_chem_anal.sh
chan-hoo: @CoryMartin-NOAA, can you answer @christinaholtNOAA's comment?

scripts/exregional_run_chem_anal.sh
chan-hoo: added.

scripts/exregional_run_chem_anal.sh
chan-hoo: moved.

scripts/exregional_run_fcst.sh
chan-hoo: Can we update this later in a separate PR for FV3-LAM?

chan-hoo: I agree. I'll modify it.

scripts/exregional_run_chem_anal.sh
CoryMartin-NOAA: Not sure I quite understand what you're asking here. This is checking to make sure that the previous forecast's files valid at the correct time exist and if not, error.

scripts/exregional_run_fcst.sh
chan-hoo: done.

scripts/exregional_run_nexus.sh
chan-hoo: done

ush/load_modules_run_task.sh
chan-hoo: The three tasks 'run_fcst', 'run_gsi', and 'run_chem_anl (JEDI)' in RRFS-CMAQ do not use the environment file in ufs-srw-app/env because the executables for these three tasks are compiled separately with different modules. Therefore, the common env file is not necessary for these tasks. In the future, if all the executables including these three are compiled with one env file, this part should be removed.

ush/config_defaults.sh
chan-hoo: fixed.

scripts/exregional_run_gsi_anal.sh
chan-hoo: done

scripts/exregional_run_gsi_anal.sh
chan-hoo: done

scripts/exregional_run_gsi_anal.sh
chan-hoo: @hongli-wang, your gsi script works only on Hera. Is this right? If so, I'll remove the 'wcoss_dell_p3' part from the script. If not, can you answer @christinaholtNOAA's comment?

scripts/exregional_run_gsi_anal.sh
chan-hoo: @hongli-wang, I agree with @christinaholtNOAA. Have you tested it before?

scripts/exregional_run_gsi_anal.sh
chan-hoo: @hongli-wang, can you answer @christinaholtNOAA's comment?

scripts/exregional_run_gsi_anal.sh
chan-hoo: I agree. I'll remove them.

scripts/exregional_run_gsi_anal.sh
chan-hoo: I've tested and removed this.

scripts/exregional_run_gsi_anal.sh
chan-hoo: removed.

scripts/exregional_run_gsi_anal.sh
chan-hoo: done.

ush/config_defaults.sh
chan-hoo: moved.

scripts/exregional_run_nexus.sh
chan-hoo: @bbakernoaa, can you answer @christinaholtNOAA's comment?

scripts/exregional_run_nexus.sh
chan-hoo: @bbakernoaa, can you answer @christinaholtNOAA's comment?

scripts/exregional_run_fcst.sh
chan-hoo: replaced with a jinja template.

ush/bash_utils/set_file_param.sh
chan-hoo: removed.

scripts/exregional_add_aqm_ics.sh
chan-hoo: As far as I know, the nco commands exit with a non-zero status.

scripts/exregional_add_aqm_lbcs.sh
chan-hoo: ncks will return a non-zero status when it fails.

scripts/exregional_run_chem_anal.sh
chan-hoo: @CoryMartin-NOAA, can you answer this? or email me how to modify this?

CoryMartin-NOAA: We can add some print statements, sure. Any error I think should result in a job failure. Probably need to at the very least check/ensure that if the python script fails, the shell script fails from the `exit 1`. What would be enough log information? A print out of what is being written to the output file?

chan-hoo:
@christinaholtNOAA, this PR is ready for another review!

christinaholtNOAA:
Thanks for letting me know. I will try to get to it early next week. Is that acceptable?

chan-hoo:
@christinaholtNOAA, sure. I'll need to update this PR with the latest develop branch by then. I found a conflict with the latest.

scripts/exregional_run_gsi_anal.sh
chan-hoo: removed.

scripts/exregional_run_gsi_anal.sh
chan-hoo: removed

ush/templates/fv3gfs_aqm_wflow.xml
chan-hoo: To combine two different XML files, we'll need to change the name of the current XML file and run the we2e tests for FV3-LAM. To make it clear, can I combine them in a separate PR later after this PR is merged?

chan-hoo: As I mentioned above, the workflow template for RRFS-CMAQ was designed to work for FV3-LAM as well. I'll run the multiple we2e tests with it and update the scripts.

chan-hoo: @christinaholtNOAA, I confirmed that the workflow template for RRFS-CMAQ worked well for FV3-LAM as well. so I've removed one template and renamed 'FV3LAM_wflow.xml' to 'regional_wflow.xml' for both.

JeffBeck-NOAA (commented):
Hi @chan-hoo, I saw you renamed the workflow from FV3-LAM to "regional" in the latest commit to this PR.  Just want to make sure you feel this is absolutely necessary?  The basis of the RRFS-CMAQ will still be the FV3-LAM, so could it not be left as is?  Thanks!

chan-hoo:
@JeffBeck-NOAA, I thought that FV3-LAM represented the uncoupled model (FV3 stand-alone). I'll return it to the original. If FV3-LAM covers all regional models, how can we distinguish between the coupled air quality model (RRFS-CMAQ) and the uncoupled regional model (???)? Don't we need to do this?

JeffBeck-NOAA:
@chan-hoo, the SRW App uses the Limited Area Model version of the FV3 in many different configurations, whether it is run in coupled mode or not.  Therefore, the workflow itself would still be referred to as "FV3LAM," and coupling, if run, is just defined within that configuration of the FV3-LAM (within var_defns.sh and the XML).  In other words, the RRFS-CMAQ is a unique and specific, coupled configuration of the FV3-LAM, just as (currently) the RRFS is a specific configuration of the FV3-LAM, with a chosen domain, DA configuration, etc.  It's the other branding (RRFS-CMAQ, RRFS, etc.) that provides more information on the configuration of the SRW App and inform people of how the FV3-LAM is being used.

chan-hoo:
@JeffBeck-NOAA, I got it. RRFS: FV3 stand-alone, RRFS-CMAQ: air quality model, RRFS-DA: FV3 stand-alone DA. Is this correct?

JeffBeck-NOAA:
> @JeffBeck-NOAA, I got it. RRFS: FV3 stand-alone, RRFS-CMAQ: air quality model, RRFS-DA: FV3 stand-alone DA. Is this correct?

@chan-hoo, yeah, so RRFS-CMAQ and RRFS are different instantiations of the SRW app.  All of these versions of the SRW App use the FV3-LAM.

scripts/exregional_add_aqm_ics.sh
gsketefian: Even if an nco commands exit with a non-zero status, the script will continue to run.  To stop immediately after a command fails, do something like this (as in other scripts) if you want the script to immediately stop and exit (which I assume you do most of the time):
```
ncks --fix_rec_dmn lev -C -O -o tmp2.nc tmp1.nc || print_err_msg_exit "\
Call to ncks failed with exit code $?."
```
And of course this is untested code...

scripts/exregional_add_aqm_lbcs.sh
gsketefian: If some nco commands fail, without checking for the exit status of each command, the script will still make it here and report a false success.  You can use the

```
some_command || print_err_msg_exit "..." 
```

approach I mentioned above to avoid this.

scripts/exregional_add_aqm_lbcs.sh
gsketefian: I think what @christinaholtNOAA is asking (correct me if I'm wrong) is whether a line like

`topofile='${OROG_DIR}/${CRES}_oro_data.tile7.halo4.nc'
`

should use double quotes instead on the right side, i.e.

`
topofile="${OROG_DIR}/${CRES}_oro_data.tile7.halo4.nc"
`

because the first method will not substitute the proper values for the bash variables `OROG_DIR` and `CRES` before putting the line for `topofile` into the namelist file (instead, the variable names `OROG_DIR` and `CRES` will appear in the file).  That's just because a string in single quotes is treated literally in bash.  Actually, maybe you want escaped single quotes since you need to put those around a string in a fortran namelist, i.e.:

`
topofile=\'${OROG_DIR}/${CRES}_oro_data.tile7.halo4.nc\'
`

Has this script been tested?  I'd be surprised if it works as-is.

scripts/exregional_run_chem_anl.sh
```diff
@@ -0,0 +1,343 @@
...
+#
+# Create links to fix files and  executables
+#
+#-----------------------------------------------------------------------
+#
+print_info_msg "$VERBOSE" "
+Creating links in the JEDI subdirectory of the current cycle's run directory 
+to the necessary executables and fix files: \"${JEDI_WORKDIR}\""
+# executables
+ln_vrfy -sf $EXECDIR/fv3jedi_error_covariance_training.x ${JEDI_WORKDIR}/.
```
gsketefian: When making symlinks, often the system doesn't tell you if the target doesn't exist and just creates the symlink anyway, and that can take some time to track down.  I encountered this error a few times using `ln` or `ln_vrfy` and so eventually wrote the function `create_symlink_to_file()` that checks for target existence (and gracefully allows for different treatment of relative symlinks between machines).  I'm not insisting you use it, just letting you know it's available.  So the line
```
ln_vrfy -sf $EXECDIR/fv3jedi_error_covariance_training.x ${JEDI_WORKDIR}/.
```
would become
```
create_symlink_to_file target="$EXECDIR/fv3jedi_error_covariance_training.x" symlink="${JEDI_WORKDIR}/."
```

scripts/exregional_run_chem_anl.sh
```diff
@@ -0,0 +1,343 @@
...
+
+relative_or_null=""
+if [ "${RUN_TASK_MAKE_OROG}" = "TRUE" ]; then
+  relative_or_null="--relative"
+fi
+
+# Symlink to halo-0 orography file with "${CRES}_" and "halo0" stripped from name.
+target="${FIXLAM}/${CRES}${DOT_OR_USCORE}oro_data.tile${TILE_RGNL}.halo${NH0}.nc"
+symlink="oro_data.nc"
+if [ -f "${target}" ]; then
```
gsketefian: So for example here, if you use `create_symlink_to_file()` you can avoid the if-statement for checking the target, and you don't need to set the variable `relative_or_null` above.  You can just say:

```
target="${FIXLAM}/${CRES}${DOT_OR_USCORE}oro_data.tile${TILE_RGNL}.halo${NH0}.nc"
symlink="oro_data.nc"
create_symlink_to_file target="$target" symlink="$symlink" relative="TRUE"
```

scripts/exregional_run_fcst.sh
```diff
@@ -466,6 +483,45 @@ if [ ${WRITE_DOPOST} = "TRUE" ]; then
   cp_vrfy ${UPP_DIR}/parm/params_grib2_tbl_new .
 fi
 
+if [ "${FCST_MODEL}" = "fv3gfs_aqm" ]; then
+  # AQM_RC_IN_FN: input file name of cmaq
+  AQM_RC_IN_FN="aqm.rc"
```
gsketefian: If `AQM_RC_IN_FN` is only a local variable, it would be nice to make it lowercase since we usually use uppercase only for the global experiment variables.  Same for `AQM_RC_IN_FP`.

Btw, should `AQM_RC_IN_FN` be replaced with the global experiment variable `AQM_RC_FN`?

scripts/exregional_run_post_stat.sh
```diff
@@ -0,0 +1,181 @@
+#!/bin/bash
+set -x
```
gsketefian: If you want this script to always use `-x`, it's better to set it in the line
```
{ save_shell_opts; set -u +x; } > /dev/null 2>&1
```
below (by changing the `+x` to `-x`), then restore whatever the shell options used to be before this call by using 
```
{ restore_shell_opts; } > /dev/null 2>&1
```

scripts/exregional_run_post_stat.sh
```diff
@@ -0,0 +1,181 @@
...
+cd_vrfy "${poststat_dir}"
+
+basetime=$( $DATE_UTIL --date "$yyyymmdd $hh" +%y%j%H%M )
+
+for fhr in $(seq -f "%03g" 1 ${fcst_len_hrs}); do
+
+  input_file="NATLEV_${basetime}f${fhr}00"
+
+  if [ ${fhr} = "001" ]; then
+     wgrib2 ${postprd_dir}/${input_file} -match ":${field1}" -match ":${field3}" -GRIB ${output_file}
```
gsketefian: Might want to do more error handling here as well (`... || print_err_msg_exit "..."`).

ush/generate_FV3LAM_wflow.sh
```diff
@@ -739,6 +792,24 @@ fi
 # It turns out that setting the variable to an empty string also works
 # to remove it from the namelist!  Which is better to use??
 #
+
+if [ "${FCST_MODEL}" = "fv3gfs_aqm" ] && [ "${CCPP_PHYS_SUITE}" = "FV3_GFS_2017_gfdlmp" ]; then
```
gsketefian: Does the `fv3gfs_aqm` run only with the `FV3_GFS_2017_gfdlmp` suite?

chan-hoo: @gsketefian, since fv3gfs_aqm couples with GFSv15, any CCPP options are not available. I'll remove the ccpp suite condition.

ush/create_aqm_rc_file.sh
```diff
@@ -0,0 +1,151 @@
...
+#-----------------------------------------------------------------------
+#
+# Call a python script to generate the experiment's actual aqm.rc file
+# from the template file.
+#
+#-----------------------------------------------------------------------
+#
+  $USHDIR/fill_jinja_template.py -q \
+                                 -u "${settings}" \
+                                 -t ${AQM_RC_FP} \
```
gsketefian, line 129: `AQM_RC_FP` doesn't seem to be set anywhere.  Is it supposed to be a global experiment variable?  If so, add to `config_defaults.sh` and/or set it in `setup.sh`.

ush/create_aqm_rc_file.sh
```diff
@@ -0,0 +1,151 @@
...
+# process the arguments provided to this script/function (which should
+# consist of a set of name-value pairs of the form arg1="value1", etc).
+#
+#-----------------------------------------------------------------------
+#
+  local valid_args=(
+cdate \
+run_dir \
+init_concentrations \
+aqm_rc_in_fn \
```
gsketefian, line 52: Doesn't seem like the input argument/variable `aqm_rc_in_fn` is used in this script.

ush/create_aqm_rc_file.sh
```diff
@@ -0,0 +1,151 @@
...
+#-----------------------------------------------------------------------
+#
+  $USHDIR/fill_jinja_template.py -q \
+                                 -u "${settings}" \
+                                 -t ${AQM_RC_FP} \
+                                 -o ${aqm_rc_in_fp} || \
+  print_err_msg_exit "\
+Call to python script fill_jinja_template.py to create a \"${AQM_RC_IN_FN}\"
+file from a jinja2 template failed.  Parameters passed to this script are:
+  Full path to template rocoto XML file:
```
gsketefian: "rocoto XML" should be changed to "resource (configuration?)" file.  Same thing two lines below.

ush/create_aqm_rc_file.sh
```diff
@@ -0,0 +1,151 @@
...
+    AQM_RC_FP = \"${AQM_RC_FP}\"
+  Full path to output rocoto XML file:
+    aqm_rc_in_fp = \"${aqm_rc_in_fp}\"
```
gsketefian, line 137: Having the string `in` in `aqm_rc_in_fp` is confusing to me because it makes it seem like it is some sort of input or initial file, while actually it is an output file.  Maybe rename it to just `aqm_rc_fp`?

scripts/exregional_add_aqm_ics.sh
```diff
@@ -0,0 +1,203 @@
...
+  Looking for tracer restart file: \"${fv_tracer_file}\""
+if [ ! -r ${fv_tracer_file} ]; then
+  if [ -r ${rst_dir}/coupler.res ]; then
+    rst_info=( $( tail -n 1 ${rst_dir}/coupler.res ) )
+    rst_date=$( printf "%04d%02d%02d%02d" ${rst_info[@]:0:4} )
+    print_info_msg "
+  Tracer file not found. Checking available restart date:
+    requested date: \"${CDATE}\"
+    available date: \"${rst_date}\""
+    if [ "${rst_date}" == "${CDATE}" ] ; then
```
gsketefian: Use single equal sign here since using single brackets (see [this](https://unix.stackexchange.com/questions/72039/whats-the-difference-between-single-and-double-equal-signs-in-shell-compari)).

scripts/exregional_run_chem_anl.sh
```diff
@@ -0,0 +1,343 @@
...
+  Looking for tracer restart file: \"${fv_tracer_file}\""
+if [ ! -r ${fv_tracer_file} ]; then
+  if [ -r ${rst_dir}/coupler.res ]; then
+    rst_info=( $( tail -n 1 ${rst_dir}/coupler.res ) )
+    rst_date=$( printf "%04d%02d%02d%02d" ${rst_info[@]:0:4} )
+    print_info_msg "
+  Tracer file not found. Checking available restart date:
+    requested date: \"${CDATE}\"
+    available date: \"${rst_date}\""
+    if [ "${rst_date}" == "${CDATE}" ] ; then
```
gsketefian: Use single equal sign here since using single brackets (see [this](https://unix.stackexchange.com/questions/72039/whats-the-difference-between-single-and-double-equal-signs-in-shell-compari)).

scripts/exregional_run_nexus.sh
```diff
@@ -0,0 +1,227 @@
...
+# set the root directory to the temporary directory
+#
+cp_vrfy ${ARL_NEXUS_DIR}/utils/python/nexus_root_parser.py .
+./nexus_root_parser.py -f ${NEXUS_WORKDIR}/NEXUS_Config.rc -d ${NEXUS_WORKDIR_INPUT}
+
+#
+#----------------------------------------------------------------------
+# Get all the files needed (TEMPORARILY JUST COPY FROM THE DIRECTORY)
+#
+if [ ${NEI2016} == "TRUE" ]; then #NEI2016
```
gsketefian: Use single equal sign here since using single brackets (see [this](https://unix.stackexchange.com/questions/72039/whats-the-difference-between-single-and-double-equal-signs-in-shell-compari)).

scripts/exregional_run_nexus.sh
```diff
@@ -0,0 +1,227 @@
...
+    cp_vrfy ${ARL_NEXUS_DIR}/utils/python/nexus_nei2016_linker.py .
+    cp_vrfy ${ARL_NEXUS_DIR}/utils/python/nexus_nei2016_control_tilefix.py .
+    mkdir_vrfy -p ${NEXUS_WORKDIR_INPUT}/NEI2016v1
+    mkdir_vrfy -p ${NEXUS_WORKDIR_INPUT}/NEI2016v1/v2020-07
+    mkdir_vrfy -p ${NEXUS_WORKDIR_INPUT}/NEI2016v1/v2020-07/${mm}
+    ./nexus_nei2016_linker.py --src_dir ${NEXUS_INPUT_BASE_DIR} --date ${yyyymmdd} --work_dir ${NEXUS_WORKDIR_INPUT}
+    ./nexus_nei2016_control_tilefix.py -f NEXUS_Config.rc -d ${yyyymmdd}
+fi
+
+if [ ${TIMEZONES} == 'TRUE' ]; then # TIME ZONES
```
gsketefian: Use single equal sign here since using single brackets (see [this](https://unix.stackexchange.com/questions/72039/whats-the-difference-between-single-and-double-equal-signs-in-shell-compari)).

scripts/exregional_run_nexus.sh
```diff
@@ -0,0 +1,227 @@
...
+    cp_vrfy -r ${NEXUS_INPUT_BASE_DIR}/TIMEZONES ${NEXUS_WORKDIR_INPUT}
+fi
+
+if [ ${MASKS} == 'TRUE' ]; then # MASKS
```
gsketefian: Use single equal sign here since using single brackets (see [this](https://unix.stackexchange.com/questions/72039/whats-the-difference-between-single-and-double-equal-signs-in-shell-compari)).

ush/create_aqm_rc_file.sh
chan-hoo: @gsketefian, the input file name of fv3gfs_aqm should be "aqm.rc". However, the users in the air quality modeling group want to use their own "aqm_rc" files with different names like "aqm.rc_ver1" and "aqm.rc_ver2". So I set two different parameters "aqm_rc_fp" for a user's template file and "aqm_rc_in_fp" for actual input file. I've modified the message.

scripts/exregional_run_chem_anl.sh
chan-hoo: updated.

scripts/exregional_run_chem_anl.sh
chan-hoo: updated

scripts/exregional_run_fcst.sh
chan-hoo: changed to lowercase. AQM_RC_IN_FN is different from AQM_RC_FN. AQM_RC_FN is the user-defined template name of 'aqm.rc', for example aqm.rc_ver1 or aqm.rc_ver2, while AQM_RC_IN_FN is the actual input file name that should be 'aqm.rc'.  So AQM_RC_FN should be replaced with AQM_RC_IN_FN before running the task "run_fcst".

scripts/exregional_run_post_stat.sh
chan-hoo: Removed

scripts/exregional_run_post_stat.sh
chan-hoo: Added

ush/create_aqm_rc_file.sh
chan-hoo: AQM_RC_FP is defined in 'setup.sh' as a global variable.

ush/create_aqm_rc_file.sh
chan-hoo: It is used in the error message.

ush/create_aqm_rc_file.sh
chan-hoo: Changed.

scripts/exregional_add_aqm_ics.sh
chan-hoo: Done

scripts/exregional_run_chem_anl.sh
chan-hoo: Done

scripts/exregional_run_nexus.sh
chan-hoo: Done

scripts/exregional_run_nexus.sh
chan-hoo: Done

scripts/exregional_run_nexus.sh
chan-hoo: Done

Issue #618, opened 2021-10-15
Pull out machine-specific information to a separate config file.


christinaholtNOAA:
## Description
I'd like to contribute a modification that pulls out as many MACHINE case statements as possible to use machine-specific configuration files that can be used during configuration and run-time so that users running on unsupported platforms do not have to modify scripts at all. 

## Solution
Remove as many MACHINE case statements as possible, and introduce a set of bash config files -- one for each supported machine. These will keep track of variables for machine architecture, run commands, and expected file paths.

## Related to 
Issue #610 -- While changes to address this issue will take into account the ability to run with a linux platform, it will still require quite a bit of script editing. For running automated testing, it will be quite difficult to specify necessary configurations to the existing tests to run on separate platforms.

mkavulich:
Fixed via #666

PR #666, opened 2022-01-13
Make platform information configurable.

christinaholtNOAA:
## DESCRIPTION OF CHANGES: 
These changes move the machine-specific configuration information from the configuration, run time, and test scripts (at least a huge portion of it, with some exceptions for run time environment management on WCOSS machines). The machine-specific information now lives in `ush/machine/*.sh` files. 

In addition, a new bash_util, boolify, was added to clean up `ush/setup.sh`. The function can be called for any boolean flag that can be provided with valid values like `("TRUE" "true" "YES" "yes" "FALSE" "false" "NO" "no")`.

I will review the code just after opening this PR to walk folks through some of the less obvious changes. 

## TESTS CONDUCTED: 

Running a large suite of tests on Hera. I will report their status when complete:
```
grid_CONUS_25km_GFDLgrid_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
grid_CONUS_3km_GFDLgrid_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
grid_GSD_HRRR_AK_50km_ics_RAP_lbcs_RAP_suite_GSD_SAR
grid_RRFS_AK_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
grid_RRFS_AK_13km_ics_RAP_lbcs_RAP_suite_GSD_SAR
grid_RRFS_AK_3km_ics_RAP_lbcs_RAP_suite_GSD_SAR
grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
grid_RRFS_CONUS_13km_ics_HRRR_lbcs_RAP_suite_GSD_SAR
grid_RRFS_CONUS_13km_ics_HRRR_lbcs_RAP_suite_HRRR
grid_RRFS_CONUS_13km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha
grid_RRFS_CONUS_13km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_2017_gfdlmp
grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GSD_SAR
grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_CPT_v0
grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_2017_gfdlmp
grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v15p2
grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v16
grid_RRFS_CONUS_25km_ics_HRRR_lbcs_HRRR_suite_GSD_SAR
grid_RRFS_CONUS_25km_ics_HRRR_lbcs_HRRR_suite_HRRR
grid_RRFS_CONUS_25km_ics_HRRR_lbcs_HRRR_suite_RRFS_v1beta
grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_GSD_SAR
grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_GSD_v0
grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_HRRR
grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha
grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_GSD_SAR
grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_HRRR
grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_RRFS_v1beta
grid_RRFS_NA_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GSD_v0
nco_grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
nco_grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
nco_grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_GFS_v15p2
nco_grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_HRRR
nco_grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
GST_release_public_v1
community_ensemble_008mems
community_ensemble_2mems
get_from_HPSS_ics_FV3GFS_lbcs_FV3GFS_fmt_grib2_2019061200
get_from_HPSS_ics_FV3GFS_lbcs_FV3GFS_fmt_grib2_2019101818
get_from_HPSS_ics_FV3GFS_lbcs_FV3GFS_fmt_grib2_2020022518
get_from_HPSS_ics_FV3GFS_lbcs_FV3GFS_fmt_grib2_2020022600
get_from_HPSS_ics_FV3GFS_lbcs_FV3GFS_fmt_grib2_2021010100
get_from_HPSS_ics_FV3GFS_lbcs_FV3GFS_fmt_nemsio_2019061200
get_from_HPSS_ics_FV3GFS_lbcs_FV3GFS_fmt_nemsio_2019101818
get_from_HPSS_ics_FV3GFS_lbcs_FV3GFS_fmt_nemsio_2020022518
get_from_HPSS_ics_FV3GFS_lbcs_FV3GFS_fmt_nemsio_2020022600
get_from_HPSS_ics_FV3GFS_lbcs_FV3GFS_fmt_nemsio_2021010100
get_from_HPSS_ics_FV3GFS_lbcs_FV3GFS_fmt_nemsio
get_from_HPSS_ics_FV3GFS_lbcs_FV3GFS_fmt_netcdf_2021062000
get_from_HPSS_ics_GSMGFS_lbcs_GSMGFS
get_from_HPSS_ics_HRRR_lbcs_RAP
get_from_HPSS_ics_RAP_lbcs_RAP
get_from_NOMADS_ics_FV3GFS_lbcs_FV3GFS_fmt_nemsio
inline_post
MET_ensemble_verification
MET_verification
nco_ensemble
nco_inline_post
new_ESGgrid
new_GFDLgrid__GFDLgrid_USE_GFDLgrid_RES_IN_FILENAMES_eq_FALSE
new_GFDLgrid__GFDLgrid_USE_GFDLgrid_RES_IN_FILENAMES_eq_TRUE
new_GFDLgrid
pregen_grid_orog_sfc_climo
specify_DOT_OR_USCORE
specify_DT_ATMOS_LAYOUT_XY_BLOCKSIZE
specify_EXTRN_MDL_SYSBASEDIR_ICS_LBCS
specify_RESTART_INTERVAL
subhourly_post_ensemble_2mems
subhourly_post
```

I only have resources to test on Jet, and can, but can other folks help with testing other platforms?

## ISSUE: 
Fixes issue mentioned in #618


scripts/exregional_make_grid.sh
```diff
@@ -92,7 +92,7 @@ case "$MACHINE" in
     module list
     { restore_shell_opts; } > /dev/null 2>&1
     export NODES=1
-    export APRUN="aprun -n 1 -N 1 -j 1 -d 1 -cc depth"
+    export RUN_CMD_SERIAL="aprun -n 1 -N 1 -j 1 -d 1 -cc depth"
```
christinaholtNOAA, line 95: In each ex-script, I have unified the use of RUN_CMD that should be used for that task. It is one of those template variables that needs to be filled in at run time, and should be done in each script before doing the `eval echo` to the RUN_CMD.

Also, I've left the MACHINE case statement in the ex-scripts to handle the WCOSS machines. The run time was difficult enough that I didn't want to break it with no way to test.

scripts/exregional_make_grid.sh
```diff
@@ -102,56 +102,25 @@ case "$MACHINE" in
...
+
+if [ -z ${RUN_CMD_SERIAL:-} ] ; then
+  print_err_msg_exit " \
+  Run command was not set in machine file. \
+  Please set RUN_CMD_SERIAL for your platform"
+else
+  RUN_CMD_SERIAL=$(eval echo ${RUN_CMD_SERIAL})
+  print_info_msg "$VERBOSE" "
+  All executables will be submitted with command \'${RUN_CMD_SERIAL}\'."
+fi
```
christinaholtNOAA, line 123: Each ex-script should check for the existence of a RUN_CMD and exits if it isn't set.

scripts/exregional_make_grid.sh
```diff
@@ -102,56 +102,25 @@ case "$MACHINE" in
...
-
-  "LINUX")
-    APRUN=time
-    ;;
-
   *)
-    print_err_msg_exit "\
-Run command has not been specified for this machine:
-  MACHINE = \"$MACHINE\""
+    source ${MACHINE_FILE}
```
christinaholtNOAA, line 110: Each ex-script will source the machine file, so values set there will override values set in `var_defns.sh` file. Each of those machine files also handles doing things like `ulimit` calls.

tests/WE2E/run_WE2E_tests.sh
```diff
@@ -729,17 +746,9 @@ VERBOSE=\"${VERBOSE}\""
...
-    elif [ "$MACHINE" = "WCOSS_DELL_P3" ]; then
-      pregen_basedir="/gpfs/dell2/emc/modeling/noscrub/UFS_SRW_App/FV3LAM_pregen"
-    elif [ "$MACHINE" = "HERA" ]; then
-      pregen_basedir="/scratch2/BMC/det/FV3LAM_pregen"
-    elif [ "$MACHINE" = "JET" ]; then
-      pregen_basedir="/mnt/lfs4/BMC/wrfruc/FV3-LAM/pregen"
-    elif [ "$MACHINE" = "CHEYENNE" ]; then
-      pregen_basedir="/glade/p/ral/jntp/UFS_CAM/FV3LAM_pregen"
-    else
+    pregen_basedir=${TEST_PREGEN_BASEDIR:-}
```
christinaholtNOAA, line 749: I've added variables in the machine config files that are prefixed with `TEST_` to indicate that those are only meant to be used in the test script here.

tests/WE2E/run_WE2E_tests.sh
```diff
@@ -837,37 +846,7 @@ SFC_CLIMO_DIR=\"${SFC_CLIMO_DIR}\""
...
-    if [ "$MACHINE" = "WCOSS_CRAY" ]; then
-      FIXLAM_NCO_BASEDIR="/gpfs/hps3/emc/meso/noscrub/UFS_SRW_App/FV3LAM_pregen"
-    elif [ "$MACHINE" = "WCOSS_DELL_P3" ]; then
-      FIXLAM_NCO_BASEDIR="/gpfs/dell2/emc/modeling/noscrub/UFS_SRW_App/FV3LAM_pregen"
-    elif [ "$MACHINE" = "HERA" ]; then
-      FIXLAM_NCO_BASEDIR="/scratch2/BMC/det/FV3LAM_pregen"
-    elif [ "$MACHINE" = "JET" ]; then
-      FIXLAM_NCO_BASEDIR="/mnt/lfs4/BMC/wrfruc/FV3-LAM/pregen"
-    elif [ "$MACHINE" = "CHEYENNE" ]; then
-      FIXLAM_NCO_BASEDIR="/needs/to/be/specified"
```
christinaholtNOAA, line 852: I did not include this placeholder in the cheyenne machine file.

tests/WE2E/run_WE2E_tests.sh
```diff
@@ -1026,51 +986,54 @@ EXTRN_MDL_FILES_LBCS=( $( printf "\"%s\" " "${EXTRN_MDL_FILES_LBCS[@]}" ))"
...
-      mrms_obs_dir="/scratch2/BMC/det/UFS_SRW_app/develop/obs_data/mrms/proc"
-      ndas_obs_dir="/scratch2/BMC/det/UFS_SRW_app/develop/obs_data/ndas/proc"
-      met_bin_exec="bin"
-    elif [ "$MACHINE" = "CHEYENNE" ]; then
-      met_install_dir="/glade/p/ral/jntp/MET/MET_releases/10.0.0"
-      metplus_path="/glade/p/ral/jntp/MET/METplus/METplus-4.0.0"
-      ccpa_obs_dir="/glade/p/ral/jntp/UFS_SRW_app/develop/obs_data/ccpa/proc"
-      mrms_obs_dir="/glade/p/ral/jntp/UFS_SRW_app/develop/obs_data/mrms/proc"
-      ndas_obs_dir="/glade/p/ral/jntp/UFS_SRW_app/develop/obs_data/ndas/proc"
-      met_bin_exec="bin"
```
christinaholtNOAA, line 1056: It's great that these are all included here, but there was no way to get this information when running a user-specified config outside the test framework. I have added these paths to the machine config file. I hope these are indeed the supported locations since the new behavior will be to use these paths by default on these machines for all users (test framework and otherwise).

tests/WE2E/run_WE2E_tests.sh
```diff
@@ -1211,7 +1167,7 @@ initial conditions (EXTRN_MDL_NAME_LBCS) combination:
 
     else 
 
-      if [ ! -d "${EXTRN_MDL_SYSBASEDIR_LBCS}" ]; then
+      if [ ! -d "$(dirname ${EXTRN_MDL_SYSBASEDIR_LBCS%%\$*})" ]; then
```
christinaholtNOAA: This variable could contain a template, so check the top level directory that does not include a `$`.

ush/bash_utils/boolify.sh
```diff
@@ -0,0 +1,69 @@
+#
```
christinaholtNOAA, line 1: I call this regularly from `ush/setup.sh` to clean up that script.

ush/bash_utils/set_bash_param.sh
```diff
@@ -119,17 +119,18 @@ Setting parameter \"$param\" in file \"$file\" to \"$value\" ..."
...
+  if [ $? = 0 ]  ; then
+    # The variable was found in existing config file. Replace its value.
+    $SED -i -r -e "s%${regex_search}%${regex_replace}%" "${file_full_path}"
 
-  $SED -i -r -e "s%${regex_search}%${regex_replace}%" "${file_full_path}"
+  else
+    # Append a new variable to the end of the config file..
+    echo $'\n'${param}=\'${value}\' >> ${file_full_path}
+
+  fi
```
christinaholtNOAA: Add functionality to add new parameters to the file when one doesn't already exist.

ush/config_defaults.sh
```diff
@@ -574,12 +580,12 @@ WRITE_DOPOST="FALSE"
...
+MET_INSTALL_DIR=""
 MET_BIN_EXEC="bin"
-METPLUS_PATH="/path/to/METPlus"
-CCPA_OBS_DIR="/path/to/observation-directory/ccpa/proc"
-MRMS_OBS_DIR="/path/to/observation-directory/mrms/proc"
-NDAS_OBS_DIR="/path/to/observation-directory/ndas/proc"
+METPLUS_PATH=""
+CCPA_OBS_DIR=""
+MRMS_OBS_DIR=""
+NDAS_OBS_DIR=""
```
christinaholtNOAA, line 588: These will now be set per machine, so remove the placeholder.

ush/machine/cheyenne.sh
```diff
@@ -0,0 +1,76 @@
...
+  case ${external_model} in
+
+    "FV3GFS")
+      location='/glade/p/ral/jntp/UFS_CAM/COMGFS/gfs.${yyyymmdd}/${hh}'
+      ;;
+    "*")
+      print_info_msg"\
+        External model \'${external_model}\' does not have a default
+      location on Cheyenne. Please set a user-defined file location."
+      ;;
```
christinaholtNOAA, line 24: In each of these machine files, I have made an effort to remove paths that no longer exist, or seem unsupported. I should also note that I really only have access to Jet/Hera to test some of them. Others, were more intuitive given the existence of mis-matched information in `ush/set_extrn_mdl_params.sh` and `ush/get_extrn_mdl_file_dir_info.sh`.

ush/machine/cheyenne.sh
```diff
@@ -0,0 +1,76 @@
...
+QUEUE_HPSS=${QUEUE_HPSS:-"regular"}
+QUEUE_FCST=${QUEUE_FCST:-"regular"}
+
+# UFS SRW App specific paths
+FIXgsm=${FIXgsm:-"/glade/p/ral/jntp/UFS_CAM/fix/fix_am"}
+FIXaer=${FIXaer:-"/glade/p/ral/jntp/UFS_CAM/fix/fix_aer"}
+FIXlut=${FIXlut:-"/glade/p/ral/jntp/UFS_CAM/fix/fix_lut"}
+TOPO_DIR=${TOPO_DIR:-"/glade/p/ral/jntp/UFS_CAM/fix/fix_orog"}
+SFC_CLIMO_INPUT_DIR=${SFC_CLIMO_INPUT_DIR:-"/glade/p/ral/jntp/UFS_CAM/fix/climo_fields_netcdf"}
+FIXLAM_NCO_BASEDIR=${FIXLAM_NCO_BASEDIR:-"/needs/to/be/specified"}
```
christinaholtNOAA, line 55: I'm not sure how important a placeholder default is here, but I guess I left it to mirror existing behavior.

ush/machine/orion.sh
```diff
@@ -0,0 +1,67 @@
...
+  external_model=${1}
+  external_file_fmt=${2}
+
+  case ${external_model} in
+
+    "*")
+      print_info_msg"\
+        External model \'${external_model}\' does not have a default
+      location on Orion. Please set a user-defined file location."
+      ;;
```
christinaholtNOAA, line 21: As best I could tell, there are no supported input data streams on disk on Orion.

ush/machine/jet.sh
```diff
@@ -0,0 +1,86 @@
...
+        "netcdf")
+          location='/public/data/grids/gfs/anl/netcdf/'
+          ;;
+      esac
+      ;;
+    "RAP")
+      location='/public/data/grids/rap/full/wrfprs/grib2'
+      ;;
+    "HRRR")
+      location='/public/data/grids/hrrr/conus/wrfprs/grib2'
```
christinaholtNOAA, line 32: I changed these from the non-existent RAPx/HRRRx paths to the current operational data stream. They will only be useful for near-real time. Otherwise, data will come from HPSS.

ush/setup.sh
```diff
@@ -132,416 +133,280 @@ fi
...
 #
 #-----------------------------------------------------------------------
 #
-if [ "$DEBUG" = "TRUE" ] && [ "$VERBOSE" = "FALSE" ]; then
+if [ "$DEBUG" = "TRUE" ]; then
   print_info_msg "
-Resetting VERBOSE to \"TRUE\" because DEBUG has been set to \"TRUE\"..."
-  VERBOSE="TRUE" 
+Setting VERBOSE to \"TRUE\" because DEBUG has been set to \"TRUE\"..."
+  VERBOSE="TRUE"
```
christinaholtNOAA, line 161: It doesn't actually matter if VERBOSE is false, just make sure that VERBOSE=TRUE if DEBUG=TRUE to clean up the logic.

ush/setup.sh
```diff
@@ -132,416 +133,280 @@ fi
...
 #
-DO_SKEB=$(echo_uppercase $DO_SKEB)
-if [ "${DO_SKEB}" = "TRUE" ] || \
-   [ "${DO_SKEB}" = "YES" ]; then
-  DO_SKEB="TRUE"
-elif [ "${DO_SKEB}" = "FALSE" ] || \
-     [ "${DO_SKEB}" = "NO" ]; then
-  DO_SKEB="FALSE"
-fi
+# USHDIR:
```
christinaholtNOAA, line 274: It could be better to view diffs in this file in "split" mode in GitHub. 

I moved the logic that sets up the directory structure to earlier in the script so we could actually find the machine file earlier. I checked (eyeballed) when I last merged that everything was the same in this script as it had been when I originally moved the code.

ush/setup.sh
```diff
@@ -132,416 +133,280 @@ fi
...
+property_name="local_path"
 #
-#-----------------------------------------------------------------------
+# Get the path to the workflow scripts
 #
-# Make sure that USE_FVCOM is set to a valid value and assign directory
-# and file names.
-# 
-# Make sure that FVCOM_WCSTART is set to lowercase "warm" or "cold"
+external_name=regional_workflow
```
christinaholtNOAA, line 328: Here, I added the necessary bits to achieve @gsketefian's TODO comment in getting the regional_workflow path from Externals.cfg.

christinaholtNOAA (commented):
Just a few comments.

christinaholtNOAA:
Follow up on the test suite. Most tests passed. These failed on the run_fcst task:

```
grid_RRFS_AK_3km_ics_RAP_lbcs_RAP_suite_GSD_SAR
grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v16
grid_RRFS_CONUS_25km_ics_HRRR_lbcs_HRRR_suite_GSD_SAR
grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_GSD_SAR
grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha
nco_grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
```

These failed on make_ics and make_lbcs:
```
grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_GSD_SAR
grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_HRRR
grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_RRFS_v1beta
```

Can anyone confirm whether these are expected to run on Hera with the top of develop?

ush/machine/jet.sh
gsketefian: Ok, sounds good.

ush/machine/orion.sh
gsketefian: I think you're right.

ush/machine/cheyenne.sh
gsketefian: Sounds good.

ush/machine/cheyenne.sh
```diff
@@ -0,0 +1,76 @@
...
+
```
gsketefian, line 56: Do we want these `RUN_CMD_...` variables to be user-specifiable as well (in which case they'd have to be defined in `config_defaults.sh` and then (re)set here)?  I'm not sure...  Users can always change it here, but this file is under version control whereas `config.sh` isn't.

scripts/exregional_make_grid.sh
gsketefian: What prevents sourcing these files in say `setup.sh`?  That way the values in `var_defns.sh` will be "correct", i.e. they will be the values that were actually used in the experiment.  Otherwise the user may be confused.  I saw some variables in the machine files (`RUN_CMD_...`, `TEST_...`) that are not in `config_defaults.sh`, but we can include them.  Otherwise, I did not see anything that would prevent sourcing of machine files during experiment generation instead of in the ex-scripts (but I may have missed something).

ush/setup.sh
gsketefian: Nice, thanks!

ush/setup.sh
```diff
@@ -3141,6 +2585,7 @@ FVCOM_FILE='${FVCOM_FILE}'
 #
 #-----------------------------------------------------------------------
 #
+NCORES_PER_NODE='${NCORES_PER_NODE}'
```
gsketefian, line 2588: I removed this `NCORES_PER_NODE` line in PR #650 because `NCORES_PER_NODE` is a primary variable that is already defined in the first part of `var_defns.sh`.  So this line just duplicates a previous line in `var_defns.sh` and so is unnecessary -- unless something changed that I didn't notice.

scripts/exregional_make_grid.sh
gsketefian: I just noticed that you do source the machine file in `setup.sh`.  That means the values in the machine files get recorded in `var_defns.sh`, which is what we want.  Then my question is, why do you need to source it again here?  What changes that makes this necessary?

christinaholtNOAA: This file is sourced in setup.sh, for sure. I didn't duplicate the variables in var_defns.sh for clarity since I was assuming we'd source it at run time. When we source it at run time, we can also include lines that handle ulimit that we may not want to source from `var_defns.sh` every time it's sourced.

ush/setup.sh
christinaholtNOAA: I can remove this line and check that it still gets set as appropriate. Just so I make sure on the rule for this one...if it has a default in `config_defaults.sh` the appropriate user-specified value gets added to the first section of `var_defns.sh`?

scripts/exregional_make_grid.sh
christinaholtNOAA: A thought just occurred to me. Keeping these sorts of machine-specific items out of the var_defns.sh file could help us to achieve more platform independence in the future. For example, if these files are the only ones that contain the platform information, we could theoretically move an experiment directory to a different machine (say I run out of Jet resources and I wanted to move my experiment to run on Hera) and be able to run without any changes to the var_defns.sh file. 

That's potentially a long way off, and definitely not something I'm suggesting we open a current issue for, but keeping the machine-specific settings separate from the App-specific settings has its benefits.

gsketefian:
> Follow up on the test suite. Most tests passed. These failed on the run_fcst task:
> 
> ```
> grid_RRFS_AK_3km_ics_RAP_lbcs_RAP_suite_GSD_SAR
> grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v16
> grid_RRFS_CONUS_25km_ics_HRRR_lbcs_HRRR_suite_GSD_SAR
> grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_GSD_SAR
> grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha
> nco_grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
> ```
> 
> These failed on make_ics and make_lbcs:
> 
> ```
> grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_GSD_SAR
> grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_HRRR
> grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_RRFS_v1beta
> ```
> 
> Can anyone confirm whether these are expected to run on Hera with the top of develop?

The three NAM tests have never worked, I think.  In fact, it might be good to remove "NAM" as an external model in `valid_param_vals.sh` (maybe I should never have included it; I know, bad coding practice!).  So yes, these are expected to fail.

The Alaska grid may have worked a long time ago but broke at some point and was difficult to debug.  Also expected to fail.

The remaining tests (all on RRFS_CONUS_25km grid) are:
```
grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v16
grid_RRFS_CONUS_25km_ics_HRRR_lbcs_HRRR_suite_GSD_SAR
grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_GSD_SAR
grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha
nco_grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2

```
I don't know why these are failing.  Do they at least run for a while and then crash, or do they crash even before the time-integration in the forecast kicks off?  If they run for a while at least, then I doubt the failures are related to this PR.  Whether it's worth debugging these or getting rid of them (because they may be irrelevant/outdated) is a decision we have to make as a group. 

Btw, I just successfully ran `grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha` but for only 6 hours; the test goes on for 24 hours, so yours probably ran for 6 hrs and then crashed.  If so, the failure was very likely there beforehand.

ush/machine/cheyenne.sh
christinaholtNOAA: If they were not already user-configurable, I'd prefer not to add that feature in this PR. What we have here are the best-practice run commands for these supported platforms, and it's not really recommended to use others. If that changes, I'd assume that code maintainers would update these files.

These commands are user-configurable in so much as we are allowing a user to provide their own machine config file. I suppose if they wanted to they could provide an alternative config file for an already supported machine to change that value for their own testing purposes.

christinaholtNOAA:
Following up on the short list of failed forecasts:

```grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v16
grid_RRFS_CONUS_25km_ics_HRRR_lbcs_HRRR_suite_GSD_SAR
grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_GSD_SAR
grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha
nco_grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
```

- grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v16 was canceled due to wall clock
- grid_RRFS_CONUS_25km_ics_HRRR_lbcs_HRRR_suite_GSD_SAR suffered from the dreaded `AVOST IN VILKA` messages before crashing with `unrealistic surface pressure` messages
- grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_GSD_SAR (same as last)
- grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha no `AVOST IN VILKA` messages, but failed with `unrealistic surface pressure` messages
- nco_grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2 failed before any forecast steps with `saturation vapor pressure table overflow` fatal error

ush/setup.sh
gsketefian: Yes, that's right.  You can check any of the `var_defns.sh` files in the tests you ran, and they should contain two definitions for `NCORES_PER_NODE` that are identical.  If you can just run one test with this line removed to make sure the second definition is removed, that would be sufficient.

ush/setup.sh
gsketefian: Ok.

ush/machine/cheyenne.sh
gsketefian: Ok, that's fine,

ush/setup.sh
christinaholtNOAA: There are two and they are identical. I can remove this line.

tests/WE2E/run_WE2E_tests.sh
gsketefian: Ok, sounds good.

tests/WE2E/run_WE2E_tests.sh
gsketefian: Yes, I agree these should be moved.  I think they are the supported locations, but we'll find out!

ush/bash_utils/boolify.sh
gsketefian: Sounds good.  I think that's what I was doing in PR #580.

ush/config_defaults.sh
gsketefian: Ok.

ush/setup.sh
gsketefian: Ok.  Well if most of the tests passed, you must have eyeballed it correctly!

ush/machine/cheyenne.sh
gsketefian: Sounds ok to me.

tests/WE2E/run_WE2E_tests.sh
gsketefian: Ok.

scripts/exregional_make_grid.sh
gsketefian: So the only reason to source the machine files in the ex-scripts is to call `ulimit`, etc, right?  The variable values shouldn't change from those in `var_defns.sh` since the machine file was used in `setup.sh` to set the values in `var_defns.sh` in the first place -- yes?  I just want to make sure that what's in `var_defns.sh` doesn't conflict with what is actually used in the ex-script.

Yeah, we can remove the machine info from `var_defns.sh` at some point so the experiment is more portable.  Then I suppose we won't have to source the machine file in `setup.sh`, just in the ex-scripts.

scripts/exregional_make_grid.sh
gsketefian: Sounds good.

scripts/exregional_make_grid.sh
```diff
@@ -102,56 +102,25 @@ case "$MACHINE" in
...
     ;;
 
 esac
+
+if [ -z ${RUN_CMD_SERIAL:-} ] ; then
+  print_err_msg_exit " \
+  Run command was not set in machine file. \
+  Please set RUN_CMD_SERIAL for your platform"
+else
+  RUN_CMD_SERIAL=$(eval echo ${RUN_CMD_SERIAL})
```
gsketefian, line 120: Do you need the `echo`?  What happens if you just do `eval ${RUN_CMD_SERIAL}`?

ush/setup.sh
gsketefian: Great, thanks!

scripts/exregional_make_grid.sh
gsketefian: Ok.

tests/WE2E/run_WE2E_tests.sh
gsketefian: Ok, sounds good.

ush/bash_utils/set_bash_param.sh
gsketefian: This is the only thing I'm a bit wary of.  My intent when I created this function was only to replace the value of an existing variable, not create a new one if the variable doesn't exist.  Adding this can sneak in a new variable into a file that should really be flagged and cause an error, e.g. when a user has a typo in a variable's name.  

I saw in `run_WE2E_tests.sh` that this is used like this (one occurrence for ICs, one for LBCs):
```
set_bash_param "${expt_config_fp}" \
  "EXTRN_MDL_SYSBASEDIR_LBCS" ${EXTRN_MDL_SYSBASEDIR_LBCS}
```
Isn't `EXTRN_MDL_SYSBASEDIR_LBCS` already present in the configuration file so you just have to replace its value?  Or did the introduction of the machine config files change this?

gsketefian:
> Following up on the short list of failed forecasts:
> 
> ```
> grid_RRFS_CONUS_25km_ics_HRRR_lbcs_HRRR_suite_GSD_SAR
> grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_GSD_SAR
> grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha
> nco_grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
> ```
> 
> * grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v16 was canceled due to wall clock
> * grid_RRFS_CONUS_25km_ics_HRRR_lbcs_HRRR_suite_GSD_SAR suffered from the dreaded `AVOST IN VILKA` messages before crashing with `unrealistic surface pressure` messages
> * grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_GSD_SAR (same as last)
> * grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha no `AVOST IN VILKA` messages, but failed with `unrealistic surface pressure` messages
> * nco_grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2 failed before any forecast steps with `saturation vapor pressure table overflow` fatal error

Thanks for that info.  Some comments and questions:
* `grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v16`:  Does the forecast start and just not complete before time is up, or does it just hang for the whole time?
* `grid_RRFS_CONUS_25km_ics_HRRR_lbcs_HRRR_suite_GSD_SAR`:  I don't know if the `GSD_SAR` suite is still used.  If so, we should ask Tanya for help with `AVOST IN VILKA`; otherwise we can remove this test.
* `grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_GSD_SAR`: Same as last. 
* `grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha`:  It would be great if this one succeeded since I know some DTC projects and university researchers are still using that suite.
* `nco_grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2`:  This should definitely work, and at some point it did.

Anyway, none of these failures should prevent us from merging this PR.

scripts/exregional_make_grid.sh
christinaholtNOAA: I think we will still want to source the machine file in `setup.sh` since we may want to do some configuration checks earlier than run-time, but I don't think that we would have any issues with diffs between the `var_defns.sh` file and the machine config.

ush/bash_utils/set_bash_param.sh
christinaholtNOAA: This change in behavior for `set_bash_param` is to support the use of machine-specific template locations specified in a machine config file. It was only used in the WE2E scripts as best I could tell, and adds the desired path to the experiment config file _before_ it runs `generate_FV3LAM_wflow.sh`. 

The behavior of how the external model file directories are set for the testing framework does indeed change with this implementation because the defaults of `EXTRN_MDL_SYSBASEDIR_LBCS ` and `EXTRN_MDL_SYSBASEDIR_LBCS` are empty strings. Before this capability to set them in the machine config, the string remained empty at this point and they were set later in the configuration layer. Now, these variables are non-empty so need to be added in the experiment config. 

I suppose the two variables don't really need to be added here since they would be added in the configuration layer when it sources the machine config and sets it up, but we would need to handle the case where the variables are not empty strings and not set to `set_to_non_default_location_in_testing_script`. 

Which way would you prefer to go with that?

scripts/exregional_make_grid.sh
christinaholtNOAA: You do need the echo, otherwise it wants to execute the command in the subshell when it's expanded:
```
hfe01:~ $ nprocs=7
hfe01:~ $ RUN_CMD_UTILS='mpirun -np $nprocs'
hfe01:~ $ test_cmd=$(eval echo $RUN_CMD_UTILS)
hfe01:~ $ echo $test_cmd 
mpirun -np 7
hfe01:~ $ test_cmd=$(eval $RUN_CMD_UTILS)
-bash: mpirun: command not found
```

christinaholtNOAA:
- `grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v16`: My apologies. This was not a failure on wallclock. It seg faulted after completing 93 time steps. 
- GSD_SAR was recently removed in the model because nobody was using it.  See the changes in [this PR](https://github.com/ufs-community/ufs-weather-model/commit/56f3483dacf9bb11597dde731a726cf643049b6d). It's probably fair to remove these tests, even before the model hash is updated...
- `grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha` do we expect that this one should work at 25km, though?

scripts/exregional_make_grid.sh
gsketefian: Ok.

gsketefian:
> * grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha

We can bring up the `GSD_SAR` issue today at the meeting.  If most agree, we can remove that suite from the tests.

For `grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v16`, maybe the `GSMGFS` data are too old to work with `GSD_SAR`, e.g. they're missing certain fields.  We can bring this up today as well and see what people say.  I would think we would want to still support `GSMGFS` as an external model, but maybe not with `GFS_v16`.

Good point about `RRFS_v1alpha` not being suitable for a 25km grid.  But it would be nice to have a coarse-grid test for this suite so we don't have to run the 3km grid to check that the suite is still working.  I ran `grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha` a few days ago but only for 6 hours, and that completed.  One solution is to just change the forecast length in the test from 24hr to 6hr.  We can discuss during the meeting as well if we have time, otherwise I say just shorten the forecast length.

ush/bash_utils/set_bash_param.sh
gsketefian: It's slowly coming back to me that I did something to get rid of that `set_to_non_default_location_in_testing_script` in PR #580, but I can't remember what and have to go look.  For now, do you mind making a new function in `bash_utils` directory, say `add_bash_param`, that just appends a (new) variable to a (bash) script and just use that?  Thanks.

scripts/exregional_make_grid.sh
gsketefian: Ok.  Just realized you're resetting `RUN_CMD_SERIAL` here, as opposed to using it in an `eval` with the rest of the arguments for the command, e.g.
`
eval $RUN_CMD_SERIAL ${exec_fp} ${rgnl_grid_nml_fp} || ...
`
(which would not be done here but later below).  That's how I did tests for PR #650, This is ok though.

gsketefian:
@christinaholtNOAA I created an issue (#668) for removing GSD_SAR.

gsketefian:
@christinaholtNOAA Approving now.  If you can address the `set_param.sh` issue one way or another before merging, that would be great.  Thanks.

gsketefian (approved):

ush/bash_utils/set_bash_param.sh
christinaholtNOAA: @gsketefian Would it work for me to change the `set_bash_param` functionality back to what it was and change the logic in `run_WE2E_tests.sh` to only update the value when the value of `EXTRN_MDL_SYSBASEDIR_*` is set to the special string `set_to_non_default_location_in_testing_script`? 

Since we don't want to put platform-specific paths in the tests anyway, that might suffice for maintaining the current (and desired) capabilities.

christinaholtNOAA: @gsketefian I went ahead and made the change, and pushed it. Let me know if this is acceptable.

gsketefian: Ok, that works.  Thanks.

christinaholtNOAA:
@gsketefian I'm fine with just including the first couple of description paragraphs. I don't know what you mean by "updating" the PR message.

gsketefian:
> @gsketefian I'm fine with just including the first couple of description paragraphs. I don't know what you mean by "updating" the PR message.

Oh I just meant making sure the original message of the PR is updated to reflect changes that were made during the review process (e.g. if new tests were run, etc).  Merging now with just the couple of description paragraphs.

scripts/exregional_make_ics.sh
```diff
@@ -90,59 +90,33 @@ case "$MACHINE" in
...
-
-  "ODIN")
-    APRUN="srun"
-    ;;
-
-  "CHEYENNE")
-    nprocs=$(( NNODES_MAKE_ICS*PPN_MAKE_ICS ))
-    APRUN="mpirun -np $nprocs"
+  *)
+    source ${MACHINE_FILE}
```
gsketefian, line 102: @christinaholtNOAA I just noticed while working on another fix that MACHINE_FILE does not get sourced for WCOSS_*.  I would think it should be, or is that intentional?  (Same in most/all of the other ex-scripts.)  Thanks.

PR #667, opened 2022-01-14
Add stand-alone Python data ingest tool.

christinaholtNOAA:
## DESCRIPTION OF CHANGES: 
This tool is a stand alone script that can be used to grab external model files from a variety of sources including the internet, disk, and HPSS. The known locations for BDP AWS, HPSS, and NOMADS have been added here for demo purposes. It will also accept command line options to support user-defined disk locations with the caveat that file names provided on the command line must match the formatting necessary in the config file. A list of options for data_stores can be provided, and each will be checked until the required files are found. The list of known locations could easily be extended to add observation file locations for verification and data assimilation (future) purposes.

This tool has not yet been hooked into the workflow. I would like to receive feedback on the tool itself before moving forward with replacing existing functionality in the run scripts. 

I envision that this tool would be called from a job script, so am unsure whether it makes sense to store it under ush or scripts.

Example of pulling grib2 files from hpss or aws for FV3GFS:
```
    python retrieve_data.py 
        --anl_or_fcst fcst 
        --cycle_date 2022010212 
        --data_stores hpss aws 
        --external_model FV3GFS 
        --file_type grib2
        --fcst_hrs 0 3 
        --output_path test_output_loc 
        --config ./templates/data_locations.yml 
        --debug 
        --file_type grib2
```
Similarly, if you wanted to check disk first:
```
    files=settings="
        'grib2':
           'fcst': gfs.t{hh}z.pgrb2.0p25.f{fcst_hr:03d}
    "
    python retrieve_data.py 
        --anl_or_fcst fcst 
        --cycle_date 2022010212 
        --data_stores hpss aws 
        --external_model FV3GFS 
        --file_type grib2
        --fcst_hrs 0 3 
        --output_path test_output_loc 
        --config ./templates/data_locations.yml 
        --debug 
        --file_type grib2
        --input_file_paths /scratch1/NCEPDEV/rstprod/com/gfs/prod/gfs.{yyyymmdd}/{hh}/atmos
        --file_name ${files}
```

## TESTS CONDUCTED: 
I've run many tests locally to try out pulling from various data sources. NOMADS never seems to work (nor does it when I run from the existing regional_workflow scripts).

## DEPENDENCIES:
Requires the packages from existing regional_workflow conda packages.

## DOCUMENTATION:
I'm not sure where it makes sense to document this tool, or if it does, since it replaces existing docs.

### DIFF
--- a/ush/retrieve_data.py
+++ b/ush/retrieve_data.py
@@ -86,6 +86,11 @@ def download_requested_files(cla, data_store, store_specs):
                     if unavailable.get(data_store) is None:
                         unavailable[data_store] = []
                     unavailable[data_store].append(target_path)
+                    os.chdir(orig_path)
+                    # Returning here assumes that if the first file
+                    # isn't found, none of the others will be. Don't
+                    # waste time timing out on every requested file.
+                    return unavailable
     os.chdir(orig_path)
     return unavailable
 
--- a/ush/retrieve_data.py
+++ b/ush/retrieve_data.py
@@ -156,6 +156,32 @@ def fill_template(template_str, cycle_date, fcst_hr=0):
         yyyymmddhh=cycle_date.strftime('%Y%m%d%H'),
         )
 
+def hsi_single_file(file_path, mode='ls'):
+
+    ''' Call hsi as a subprocess for Python and return information about
+    whether the file_path was found.
+
+    Arguments:
+        file_path    path on HPSS
+        mode         the hsi command to run. ls is default. may also
+                     pass "get" to retrieve the file path
+
+    '''
+    cmd = f'hsi {mode} {file_path}'
+
+    logging.info(f'Running command \n {cmd}')
+    try:
+        output = subprocess.run(cmd,
+                                check=True,
+                                shell=True,
+                                )
+    except subprocess.CalledProcessError as excep:
+        logging.warning(f'{file_path} is not available!')
+        return ''
+    #logging.debug(f'OUTPUT for hsi is: \n {output.stderr.decode("utf-8") }')
+
+    return file_path
+
 def htar_requested_files(cla, store_specs):
 
     ''' This function interacts with the "hpss" protocol in a
@@ -198,24 +224,13 @@ def htar_requested_files(cla, store_specs):
         file_path = os.path.join(archive_path, archive_file_names[0])
         file_path = fill_template(file_path, cla.cycle_date)
 
-        cmd = f'hsi ls {file_path}'
-        try:
-            output = subprocess.run(cmd,
-                                    capture_output=True,
-                                    check=True,
-                                    shell=True,
-                                    )
-        except subprocess.CalledProcessError as excep:
-            logging.warning(f'{file_path} is not available!')
-            continue
-        logging.debug(f'{output}')
-        logging.debug(f'OUTPUT for hsi is: \n {output.stderr.decode("utf-8") }')
-        if output.returncode == 0:
-            existing_archive = file_path
+        existing_archive = hsi_single_file(file_path)
+
+        if existing_archive:
             logging.info(f'Found HPSS file: {file_path}')
             break
 
-    if existing_archive is None:
+    if not existing_archive:
         logging.warning(f'No archive files were found!')
         return unavailable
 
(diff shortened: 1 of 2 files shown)


dmwright526:
@christinaholtNOAA, Sorry for taking so long to look at this! A few comments/clarifications on the script:

-  Could you clarify the difference in the flags "--config" and "--input_file_paths"? Does "--input_file_paths" allow for not using the .yaml template?

-  Based upon the answer to the previous bullet, "--config" might need to be changed to a required input variable so better error messaging could be displayed.

- This might come more when integrating this into get_extrn* routines, but this routine will overwrite existing data in a directory. For example, the file name for HRRR from hpss does not have any date associated with the name (hrrr.t12z.wrfnatf00.grib2), so if you try to pull down more than 24 hours worth analysis data to use as boundary conditions it will overwrite data. A solution to this might be to attach the cycle_date to the start of the file name. Just a thought.

christinaholtNOAA:
@dmwright526 Thanks for the review!

Here are a couple of responses:

> Could you clarify the difference in the flags "--config" and "--input_file_paths"? Does "--input_file_paths" allow for not using the .yaml template?

Great question. That isn't entirely clear. The `--config` flag is meant to take the `data_locations.yml` path so that the script can access known locations and file naming conventions. The `--input_file-paths` is meant to take specific paths to files on disk. This will be the mechanism for the scripts to maintain much of their existing behavior -- setting up known paths for machines (moving to config files in PR #666) or configuring it for a user-defined path. This last one is meant to be used in conjunction with the `--file_names` flag so that users can ingest files named anything...Jet for example has some different naming conventions than other platforms for historical reasons.

> Based upon the answer to the previous bullet, "--config" might need to be changed to a required input variable so better error messaging could be displayed.

I think this is a great suggestion. I think in most situations, it will be helpful to have this file information present, and isn't too expensive to load it when the information may not be needed.

> This might come more when integrating this into get_extrn* routines, but this routine will overwrite existing data in a directory. For example, the file name for HRRR from hpss does not have any date associated with the name (hrrr.t12z.wrfnatf00.grib2), so if you try to pull down more than 24 hours worth analysis data to use as boundary conditions it will overwrite data. A solution to this might be to attach the cycle_date to the start of the file name. Just a thought.

That is correct. It will overwrite files on disk (or continue downloading a partial file with wget based on the flags that are currently hard-coded). 

The output directory is set as part of the regional_workflow tasks that grab data. My plan is to put that burden on the caller such that this script can be as generic as possible. In other words, this tool should grab expected data, and put it where the user asks it to go. I think the caller should make sure it's asking for the right thing. I will make sure to provide the appropriate checks and logging in the appropriate callers to mitigate that concern. I think many of them already exist in some form or another.

One limitation of the script is that it does not loop over output directories. In the context of the regional_workflow, I believe all of the data pulling tasks (for external model data) follow this strategy of pulling only files from a single cycle in a single task. 


Let me know if that dredged up some additional questions...I'm sure it must have. Thanks again!

dmwright526:
Thank you for the detailed response, @christinaholtNOAA ! This all makes sense to me. 

I think the final component, as you mentioned, is the documentation (which most of it is already self contained in retrieve_data.py -h). I think adding a quick description in a comment block at the top of the file would be good just to keep documentation with the file. Most of what you have written here for this PR description and in the file could be copied and pasted to give the user a starting point and for use in the User's Guide. One thing that might be good to highlight is the case sensitivity of the variables, specifically external_model.

I did test out the script on Cheyenne and it will hang if you pass hpss to it, but that would be another check to make by the caller.

Were you looking to merge this before integration?

christinaholtNOAA:
@dmwright526 I can certainly add the documentation as you suggest. I will attempt to get to that between AMS talks this week, along with a bit of a helper to allow fewer case-sensitive concerns. I can also add a check for the compatibility of hpss calls and the machine you're on.

I would like to merge this one before integrating it, if that's okay with folks.

dmwright526:
@christinaholtNOAA Sounds good to me.

gsketefian:
@christinaholtNOAA I am looking through the scripts now but will approve since @dmwright526 already reviewed them.  You can merge whenever you and @dmwright526 feel it's good to go.

gsketefian (approved):

christinaholtNOAA:
Thanks! I am working on addressing the comments now and will hopefully push them in the next hour or so.

christinaholtNOAA:
@dmwright526 I believe I have addressed the concerns that you highlighted. Please take a look when you have time and let me know if that is sufficient. I'm happy to make further adjustments.

dmwright526:
@christinaholtNOAA This all looks great to me! I will go ahead and merge.

PR #674, opened 2022-02-03
Adding python utilities in preparation for providing a python workflow generation layer

danielabdi-noaa:
## DESCRIPTION OF CHANGES: 
This PR adds python utility functions that mirror those under `ush/bash_utils` in preparation for an ensuing PR
that aims to replace the "workflow generation layer" of regional_workflow with python instead of shell scripts.
These changes should not affect existing code in any way because it is a simple addition of utility functions
in a separate folder python_utils

## TESTS CONDUCTED: 

Test to ensure sanity of existing regional_workflow is not necessary for this particular PR, however, the python_utils themselves could be tested with unit-tests from the `ush` directory

```
$ python3 -m unittest -b python_utils/test_python_utils.py
.....................
----------------------------------------------------------------------
Ran 17 tests in 0.219s

OK
```

## DEPENDENCIES:
No dependencies

## DOCUMENTATION:
None

## ISSUE (optional): 
None

## CONTRIBUTORS (optional): 
Suggestions from @christinaholtNOAA


ush/python_utils/check_for_preexist_dir_file.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+
+    check_var_valid_value(method, ['delete', 'rename', 'quit'])
+
+    if os.path.exists(path):
+        if method == 'delete':
+            rm_vrfy(' -rf ', path)
+        elif method == 'rename':
+            now = datetime.now()
+            d = now.strftime("_%Y%m%d_%H%M%S")
```
christinaholtNOAA, line 27: I think this changes the behavior of the "rename" functionality. I don't mind the this new naming convention, where the original named directories with a suffix `_old001`, for example.

ush/python_utils/check_var_valid_value.py
```diff
@@ -0,0 +1,20 @@
...
+
+    Args:
+        var: the variable
+        values: list of valid values
+    Returns:
+        True: if var has valid value, exit(1) otherwise
+    """
+
+    if var not in values:
+        if not err_msg:
```
christinaholtNOAA: I'm not entirely sure if it really matters here, but should `err_msg` be checked for not being `None` specifically? If other err_msg values that evaluate to False are passed, it would be overwritten. I'm not really sure whether that is a legit use case or not, though.

ush/python_utils/config_parser.py
```diff
@@ -0,0 +1,134 @@
...
+    return None 
+
+def load_shell_config_complex(config_file):
+    """ If the shell script has more logic than just setting variables,
+    we source the config script in a subshell and gets the variables it set
+    """
+
+    # Save env vars before and after sourcing the scipt and then
+    # do a diff to get variables specifically defined/updated in the script
+    # Method sounds brittle but seems to work ok so far
```
christinaholtNOAA, line 25: Is there any reason not to always do this method? Is `load_shell_config` still around to match the bash util?

ush/python_utils/create_symlink_to_file.py
```diff
@@ -0,0 +1,49 @@
...
+        target: target file
+        symlink: symbolic link to target file
+        relative: optional argument to specify relative symoblic link creation
+    Returns:
+        None
+    """
+
+    print_input_args(locals())
+
+    if target == None:
```
christinaholtNOAA: I've seen best practice dictate using `is None` rather than `== None` because of how some classes can be handled. I doubt this is necessary a problem here, though.

ush/python_utils/run_command.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python3
+
+import subprocess
+
+def run_command(cmd):
+    """ Run systme command in a subprocess
```
christinaholtNOAA: Typo: systme -> system

ush/python_utils/environment.py
```diff
@@ -0,0 +1,240 @@
...
+    """
+    try:
+        v = datetime.strptime(s, "%Y%m%d%H%M")
+        return v
+    except:
+        try:
+            v = datetime.strptime(s, "%Y%m%d")
+            return v
+        except:
+            return None
```
christinaholtNOAA, line 25: Is it okay to "fail" without note here?

ush/python_utils/environment.py
```diff
@@ -0,0 +1,240 @@
...
+
+def date_to_str(d,short=False):
+    """ Get string from python datetime object.
+    By default it converts to YYYYMMDDHHMM format unless
+    told otherwise with `short` or HHMM=0
+
+    Args:
+        d: datetime object
+    Returns:
+        string in YYYYMMDDHHMM format
```
christinaholtNOAA: Maybe `string in YYYYMMDDHHMM or YYMMDD format`? Just in case folks skim over the actual description.

ush/python_utils/environment.py
```diff
@@ -0,0 +1,240 @@
...
+    """
+    if short or (d.hour == 0 and d.minute == 0):
+        v = d.strftime("%Y%m%d")
+    else:
+        v = d.strftime("%Y%m%d%H%M")
+    return v
+
+def str_to_type(s, just_get_me_the_string = False):
+    """ Check if the string contains a float, int, boolean, or just reguar string.
```
christinaholtNOAA: reguar -> regular

ush/python_utils/environment.py
```diff
@@ -0,0 +1,240 @@
...
+        a float, int, boolean, date, or the string itself when all else fails
+    """
+    s = s.strip('"\'')
+    if not just_get_me_the_string:
+        if s.lower() in ['true','yes','yeah']:
+            return True
+        elif s.lower() in ['false','no','nope']:
+            return False
+        v = str_to_date(s)
+        if v:
```
christinaholtNOAA: Here it might be good to use `if v is not None` since `str_to_date` returns None when it doesn't return a date time object.

ush/python_utils/environment.py
```diff
@@ -0,0 +1,240 @@
...
+
+    Args:
+        v: a variable of the above types
+    Returns:
+        a string
+    """
+    if isinstance(v,bool):
+        return ("TRUE" if v else "FALSE")
+    elif isinstance(v,int) or isinstance(v,float):
+        return str(v)
```
christinaholtNOAA: Do we need this branch explicitly since it does the same thing as the final return statement, i.e. `str(v)`?

ush/python_utils/environment.py
```diff
@@ -0,0 +1,240 @@
...
+            if "=" in itm:
+                idx = itm.find("=")
+                itm = itm[idx+1:]
+            lst.append(str_to_type(itm))
+        return lst
+    else:
+        return str_to_type(v)
+
+def set_env_var(param,value):
+    """ Set an environement variable
```
christinaholtNOAA: environement -> environment

ush/python_utils/environment.py
```diff
@@ -0,0 +1,240 @@
...
+        Returns either a string, list of strings or None
+    """
+
+    if not param in os.environ:
+        return None
+    else:
+        value = os.environ[param]
+        return str_to_list(value)
+
+def import_vars(dictionary=os.environ, target_dict=None, env_vars=None):
```
christinaholtNOAA: I am worried about using a dict in the instantiation of the `dictionary` variable default. [Check out this weird Pythonism here](https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html).

The same is true for export_vars below.

ush/python_utils/environment.py
```diff
@@ -0,0 +1,240 @@
...
+    """ Import all (or select few) environment/dictionary variables as python global 
+    variables of the caller module. Call this function at the beginning of a function
+    that uses environment variables.
+
+    Note that for ready-only environmental variables, calling this function once at the
```
christinaholtNOAA: ready-only -> read-only?

ush/python_utils/run_command.py
```diff
@@ -0,0 +1,25 @@
...
+    Returns:
+        Tuple of (exit code, std_out, std_err)
+    """
+    proc = subprocess.Popen(cmd,
+                            stdout=subprocess.PIPE,
+                            stderr=subprocess.PIPE,
+                            shell=True,
+                            universal_newlines=True)
+
+    std_out, std_err = proc.communicate()
```
christinaholtNOAA, line 19: What happens here if `cmd` fails?

ush/python_utils/is_array.py
```diff
@@ -0,0 +1,14 @@
...
+
+def is_array(var):
+    """ Checks if variable is an instance of a list
+
+    Args:
+        var: the variable
+    Returns:
+        True if variable is a list
+    """
+    if type(var) == list:
```
christinaholtNOAA: I'm curious about the use of `type` here instead of `isinstance()`. Is there something I'm missing, or is it just a different way of doing the same thing? Maybe then it's just a one-liner:

``` return isinstance(var, list)```

File: ush/python_utils/is_element_of.py [hunk 16]
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env python3
+
+def is_element_of(elem, arr):
```
christinaholtNOAA: Do we need this one for Python given that this is a super easy, standard built-in capability of Python?

ush/python_utils/process_args.py
```diff
@@ -0,0 +1,64 @@
...
+              arg_val = \"{arg_value}\"\n''')
+        check_var_valid_value(arg_name, valid_arg_names, err_msg)
+
+        idx = valid_arg_names.index(arg_name)
+        if values_args[idx] != None:
+            print_err_msg_exit(f'''
+                The current argument has already been assigned a value:
+                  arg_name = \"{arg_name}\"
+                  key_value_pair = {kwargs}
+                Please assign values to arguments only once on the command line.''')
```
christinaholtNOAA, line 60: Does this one also need `dedent`?

ush/python_utils/set_file_param.py
```diff
@@ -0,0 +1,71 @@
...
+        regex_replace = f"\\1{value}\\3"
+    elif file_ == DIAG_TABLE_FN:
+        regex_search = f"(.*)(<{param}>)(.*)"
+        regex_replace = f"\\1{value}\\3"
+    elif file_ == MODEL_CONFIG_FN:
+        regex_search = f"^(\s*{param}:\s*)(.*)"
+        regex_replace = f"\\1{value}"
+    elif file_ == GLOBAL_VAR_DEFNS_FN:
+        regex_search = f'(^\s*{param}=)(\".*\")?([^ \"]*)?(\(.*\))?(\s*[#].*)?'
+        regex_replace = f"\\1{value}\\5"
```
christinaholtNOAA: Can many of these be removed now? I don't _think_ several of these files are currently supported in this way.

ush/python_utils/test_python_utils.py
```diff
@@ -0,0 +1,132 @@
...
+    def test_change_case(self):
+        self.assertEqual( uppercase('upper'), 'UPPER' )
+        self.assertEqual( lowercase('LOWER'), 'lower' )
+    def test_check_for_preexist_dir_file(self):
+        cmd_vrfy('mkdir -p dir')
+        self.assertTrue( os.path.exists('dir') )
+        check_for_preexist_dir_file('dir-1', 'rename')
+        dirs = glob.glob('dir*')
+        self.assertEqual( len(dirs), 1)
+        rm_vrfy('-rf dir*')
```
christinaholtNOAA: Should we also check for the existence of 2 dirs if the first directory already exists?

ush/python_utils/test_python_utils.py
```diff
@@ -0,0 +1,132 @@
...
+    def test_check_var_valid_value(self):
+        self.assertTrue( check_var_valid_value('rice', [ 'egg', 'spam', 'rice' ]) )
```
christinaholtNOAA, line 32: Maybe also one `assertFalse` here?

ush/python_utils/test_python_utils.py
```diff
@@ -0,0 +1,132 @@
...
+    def test_count_files(self):
+        cnt = count_files('py')
+        self.assertGreater(cnt, 1)
```
christinaholtNOAA: Would it make sense to do a call to glob.glob here to find out what the exact number of files is when the test is run, and require that it matches that?

That might take care of ensuring that changes in test_count_files won't break expected behavior?

ush/python_utils/test_python_utils.py
```diff
@@ -0,0 +1,132 @@
...
+    def test_filesys_cmds(self):
+        dPATH=f'{self.PATH}/test_data/dir'
+        mkdir_vrfy(dPATH)
+        self.assertTrue( os.path.exists(dPATH) )
+        cp_vrfy(f'{self.PATH}/change_case.py {dPATH}/change_cases.py')
```
christinaholtNOAA: Should this be called with separate arguments instead of one string? I see that it will work this way, but maybe it makes sense to stay more consistent with other copy commands we are used to in Python....like `cp(source, target)` type syntax?

ush/python_utils/test_python_utils.py
```diff
@@ -0,0 +1,132 @@
...
+        self.assertTrue( os.path.exists(f'{dPATH}/change_cases.py') )
+        cmd_vrfy(f'rm -rf {dPATH}')
+        self.assertFalse( os.path.exists('tt.py') )
+    def test_get_charvar_from_netcdf(self):
+        FILE=f'{self.PATH}/test_data/sample.nc'
```
christinaholtNOAA, line 46: Where does this file come from?

ush/python_utils/test_python_utils.py
```diff
@@ -0,0 +1,132 @@
...
+            {'arg1': None,
+             'arg2': 'bye',
+             'arg3': 'hello',
+             'arg4': ['this', 'is', 'an', 'array']} )
+    def test_print_input_args(self):
+        valid_args = { "arg1":1, "arg2":2, "arg3":3, "arg4":4 }
+        self.assertEqual( print_input_args(valid_args), 4 )
+    def test_import_vars(self):
+        #test import
+        global MYVAR
```
christinaholtNOAA, line 94: I'm a little concerned with how this global might interact with the test. Why is a global needed here?

ush/python_utils/test_python_utils.py
```diff
@@ -0,0 +1,132 @@
...
+        import_vars(dictionary=dictionary)
+        self.assertEqual( Hello, "World!" )
+    def test_config_parser(self):
+        cfg = { "HRS": [ "1", "2" ] }
+        shell_str = cfg_to_shell_str(cfg)
+        self.assertEqual( shell_str, 'HRS=( "1" "2" )\n')
+    def test_print_msg(self):
+        self.assertEqual( print_info_msg("Hello World!", verbose=False), False)
+    def setUp(self):
+        define_macro_utilities();
```
christinaholtNOAA: It's a bit unclear what the `setUp` method does. Could you leave a few comments that point future developers in the right direction here?

christinaholtNOAA (commented):
@danielabdi-noaa This is really nice work! I left several minor comments below, and a few more general questions for folks from DTC (maybe @gsketefian, @mkavulich, and @JeffBeck-NOAA) could skim some of those for feedback?

ush/python_utils/environment.py
danielabdi-noaa: Note that the default `dictionary=os.environ` is passed by reference (a pointer), so `export_vars` does not keep its own persistent copy of dictionary, which is the reason why I do not return a new dictionary at the end of `export_vars`. Basically the behavior I want is what is what the python anti-pattern link wants to avoid. Maybe we can come up with something that breaks the example below? I think if export_vars had kept its own copy of the dictionary, it shouldn't know about the `del` call outside `export_vars` and return all (K1,K2,K3) instead of (K2,K3) which it does now.

```Python
import os

def export_vars(k,v,dict=os.environ):
    dict[k] = v
    return dict

export_vars("K1", "V1")
del os.environ["K1"]
export_vars("K2", "V2")
last_dict = export_vars("K3", "V3")

for k,v in os.environ.items():
    if k in ["K1","K2","K3"]:
        print(f"{k}={v}")
for k,v in last_dict.items():
    if k in ["K1","K2","K3"]:
        print(f"{k}={v}")
```

ush/python_utils/check_for_preexist_dir_file.py
danielabdi-noaa: I thought I could get this through the cracks without anyone noticing :) Yes, the behavior is different but I thought that it is better to save with the time stamp, instead of iterating through files in the directory. Let me know if I should revert back to old behavior ...

ush/python_utils/test_python_utils.py
danielabdi-noaa: setUp, wchich btw comes with the unittest class, is sort of like a constructor of and is called before each test case is run. If I need to download some files before running the unittests, it would be appropriate to put that in setUp.

ush/python_utils/test_python_utils.py
danielabdi-noaa: I haven't thought about `global MYVAR` affecting other unittests so good catch. The `global` is needed because of reasons I mentioned in the doc string of `export_vars`. Any environment variable that we want to modify need to be tagged with global otherwise python assumes it is a local variable. I have searched fruitlessly for a solution to save the programmer from tagging  environment variables, put to globals() dictionary using `export_vars`, as global. Maybe somehow notifying the python interpreter implicitly, but there doesn't seem to be any other solution. The other solution I thought about is to put env variables to locals() instead of globals(), but the former turns out to be not mutable at all ...

ush/python_utils/test_python_utils.py
danielabdi-noaa: I grabbed the smallest netcdf file I could find from the web, but let me know if there is some netcdf file on the ufs-weather-app repo somewhere ...

ush/python_utils/config_parser.py
danielabdi-noaa: yes, this is always used now. I kept the other one around in case we decide on a strictly key-value pair specification for config files but i can definately remove the other one.

ush/python_utils/create_symlink_to_file.py
danielabdi-noaa: Ok I will replace all instances of `==/!= None` with `is/is not None`

ush/python_utils/run_command.py
danielabdi-noaa: exit code, and std_err message will be set accordingly. It is up-to the caller module to check for errors though, however, i have noticed not all parts of workflow generation code check for a successful execution of a command so maybe doing this in this central place is better ? During development of the python workflow code, I've debugged errors of commands called executed in a sub-shell, such fill_jinja_template, by examining std_err and exit code.

ush/python_utils/is_array.py
danielabdi-noaa: This is not used in the python workflow generator so I will remove it all in all

ush/python_utils/is_element_of.py
danielabdi-noaa: Will be removed

ush/python_utils/process_args.py
danielabdi-noaa: `print_err_msg_exit` already does that so no need

ush/python_utils/set_file_param.py
danielabdi-noaa: This utility is not used at all in the python workflow generation, so I will remove it all in all.

ush/python_utils/check_var_valid_value.py
danielabdi-noaa: Done

ush/python_utils/run_command.py
danielabdi-noaa: Done

ush/python_utils/environment.py
danielabdi-noaa: Done

ush/python_utils/environment.py
danielabdi-noaa: Done

ush/python_utils/environment.py
danielabdi-noaa: Done

ush/python_utils/environment.py
danielabdi-noaa: Done

ush/python_utils/environment.py
danielabdi-noaa: Done

ush/python_utils/environment.py
danielabdi-noaa: Done

ush/python_utils/test_python_utils.py
danielabdi-noaa: Reason why I am stuck with "happy assertions" is because `print_err_msg_exit` calls these functions make exits the program so there is not return value to test for when they fail ...

ush/python_utils/test_python_utils.py
danielabdi-noaa: Good idea, this will also be a test of `count_files`

ush/python_utils/test_python_utils.py
danielabdi-noaa: Done

ush/python_utils/test_python_utils.py
danielabdi-noaa: Hmmm `count_files` does exactly that, i.e. glob.glob, so this would be a self-fulfilling prophecy ? Maybe some other way to count py files ...

ush/python_utils/environment.py
danielabdi-noaa: This func is used in `str_to_type` where we test for best matching data type, so I can't just `print_err_msg_exit` here. However, in the workflow code in `setup.py` the returned datetime object is tested if it is vald (`not None`). Maybe I can pass an additional optional argument to change the behaviour to either `test_is_date` or `convert_to_date_or_fail`?

ush/python_utils/test_python_utils.py
danielabdi-noaa: Ok took care of this with `ls -l *.py | wc -l`

ush/python_utils/check_for_preexist_dir_file.py
gsketefian: @danielabdi-noaa That sounds good to me.

ush/python_utils/check_for_preexist_dir_file.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python3
+
+import os
+from datetime import datetime
+from .print_msg import print_info_msg, print_err_msg_exit
+from .check_var_valid_value import check_var_valid_value
+from .filesys_cmds_vrfy import rm_vrfy, mv_vrfy
+
+def check_for_preexist_dir_file(path, method):
+    """ Check for preexisting directory and, if present, deal with it
```
gsketefian: `path` could be file as well (i.e. not just a directory).

ush/python_utils/check_for_preexist_dir_file.py
danielabdi-noaa: Ok thanks I will leave it as is then!

ush/python_utils/check_for_preexist_dir_file.py
danielabdi-noaa: Fixed

ush/python_utils/config_parser.py
danielabdi-noaa: Removed the unused config loader now.

ush/python_utils/environment.py
christinaholtNOAA: I think the caller testing the return value is likely sufficient. Thanks!

ush/python_utils/environment.py
christinaholtNOAA: This one is likely a dumb question. Is there any chance that when you import this function for the first time on a front end node, it gets cached in __pycache__ and then when you call it on the compute node, you are still holding on to some of the front-end environment that may be hard to debug? I know that's outside the realm of tests that we can do, but I'm just a bit curious if that could cause us problems...

Would this implementation give you different results than you're trying for?

```
def import_vars(dictionary=None,....):

    ...
    if dictionary is None:
        dictionary = os.environ
```
That _should_ capture the a current snapshot of the environment at run time, but won't cache the environment from the first import of the function.

ush/python_utils/run_command.py
christinaholtNOAA: Thanks. I used subprocess recently on another project (but not with stdout and stderr args) and was having trouble with a failed command doing sys.exit(). Good to know. :)

ush/python_utils/test_python_utils.py
christinaholtNOAA: That may be something we'd want to address in future iterations!

ush/python_utils/test_python_utils.py
christinaholtNOAA: Oh, I missed that it's stored with the repo. That's fine.

ush/python_utils/test_python_utils.py
christinaholtNOAA: I think this could be the thing that is symptom of the problem I was concerned with in export_vars.

ush/python_utils/test_python_utils.py
christinaholtNOAA: Would you mind adding those comments to the code?

christinaholtNOAA (commented):
@danielabdi-noaa thanks for addressing all those comments! I am still slightly concerned about the use of an immutable object in a function definition (see the comments below). All else looks great though!

ush/python_utils/environment.py
danielabdi-noaa: Ok I will do that. Just wanted to provide default values to save the developer from passing dictionary every time.

ush/python_utils/test_python_utils.py
danielabdi-noaa: It looks like I had a bug there, meant to check `dir` instead of `dir-1`. I have fixed this now but the number of directories is still 1 after successfully renaming of the original directory.

ush/python_utils/test_python_utils.py
danielabdi-noaa: Ok sounds good

ush/python_utils/test_python_utils.py
danielabdi-noaa: You probably will not like `setup.py` in the python workflow code where I was forced to declare lots of variables as `global`. This is basically what I didn't like about imitating the shell scripts with python, but atleast it is a start and maybe we will come up with a better solution.

ush/python_utils/test_python_utils.py
danielabdi-noaa: Done

christinaholtNOAA (approved):
@danielabdi-noaa This looks great! Thanks. :)

danielabdi-noaa:
Hi @gsketefian, I was wondering if you have more suggestions for this PR?

ush/python_utils/config_parser.py
```diff
@@ -0,0 +1,103 @@
...
+    Args:
+         config_file: path to config file script
+    Returns:
+         dictionary that should be equivalent to one obtained from parsing a yaml file.
+    """
+
+    # Save env vars before and after sourcing the scipt and then
+    # do a diff to get variables specifically defined/updated in the script
+    # Method sounds brittle but seems to work ok so far
+    code = dedent(f'''
```
gsketefian: This does sound brittle!  Sometimes bash will add its own variables in the background depending on what the script does, and using this method will add unexpected variables.  But if it's working, this is ok for now.  

Is this (i.e. the `load_shell_config` function) going to be needed after PR #691 goes in?

ush/python_utils/test_python_utils.py
```diff
@@ -0,0 +1,114 @@
...
+        self.assertEqual( get_elem_inds(arr, 'egg', 'last' ) , 4 )
+        self.assertEqual( get_elem_inds(arr, 'egg', 'all' ) , [0, 2, 4] )
+    def test_get_manage_externals_config_property(self):
+        self.assertIn( \
+            'regional_workflow',
+            get_manage_externals_config_property( \
+                f'{self.PATH}/../../../Externals.cfg',
+                'regional_workflow',
+                'repo_url'))
+    def test_interpol_to_arbit_CRES(self):
```
gsketefian, line 63: For regional applications, `RES` ranges from hundreds (~400 for a 25km domain) to thousands (~3400 for a 3km domain).  Can we add these typical values as well?

ush/python_utils/test_python_utils.py
```diff
@@ -0,0 +1,114 @@
...
+        self.assertTrue( val and (val.split()[0], '955.5,'))
+    def test_run_command(self):
+        self.assertEqual( run_command('echo hello'), (0, 'hello', '') )
+    def test_get_elem_inds(self):
+        arr = [ 'egg', 'spam', 'egg', 'rice', 'egg']
+        self.assertEqual( get_elem_inds(arr, 'egg', 'first' ) , 0 )
+        self.assertEqual( get_elem_inds(arr, 'egg', 'last' ) , 4 )
+        self.assertEqual( get_elem_inds(arr, 'egg', 'all' ) , [0, 2, 4] )
+    def test_get_manage_externals_config_property(self):
+        self.assertIn( \
```
gsketefian, line 57: Why do use the App's `Externals.cfg` file here for testing instead of creating one specifically for testing (whose contents won't change without this testing script also being run changed and rerun)?  It's not likely, but it is possible the name `regional_workflow` may change.  Also, the path to it (`../../../Externals.cfg`) may change.

ush/python_utils/__init__.py
```diff
@@ -0,0 +1,20 @@
+from .change_case import uppercase, lowercase
+from .check_for_preexist_dir_file import check_for_preexist_dir_file
+from .check_var_valid_value import check_var_valid_value
+from .count_files import count_files
+from .create_symlink_to_file import create_symlink_to_file
+from .define_macro_utilities import define_macro_utilities
```
gsketefian: Looks like `define_macro_utilities` needs to be changed to `define_macos_utilities`.  Also at several other locations below.

ush/python_utils/environment.py
```diff
@@ -0,0 +1,246 @@
...
+        target_dict[k] = str_to_list(v) 
+
+def export_vars(dictionary=None, source_dict=None, env_vars=None):
+    """ Export all (or select few) global variables of the caller module
+    to either the environement/dictionary. Call this function at the end of
+    a function that updates environment variables.
+
+    Args:
+        dictionary: target dictionary to set (default=os.environ)
+        source_dict: source dictionary (default=caller modules globals())
```
gsketefian, line 222: "caller modules" --> "caller module's" (i.e. add apostrophe; not trying to nitpick, but helps in understanding!)

ush/python_utils/config_parser.py
```diff
@@ -0,0 +1,103 @@
...
+      (set -o posix; set) > /tmp/t1
+      {{ . {config_file}; set +x; }} &>/dev/null
+      (set -o posix; set) > /tmp/t2
+      diff /tmp/t1 /tmp/t2 | grep "> " | cut -c 3-
+    ''')
+    (_,config_str,_) = run_command(code)
```
gsketefian, line 32: To be on the safe side, it might be a good idea to specify the shell in which `code` is running to be `bash`.  All of the executable scripts start with `#!/bin/bash`, but in case that is ever forgotten, it would be good to have the shell specified here.

ush/python_utils/test_python_utils.py
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env python3
```
gsketefian, line 1: Can we add documentation on how to run these unit tests in this file?  I mean just add at the top here what you say in the PR message, something like:

To run these unit tests, from the `ush` directory issue:
`python3 -m unittest -b python_utils/test_python_utils.py`

gsketefian:
@danielabdi-noaa Since my comments/questions don't involve anything critical, I'll approve.  You can address them and/or merge at your convenience.  Sorry for the delay in getting this review in.  Took me some time to understand some of the python I hadn't seen before.

gsketefian (approved):

dshawul:
@gsketefian Thanks for the suggestions! I am addressing them and will merge once it is complete.

ush/python_utils/config_parser.py
danielabdi-noaa: Unfortunately there doesn't seem to be any other way. I've tried a simpler version of loading config files that just assign values, i.e. VAR = VALUE, but even some config.sh files contain math calculations, if conditions etc that make parsing difficult. Running a subshell and getting the env variables set/updated seem to work fine so far though.

ush/python_utils/test_python_utils.py
danielabdi-noaa: Done

ush/python_utils/config_parser.py
danielabdi-noaa: Done

ush/python_utils/environment.py
danielabdi-noaa: Done

ush/python_utils/__init__.py
danielabdi-noaa: Done

Issue #603, opened 2021-09-20
crontab automation does not work on Cheyenne with tcsh


jwolff-ncar:
# Description

When I add the following lines to my config.sh
USE_CRON_TO_RELAUNCH="TRUE"
CRON_RELAUNCH_INTVL_MNTS=03

I have an entry in my cron on Cheyenne that looks like:
*/03 * * * * cd /glade/scratch/jwolff/SRW_training/expt_dirs/test_cron && ./launch_FV3LAM_wflow.sh >> ./log.launch_FV3LAM_wflow 2>&1

But when it tries to run I get the error message (in .log.launch_FV3LAM_wflow):
./launch_FV3LAM_wflow.sh: line 113: module: command not found
./launch_FV3LAM_wflow.sh: line 114: module: command not found

It then deletes that line from my cron because it thinks it was successful:
Summary of workflow status:
~~~~~~~~~~~~~~~~~~~~~~~~~~

  0 out of 0 cycles completed.
  Workflow status:  SUCCESS

========================================================================
End of output from script "launch_FV3LAM_wflow.sh".
========================================================================

The end-to-end run of the workflow for the forecast experiment specified 
by expt_name has completed with the following workflow status (wflow_-
status):
  expt_name = "test_cron"
  wflow_status = "SUCCESS"
Thus, there is no need to relaunch the workflow via a cron job.  Remo-
ving from the crontab the line (CRONTAB_LINE) that calls the workflow
launch script for this experiment:
  CRONTAB_LINE = "*/03 * * * * cd /glade/scratch/jwolff/SRW_training/expt_dirs/test_cron && ./launch_FV3LAM_wflow.sh >> ./log.launch_FV3LAM_wflow 2>&1"

## Steps to Reproduce

This may be an environment issue related to using tcsh so these steps would need to be conducted in that environment.

1. Generate an experiment on Cheyenne with the following lines in the config.sh
USE_CRON_TO_RELAUNCH="TRUE"
CRON_RELAUNCH_INTVL_MNTS=03

2. Check the log.launch_FV3LAM_wflow file for errors

jwolff-ncar:
@gsketefian We had some back and forth in Slack (under the SRW App Users' Training workspace) on a couple of options for a possible fix for this. Can we identify how to proceed and get this included?

gsketefian:
[...]

PR #675, opened 2022-02-07
Use non-login shell for `launch_FV3LAM_wflow.sh`; remove support for `WCOSS_CRAY`; fix cron capability for `tcsh` users on Cheyenne

gsketefian:
## DESCRIPTION OF CHANGES: 
This PR:
* Makes the `launch_FV3LAM_wflow.sh` script run in a non-login shell to avoid sourcing of user-specific login files (e.g `~/.bash_profile`, `~/.bash_login`, and `~/.profile`).
* Generalizes the loading of system scripts to initialize not only the `module` command but others (by sourcing `/etc/profile`).  This is necessary if not using a login shell in `launch_FV3LAM_wflow.sh`. 
* Adds an optional argument to `launch_FV3LAM_wflow.sh` that specifies whether it is being called from a `cron` job.  This is necessary because on Cheyenne, the version of `crontab` to use depends on whether or not the parent script is being called from a `cron` job.
* Removes support for `WCOSS_CRAY` since it will be decommissioned soon and is difficult to support.
* Removes machine-dependent `case` statements from ex-scripts.

One of the consequences of this use of non-login shell for `launch_FV3LAM_wflow.sh` is that the workflow's `cron` capability now works for `tcsh` users on Cheyenne.

## TESTS CONDUCTED: 
Ran the two WE2E tests `deactivate_tasks` and `grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2` on Hera, Jet, Cheyenne, and Orion.  All passed.

Note that I used `tcsh` as my login shell on Cheyenne for the purposes of testing this PR (and the tests succeeded).

## ISSUE (optional): 
Resolves Issue #603.

## CONTRIBUTORS:
@chan-hoo, @willmayfield, @mkavulich


ush/machine/cheyenne.sh
```diff
@@ -35,8 +32,14 @@ EXTRN_MDL_SYSBASEDIR_LBCS=${EXTRN_MDL_SYSBASEDIR_LBCS:-$(file_location \
   ${EXTRN_MDL_NAME_LBCS} \
   ${FV3GFS_FILE_FMT_ICS})}
 
-# System Installations
-MODULE_INIT_PATH=${MODULE_INIT_PATH:-/glade/u/apps/ch/opt/lmod/8.1.7/lmod/8.1.7/init/sh}
+# System scripts to source to initialize various commands within workflow
+# scripts (e.g. "module").
+if [ -z ${ENV_INIT_SCRIPTS_FPS:-""} ]; then
```
gsketefian, line 37: I generalized the `MODULE_INIT_PATH` variable to the array `ENV_INIT_SCRIPTS_FPS` (in all the macine files) because on some machines, more than one system script may need to be sourced to enable the `module` and other commands.

ush/machine/cheyenne.sh
```diff
@@ -35,8 +32,14 @@ EXTRN_MDL_SYSBASEDIR_LBCS=${EXTRN_MDL_SYSBASEDIR_LBCS:-$(file_location \
...
+  ENV_INIT_SCRIPTS_FPS=( "/etc/profile.d/modules.sh" "/etc/profile.d/pbs.sh" )
+fi
+
+# Commands to run at the start of each workflow task.
+PRE_TASK_CMDS='{ ulimit -s unlimited; ulimit -a; }'
```
gsketefian, line 42: Placed the commands that were previously called at the end of the machine files in the new variable `PRE_TASK_COMMANDS` so they aren't run every time the machine file is sourced but only as necessary.

scripts/exregional_make_grid.sh
```diff
@@ -107,8 +107,8 @@ case "$MACHINE" in
     ;;
 
   *)
-    source ${MACHINE_FILE}
-    ;;
+    source $USHDIR/source_machine_file.sh
```
gsketefian: Added a wrapper around sourcing of the machine file in order to include actions we may want to do for all those files (e.g. do `set -x` depending on the `DEBUG` variable).  See the new file `source_machine_file.sh` below.

scripts/exregional_make_grid.sh
```diff
@@ -107,8 +107,8 @@ case "$MACHINE" in
...
+    eval ${PRE_TASK_CMDS}
```
gsketefian: I moved the commands (`ulimit ...`) at the end of the machine files into the new variable `PRE_TASK_CMDS` so that they are not automatically executed when the machine file is sourced.  Execute them here though.

ush/constants.sh
```diff
@@ -15,3 +15,11 @@ degs_per_radian=$( bc -l <<< "360.0/(2.0*$pi_geom)" )
 # Radius of the Earth in meters.
 radius_Earth="6371200.0"
 
+#
+#-----------------------------------------------------------------------
+#
+# Other.
+#
+#-----------------------------------------------------------------------
+#
```
gsketefian, line 24: Convenience variable since this array shows up in multiple places.  Will use it in `valid_param_vals.sh` in a future PR.

ush/generate_FV3LAM_wflow.sh
```diff
@@ -59,6 +59,7 @@ ushdir="${scrfunc_dir}"
 #-----------------------------------------------------------------------
 #
 . $ushdir/source_util_funcs.sh
+. $ushdir/get_crontab_contents.sh
```
gsketefian, line 62: This new function is needed because of the way the `crontab` function is set up on Cheyenne.  The path to the command depends on whether or not it is being called from a cron job.  Also, on Cheyenne, `crontab` adds header lines to its output.  This function strips out those headers so the user's cron table doesn't unnecessarily grow.

ush/launch_FV3LAM_wflow.sh
```diff
@@ -93,12 +90,48 @@ fi
...
+. $USHDIR/init_env.sh
+. $USHDIR/get_crontab_contents.sh
+#
+#-----------------------------------------------------------------------
+#
+# Declare arguments.
+#
+#-----------------------------------------------------------------------
+#
+valid_args=( \
```
gsketefian, line 110: `FV3LAM_wflow.sh` now needs an optional argument (`called_from_cron`) in order to tell whether or not it is being called from a cron job.  This is needed on Cheyenne because the path to the `crontab` command (which is called in this script) depends on this distinction.

ush/load_modules_run_task.sh
```diff
@@ -65,56 +67,13 @@ fi
...
-      print_err_msg_exit "\
-      The script to source to initialize lmod (module loads) has not yet been
-      specified for the current machine (MACHINE):
-        MACHINE = \"$MACHINE\""
-    fi
-    ;;
-#
-esac
+env_init_scripts_fps_str="( "$(printf "\"%s\" " "${ENV_INIT_SCRIPTS_FPS[@]}")")"
+init_env env_init_scripts_fps="${env_init_scripts_fps_str}"
```
gsketefian, line 76: Replaced `case` statement with the new `init_env()` function that sources the necessary system scripts to make `module` and other commands (e.g. `qsub` on Cheyenne) available.  I created a separate function because the block of code is also needed in `launch_FV3LAM_wflow.sh`.

ush/launch_FV3LAM_wflow.sh
```diff
@@ -93,12 +90,48 @@ fi
...
+#
+#-----------------------------------------------------------------------
+#
+# Initialize the environment, e.g. by making the "module" command as well
+# as others available.
+#
+#-----------------------------------------------------------------------
+#
+env_init_scripts_fps_str="( "$(printf "\"%s\" " "${ENV_INIT_SCRIPTS_FPS[@]}")")"
+init_env env_init_scripts_fps="${env_init_scripts_fps_str}"
```
gsketefian, line 134: The new `init_env()` function sources one or more system scripts (now specified by the array `ENV_INIT_SCRIPTS_FPS` in the machine files) that enable use of certain commands like `module` (and `qsub` on Cheyenne).  These commands are not defined by default if not using a login shell.

ush/launch_FV3LAM_wflow.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -l
+#!/bin/bash
```
gsketefian, line 1: Removed the `-l` flag so that this script runs in a non-login shell.  This means a user's specific environment doesn't get loaded (e.g. the files `~/.bash_profile`, `~/.bash_login`, and `~/.profile` aren't sourced).  This is the only remaining workflow script in which the `-l` is (was) still used.  But removing `-l` also means `/etc/profile` isn't sourced, which means certain commands (like `module`) will have to be initialized (which is what the `init_env` function and `ENV_INIT_SCRIPTS_FPS` variable now do).

ush/launch_FV3LAM_wflow.sh
```diff
@@ -155,23 +187,22 @@ cd_vrfy "$exptdir"
...
-# file.
+# found, it means the end-to-end run of the workflow failed, so set the
+# status of the workflow to "FAILURE".
 #
 #-----------------------------------------------------------------------
 #
 tmp_fn="rocotorun_output.txt"
 rocotorun_cmd="rocotorun -w \"${WFLOW_XML_FN}\" -d \"${rocoto_database_fn}\" -v 10"
-eval ${rocotorun_cmd} > ${tmp_fn} 2>&1
+eval ${rocotorun_cmd} > ${tmp_fn} 2>&1 || \
```
gsketefian, line 197: When rocoto commands fail, the script should exit with an appropriate message.

ush/machine/odin.sh
```diff
@@ -36,6 +34,15 @@ EXTRN_MDL_SYSBASEDIR_LBCS=${EXTRN_MDL_SYSBASEDIR_LBCS:-$(file_location \
   ${EXTRN_MDL_NAME_LBCS} \
   ${FV3GFS_FILE_FMT_ICS})}
 
+# System scripts to source to initialize various commands within workflow
+# scripts (e.g. "module").
+if [ -z ${ENV_INIT_SCRIPTS_FPS:-""} ]; then
```
gsketefian, line 39: Not sure what system script is needed here (probably `/etc/profile` but I can't test since I have no access to Odin).

ush/machine/stampede.sh
```diff
@@ -36,6 +34,15 @@ EXTRN_MDL_SYSBASEDIR_LBCS=${EXTRN_MDL_SYSBASEDIR_LBCS:-$(file_location \
   ${EXTRN_MDL_NAME_LBCS} \
   ${FV3GFS_FILE_FMT_ICS})}
 
+# System scripts to source to initialize various commands within workflow
+# scripts (e.g. "module").
+if [ -z ${ENV_INIT_SCRIPTS_FPS:-""} ]; then
+  ENV_INIT_SCRIPTS_FPS=()
```
gsketefian, line 40: Not sure what system script is needed here (probably `/etc/profile` but I can't test since I have no access to Stampede).

chan-hoo:
@gsketefian, I'll test it on the wcoss dell and cray this afternoon.

gsketefian:
@willmayfield @mkavulich Do you guys mind testing this PR on Cheyenne?  It would be great if one of you could do it with a `bash` login shell and the other with `tcsh`.  Not sure who should do which.  It takes about a day to switch login shells (you do it [here](https://kb.ucar.edu/display/RC/Systems+Accounting+Manager)).

chan-hoo:
@gsketefian, it works well on wcoss_dell_p3. However, the crontab command was not created correctly on wcoss_cray. I think this error is not related to this pr. Can you take a look at it? If this is a new bug on wcoss_cray, I'll open a new issue for this. I copied the log file on Hera (lines 347, 376):  /scratch2/NCEPDEV/fv3-cam/Chan-hoo.Jeon/logs/log.generate_FV3LAM_wflow

chan-hoo:
@gsketefian, I found a critical issue on wcoss_dell_p3. The tasks were submitted well, but the module file is not loaded correctly. See the log file: "/scratch2/NCEPDEV/fv3-cam/Chan-hoo.Jeon/logs/make_grid.log". The current develop branch works well without this issue.

gsketefian:
> @gsketefian, I found a critical issue on wcoss_dell_p3. The tasks were submitted well, but the module file is not loaded correctly. See the log file: "/scratch2/NCEPDEV/fv3-cam/Chan-hoo.Jeon/logs/make_grid.log". The current develop branch works well without this issue.

@chan-hoo This debugging is going to be difficult to do in this PR.  I'm wondering if we can do a google meet and debug together on WCOSS_*.  I have a meeting from 2:30-4pm ET today but free before and after, or we try next week.  Thanks.

gsketefian:
> @gsketefian, I've tested this pr on the wcoss_dell_p3. It worked well. However, it has many issues on the wcoss cray. I think it would be better for me to open a separate pr for the cray. What do you think of this? My concern is that I'll have time to fix the issues on the cray next week.

That is ok with me if you don't mind it being broken on the cray.  Do you know if the current `develop` branch works on the cray?  If not, then that would definitely be ok.  If it does work, we can leave the `case` statements in the ex-scripts so that the cray's machine file doesn't get sourced (and we just run the previous set of commands that work).

chan-hoo:
@gsketefian, if you are ok, I'll fix the issues on the cray next Tuesday.

scripts/exregional_make_sfc_climo.sh
```diff
@@ -169,7 +169,7 @@ does not exist:
 Please ensure that you've built this executable."
 fi
 
-${RUN_CMD_UTILS} ${exec_fp} || \
+${RUN_CMD_SFC_CLIMO} ${exec_fp} || \
 print_err_msg_exit "\
```
JeffBeck-NOAA: Hey @chan-hoo, just curious why SFC_CLIMO has to be separated on WCOSS_CRAY from the other pre-processing utility tasks?

chan-hoo: @JeffBeck-NOAA, the command `RUN_CMD_UTILS` didn't work for the task `make_sfc_climo` on the cray.

chan-hoo:
@gsketefian, this pr works on the wcoss dell and cray well now. However, it doesn't work on hera: ./generate_FV3LAM_wflow.sh: line 509: crontab_contents: unbound variable

JeffBeck-NOAA:
> @gsketefian, this pr works on the wcoss dell and cray well now. However, it doesn't work on hera: ./generate_FV3LAM_wflow.sh: line 509: crontab_contents: unbound variable

@chan-hoo, thanks for getting this working on the WCOSS Cray and Dell.  I think @gsketefian was able to get this to run on Hera today without any problem, so I'll let him comment.  

A note on the WCOSS Cray, I spoke with Jacob today regarding the move to WCOSS2 that's coming soon.  Since the WCOSS Cray will be retired this summer, we should be able to wind down our support for that platform at this point.

gsketefian:
@chan-hoo I'm going to include parts of our email exchange here (and my responses) so others understand the background.

> On Wed, Feb 23, 2022 at 7:01 AM Chan-Hoo Jeon - NOAA Affiliate <[chan-hoo.jeon@noaa.gov](mailto:chan-hoo.jeon@noaa.gov)> wrote:
> Gerard,
> 
> The problem was 
> 
> ```
>   if [ "$MACHINE" = "WCOSS_DELL_P3" ]; then
>     grep_output=$( grep "^${crontab_line_esc_astr}$" "/u/$USER/cron/mycrontab" )
>   else
>     grep_output=$( echo "${crontab_contents}" | grep "^${crontab_line_esc_astr}$" )
>   fi
> 
>   exit_status=$?
> 
>   if [ "${exit_status}" -eq 0 ]; then
> ```
> Even in case grep_output="" (nothing), exit_status=0 (It is not supposed to be zero).
> So, nothing copies to the crontab file.
> 
> Do you have an idea of how to fix this?
> 
> Thanks,
> Chan-Hoo
> 

It's strange that on the cray, using 
```
grep_output=$( echo "${crontab_contents}" | grep "^${crontab_line_esc_astr}$" )
```
doesn't work (i.e. the exit status of `grep` is 0 even if the line is not found in the cron table) but
```
grep_output=$( crontab -l | grep "^${crontab_line_esc_astr}$" )
```
works (i.e. returns a non-zero exit status if the line is not found).   Since both commands use `grep`, if the line is not in the cron table, then the exit status should be non-zero regardless of whether the input to `grep` came from `echo` or from `crontab -l`.  

> 
> On Wed, Feb 23, 2022 at 5:36 AM Chan-Hoo Jeon - NOAA Affiliate <chan-hoo.jeon@noaa.gov> wrote:
> The original expression works well.
> So I will add it to the if statement.
> 
> ```
>   elif [ "$MACHINE" = "WCOSS_CRAY" ]; then
>     grep_output=$( crontab -l | grep "^${crontab_line_esc_astr}$" )
> 
> ```
> Chan-Hoo
> 

I am reluctant to use this fix because it defeats the purpose of having the new function `get_crontab_contents.sh` that deals with the oddities of getting the contents of the cron table on each machine all in one place instead of having to put if-statements in various places in the scripts.  An alternative fix might be to look at the contents of `grep_output` instead of `exit_status` in determining what to do in the if-statement that follows these lines.

But -- since it is such a hassle going back and forth to debug this on the cray (because I don't have access to it) and since @JacobCarley-NOAA gave the ok to remove support for the cray from the SRW App, unless you have an objection I am going to go ahead and remove the cray in my next commit.

Btw, I did test this latest version on Hera and it worked for me.  I will re-test once I make the latest changes to remove the cray.

chan-hoo:
@gsketefian, I don't have any objection for the change at all. It will make my work easier than before :)

gsketefian:
@chan-hoo Ok, great!  Makes life a lot easier!  I just pushed a commit that removes support for WCOSS_CRAY.  Do you mind testing on WCOSS_DELL_P3 one more time?

@willmayfield Do you mind testing the latest (hash 930f4cc) on Cheyenne?  I tested with `tcsh` and it worked for me, but it would be great if you can as well.  It would be nice to also test with `bash`.  Can someone else test with `bash` on Cheyenne (since it takes a day to switch login shells)?  @mkavulich, do you happen to have time?

I tested on Hera and it works.  Currently testing on Jet.  I can't test on Orion today since it's down for maintenance, so I will try tomorrow.

Thanks all.

scripts/exregional_make_ics.sh
```diff
@@ -86,27 +86,12 @@ export OMP_STACKSIZE=${OMP_STACKSIZE_MAKE_ICS}
...
-    ;;
-
-esac
+source $USHDIR/source_machine_file.sh
+eval ${PRE_TASK_CMDS}
 
 nprocs=$(( NNODES_MAKE_ICS*PPN_MAKE_ICS ))
 
-if [ -z ${RUN_CMD_UTILS:-} ] ; then
+if [ -z "${RUN_CMD_UTILS:-}" ] ; then
```
gsketefian, line 94: Need to have this in (double) quotes because on Cheyenne, if `RUN_CMD_UTILS` contains spaces (e.g. `mpirun -np ...`), the `-z` operator will fail with a `too many arguments` error.  Same in all the other ex-scripts.

ush/templates/FV3LAM_wflow.xml
```diff
@@ -177,7 +177,7 @@ MODULES_RUN_TASK_FP script.
 
     &RSRV_DEFAULT;
     <command>&LOAD_MODULES_RUN_TASK_FP; "&MAKE_GRID_TN;" "&JOBSDIR;/JREGIONAL_MAKE_GRID"</command>
-  {%- if machine in ["WCOSS_DELL_P3", "WCOSS_CRAY"]  %}
+  {%- if machine in ["WCOSS_DELL_P3"]  %}
```
gsketefian, line 180: Removing support for WCOSS_CRAY.

chan-hoo:
@gsketefian, it works well on the dell.

chan-hoo (approved):

gsketefian:
> @gsketefian, it works well on the dell.

Great, thanks @chan-hoo!

JeffBeck-NOAA (approved):
Approving based on previous assessment.  We need to have a final test run on Cheyenne to confirm things work.  @willmayfield or @mkavulich, do either of you have time to test (with tcsh and/or bash)?

willmayfield:
@JeffBeck-NOAA I'll be testing it today.

willmayfield (approved):
tested on Cheyenne with bash and tcsh.

PR #689, opened 2022-03-03
Add NOAA cloud platforms to SRW

mark-a-potts:
## DESCRIPTION OF CHANGES: 
Adds all three NOAA cloud platforms to the SRW application with both Intel and GCC compilers pre-configured.

## TESTS CONDUCTED: 
The full workflow based on a modified GST has been successfully completed using Intel compilers on both AWS and Azure platforms.
The full workflow currently fails when using GNU compilers due to a bug in (most likely) the UPP.

## DEPENDENCIES:
None

## DOCUMENTATION:
WIP


christinaholtNOAA (changes requested):
@mark-a-potts Thanks for adding the files and example on how this is done! My apologies for using your PRs as a "big picture" question forum.

Reviewers -- please see my comments in the corresponding SRW PR about how we might best support new platforms.

ush/machine/noaacloud.sh
```diff
@@ -0,0 +1,82 @@
...
+#NDAS_OBS_DIR="/scratch2/BMC/det/UFS_SRW_app/develop/obs_data/ndas/proc"
+#MET_BIN_EXEC="bin"
+
+# Test Data Locations
+#TEST_PREGEN_BASEDIR=/scratch2/BMC/det/FV3LAM_pregen
+#TEST_COMINgfs=/scratch2/NCEPDEV/fv3-cam/noscrub/UFS_SRW_App/COMGFS
+#TEST_EXTRN_MDL_SOURCE_BASEDIR=/scratch2/BMC/det/Gerard.Ketefian/UFS_CAM/staged_extrn_mdl_files
+#TEST_ALT_EXTRN_MDL_SYSBASEDIR_ICS=/scratch2/BMC/det/UFS_SRW_app/dummy_FV3GFS_sys_dir
+#TEST_ALT_EXTRN_MDL_SYSBASEDIR_LBCS=/scratch2/BMC/det/UFS_SRW_app/dummy_FV3GFS_sys_dir
+
```
christinaholtNOAA: Consider removing commented lines.

christinaholtNOAA (approved):
After our conversation in the EPIC Sprint Review this morning, I am less concerned about this not being a traditionally "supported" platform. Just one quick requested change from me.

ush/machine/noaacloud.sh
mark-a-potts: I can definitely remove the commented lines for this PR, but ultimately, I think we would like to be able to use both MetPlus and run tests the same way they are run on other Tier 1 machines. That said, I just checked on the size of those test directories and one of them is 2.3 TB. Is all of that data needed for the tests? (I am assuming, perhaps incorrectly, that this is data used for regression testing of the SRW?)

JeffBeck-NOAA: @mark-a-potts, the obs paths specified above are for the graduate student and workflow end-to-end (WE2E) tests for the Tier 1 machines.  The 2.3 TB is coming from the different external model data required for all of these tests.

mark-a-potts: Just removed the commented lines and pushed the changes.

PR #692, opened 2022-03-04
Bug fixes in machine files

gsketefian:
## DESCRIPTION OF CHANGES: 
Cleaning up bugs in the machine files.  The first bug prompted this PR, and the rest were found subsequently.  The bugs (and their fixes) are as follows:

1) A space is missing after the `print_info_msg` and `print_err_msg_exit` function calls in the `file_location` functions.  Inserting a space gets passed this bug, but subsequent issues were found as described below.

**For machine files that call the `print_info_msg` function in `file_location` (`cheyenne.sh`, `hera.sh`, `jet.sh`, and `orion.sh`):**
Fixing this bug leads to other failures because when the "*" stanza is encountered in the `file_location` function, 
the `EXTRN_MDL_SYSBASEDIR_ICS|LBCS` variable gets set to the message that `file_location` returns.  Since that message contains spaces, it leads to other failures in downstream scripts (the ex-scripts).  Simply removing the printing out of the message (thus causing `EXTRN_MDL_SYSBASEDIR_ICS|LBCS` to be set to a null string) fixes the failures, so this was the fix implemented.  If desired, a message for an empty value for `EXTRN_MDL_SYSBASEDIR_ICS|LBCS` can be placed in another script (where those variables are used).

**For machine files that use `print_err_msg_exit` in `file_location` (`stampede.sh` and `wcoss_dell_p3.sh`):**
These should not exit if the file location is not available since the experiment can still complete successfully.  So just removing the `print_err_msg_exit` call should work (and make the behavior of these machine files consistent with the set above).

2) In all the machine files, the variable `FV3GFS_FILE_FMT_ICS` should be changed to `FV3GFS_FILE_FMT_LBCS` in the definition of `EXTRN_MDL_SYSBASEDIR_LBCS`.  This was fixed in all the files.

3) In `stampede.sh`, a variable named `SYSBASEDIR_ICS` is defined.  This is a typo.  Modify to `EXTRN_MDL_SYSBASEDIR_ICS`.

## TESTS CONDUCTED: 
Ran the WE2E test `grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_GSD_SAR` on:
* Hera -- successful
* Jet -- successful except for UPP tasks
* Cheyenne -- successful except for UPP tasks

The UPP task failures are new and being experienced by other PRs as well (e.g. #689).  The original issue with machine files seems resolved.

## CONTRIBUTORS (optional): 
@JeffBeck-NOAA encountered and reported the original error.


ush/machine/cheyenne.sh
```diff
@@ -71,6 +68,6 @@ NDAS_OBS_DIR="/glade/p/ral/jntp/UFS_SRW_app/develop/obs_data/ndas/proc"
 MET_BIN_EXEC="bin"
```
JeffBeck-NOAA, line 68: Does it matter that RUN_CMD_SERIAL and MET_BIN_EXEC use double quotes, and the others use single quotes?  Are the single quotes used for commands that are "expandable"?

ush/machine/cheyenne.sh
```diff
@@ -71,6 +68,6 @@ NDAS_OBS_DIR="/glade/p/ral/jntp/UFS_SRW_app/develop/obs_data/ndas/proc"
...
 
 # Test Data Locations
-TEST_PREGEN_BASEDIR=/glade/p/ral/jntp/UFS_CAM/FV3LAM_pregen
-TEST_COMINgfs=/glade/scratch/ketefian/NCO_dirs/COMGFS
-TEST_EXTRN_MDL_SOURCE_BASEDIR=/glade/p/ral/jntp/UFS_SRW_app/staged_extrn_mdl_files
```
JeffBeck-NOAA, line 76: These paths are in double quotes, but the "location" path above is in single quotes.  Is that OK?

JeffBeck-NOAA (approved):

ush/machine/cheyenne.sh
gsketefian: If you want to use templated variables (as we do for `RUN_CMD_UTILS` since they contain references to variables like `nprocs` that are not yet defined), then use single quotes.  For variables whose definitions don't include references to other variables, either single or double quotes work.

ush/machine/singularity.sh
```diff
@@ -9,6 +9,7 @@ function file_location() {
   external_model=${1}
   external_file_fmt=${2}
 
+  location=""
   case ${external_model} in
 
     "*")
```
gsketefian: @EdwardSnyder-NOAA I'm wondering how you were able to run a successful WE2E test with this `print_info_msg` call since this causes `EXTRN_MDL_SYSBASEDIR_ICS|LBCS` to be set to the message itself, and that leads to failures downstream in the `exregional_...` scripts -- at least on the on-prem machines.  I wanted to remove this `case` statement since it doesn't have any stanzas for which `location` is set to something (just like `orion.sh` above).  But wanted to ask you first.

christinaholtNOAA (approved):

ush/machine/singularity.sh
EdwardSnyder-NOAA: It is fine to remove that case statement. We haven't ran the WE2E on the singularity container yet.

gsketefian: Ok, thanks.

PR #698, opened 2022-03-10
Adds an alternative python workflow generation path

danielabdi-noaa:
## DESCRIPTION OF CHANGES: 
This PR provides an alternative workflow generation path using python3.

### Hera
All W2E tests have been run on Hera using the python workflow generator and the existing shell script workflow generator.
Although there were about 24 failures, both workfkow generation paths result in the same tests successfully completed.

### Orion
All W2E tests have been run on Hera and there are 21 failures, even better than Hera.

### Jet
On hold. Jet seems to have a lot of missing data, especially with latest develop, which could not even generate the workflow
because of missing data:
```
EXTRN_MDL_SOURCE_BASEDIR_ICS = "/mnt/lfs4/BMC/wrfruc/UFS_SRW_app/staged_extrn_mdl_files/FV3GFS/nemsio"
```

### Unit tests
Unittest on the new files can be done as follows
```
python3 -m unittest -b \
	check_ruc_lsm.py \
	create_diag_table_file.py \
	create_model_configure_file.py \
	link_fix.py \
	set_cycle_dates.py \
	set_extrn_mdl_params.py \
	set_FV3nml_sfc_climo_filenames.py \
	set_FV3nml_ens_stoch_seeds.py \
	set_gridparams_ESGgrid.py \
	set_gridparams_GFDLgrid.py \
	set_ozone_param.py \
	set_predef_grid_params.py \
	set_thompson_mp_fix_files.py \
        get_crontab_contents.py

..............
----------------------------------------------------------------------
Ran 14 tests in 1.448s

OK
```
## DEPENDENCIES:
None

## DOCUMENTATION:
None

## ISSUE (optional): 
None

## CONTRIBUTORS (optional): 
@christinaholtNOAA for suggestion


ush/create_diag_table_file.py
```diff
@@ -0,0 +1,80 @@
...
+    Returns:
+        Boolean
+    """
+
+    #process input arguments
+    valid_args = [ "run_dir" ]
+    dictionary = process_args(valid_args, **kwargs)
+    print_input_args(dictionary)
+    import_vars(dictionary=dictionary)
+
```
christinaholtNOAA, line 22: Is this process of passing arguments and checking them necessary in python? Would we get the same functionality by requiring a single parameter be passed to `create_diag_table_file`, and then printing it  print it with `locals()`?

ush/create_diag_table_file.py
```diff
@@ -0,0 +1,80 @@
...
+        to create:
+        
+            diag_table_fp = \"{diag_table_fp}\"''', verbose=VERBOSE)
+
+    settings = dedent(f'''
+            starttime: !datetime {type_to_str(CDATE)}
+            cres: {CRES}''')
+
+    #call fill jinja
+    (err,_,_) = run_command( f'''{USHDIR}/fill_jinja_template.py -q -u "{settings}" -t {DIAG_TABLE_TMPL_FP} -o {diag_table_fp}''')
```
christinaholtNOAA: Could this become a Python native call by importing `fill_jinja_template`? Maybe that standalone tool could use some work to make that more straightforward, though?

ush/create_diag_table_file.py
```diff
@@ -0,0 +1,80 @@
...
+        USHDIR = os.path.dirname(os.path.abspath(__file__))
+        DIAG_TABLE_FN="diag_table"
+        DIAG_TABLE_TMPL_FP = f'{USHDIR}/templates/{DIAG_TABLE_FN}.FV3_GFS_v15p2'
+        set_env_var('DEBUG',True)
+        set_env_var('VERBOSE',True)
+        set_env_var("USHDIR",USHDIR)
+        set_env_var("DIAG_TABLE_FN",DIAG_TABLE_FN)
+        set_env_var("DIAG_TABLE_TMPL_FP",DIAG_TABLE_TMPL_FP)
+        set_env_var("CRES","C48")
+        set_env_var("CDATE","2021010106")
```
christinaholtNOAA, line 78: I love that the test is here so close to the code!

ush/create_model_configure_file.py
```diff
@@ -0,0 +1,241 @@
...
+    #-----------------------------------------------------------------------
+    #
+    print_info_msg(f'''
+        Creating a model configuration file (\"{MODEL_CONFIG_FN}\") in the specified
+        run directory (run_dir):
+          run_dir = \"{run_dir}\"''', verbose=VERBOSE)
+    #
+    # Extract from cdate the starting year, month, day, and hour of the forecast.
+    #
+    yyyy=cdate.year
```
christinaholtNOAA, line 46: Is cdate meant to be a datetime object?

ush/generate_FV3LAM_wflow.py
```diff
@@ -0,0 +1,1029 @@
...
+        
+        Instructions for setting up python environments can be found on the web:
+        https://github.com/ufs-community/ufs-srweather-app/wiki/Getting-Started
+        ''', stack_trace=False)
+
+# Check for non-standard python packages
+try:
+    import jinja2
+    import yaml
+    #import f90yml
```
christinaholtNOAA: I suppose this one isn't needed here?

ush/generate_FV3LAM_wflow.py
```diff
@@ -0,0 +1,1029 @@
...
+            slash_ensmem_subdir = f"/mem#{ensmem_indx_name}#"
+
+        #get time string
+        (_,time_str,_) = \
+            run_command(f'{DATE_UTIL} -d "{date_to_str(DATE_FIRST_CYCL,True)} +{DT_ATMOS} seconds" +%M:%S')
+
+        cycl_hrs_str = [ f"{c:02d}" for c in CYCL_HRS ]
+
+        #yaml-string of settings
+        settings = f"""
```
christinaholtNOAA: Does it make sense to make this a dict (python native) and then use yaml dump to put it in the appropriately formatted string? Maybe we can leave that for later when we work toward different config organization?

ush/set_FV3nml_sfc_climo_filenames.py
```diff
@@ -0,0 +1,139 @@
...
+
+    dummy_run_dir = EXPTDIR + "/any_cyc"
+    if DO_ENSEMBLE == "TRUE":
+        dummy_run_dir += "/any_ensmem"
+
+    for i,mapping in enumerate(FV3_NML_VARNAME_TO_SFC_CLIMO_FIELD_MAPPING):
+        (_,nml_var_name,_) = run_command( f''' printf "%s\n" "{mapping}" | 
+                  {SED} -n -r -e "s/{regex_search}/\\1/p" ''')
+        (_,sfc_climo_field_name,_) = run_command(f''' printf "%s\n" "{mapping}" | 
+                          {SED} -n -r -e "s/{regex_search}/\\2/p" ''')
```
christinaholtNOAA: These lines might be able to leverage native python calls so we can more easily capture errors than with bash calls outside the python environment.

Also, I don't see `i` used here as a variable. Maybe `enumerate` isn't needed?

ush/set_FV3nml_sfc_climo_filenames.py
```diff
@@ -0,0 +1,139 @@
...
+
+        check_var_valid_value(sfc_climo_field_name, SFC_CLIMO_FIELDS)
+
+        fp = f'{FIXLAM}/{CRES}.{sfc_climo_field_name}.{suffix}'
+        if RUN_ENVIR != "nco":
+            (_,fp,_) = run_command( f'realpath --canonicalize-missing --relative-to="{dummy_run_dir}" "{fp}"' )
```
christinaholtNOAA: Can this path be built with python-native calls?

ush/set_cycle_dates.py
```diff
@@ -0,0 +1,57 @@
...
+
+    #calculate date increment
+    if incr_cycl_freq <= 24:
+        incr_days = 1
+    else:
+        incr_days = incr_cycl_freq // 24
+        if incr_cycl_freq % 24 != 0:
+            print_err_msg_exit(f'''
+                INCR_CYCL_FREQ is not divided by 24:
+                  INCR_CYCL_FREQ = \"{incr_cycl_freq}\"''')
```
christinaholtNOAA, line 36: @gsketefian Why is cycling frequency bound to be daily?

ush/set_ozone_param.py
```diff
@@ -0,0 +1,233 @@
...
+    num_symlinks=len(CYCLEDIR_LINKS_TO_FIXam_FILES_MAPPING)
+    
+    for i in range(num_symlinks):
+      mapping=CYCLEDIR_LINKS_TO_FIXam_FILES_MAPPING[i]
+      (_,symlink,_)=run_command(f'''printf "%s\n" "{mapping}" | \
+                 {SED} -n -r -e "s/{regex_search}/\\1/p"''')
+      if symlink == ozone_symlink:
+        regex_search="^[ ]*([^| ]+[ ]*)[|][ ]*([^| ]*)[ ]*$"
+        (_,mapping_ozone,_)=run_command(f'''printf "%s\n" "{mapping}" | \
+                         {SED} -n -r -e "s/{regex_search}/\\1/p"''')
```
christinaholtNOAA: Repeat from a previous comment...can these leverage python-native commands instead of subprocesses?

ush/set_predef_grid_params.py
```diff
@@ -0,0 +1,1214 @@
+#!/usr/bin/env python3
+
```
christinaholtNOAA, line 2: I'd love to see all this move to a YAML file, but may be outside the scope of this PR.

christinaholtNOAA (changes requested):
@danielabdi-noaa Thank you for this huge effort! I have gotten about halfway through a review, and a few of the comments I've made so far are applicable in many places within the code. In general I think it is useful to use Python-native commands where we can so that we can more easily leverage exception handling. We may also find that it will be faster than spinning up subprocesses to do things Python will handle for us (mostly untested conjecture). 

I have not finished my review just yet. I only made it through the `ush/set_thompson_mp_fix_files.py` and need to step away from it for a little while (maybe until Monday).

ush/create_diag_table_file.py
danielabdi-noaa: I have replaced all instances of such use of `fill_jinja_template` and `set_namelist` with a direct python function call. The arguments are passed a list of strings since that is the easiest thing to do. I could not pass it as one long string and break it down because of `settings` etc that are multiline strings... Let me know if you think of a better approach.

christinaholtNOAA: Big never mind to the comment below. I saw the mods you made to fill_jinja_template. Makes total sense!

Thanks for doing that. I was testing it out in a Python interactive session and finding that I wasn't able to just pass the list to the imported module. I needed these steps: 

```
import sys
### Probably want to save the original sys.argv here, and reset it after this step
sys.argv = ["-q", "-u", settings_str, "-t", DIAG_TABLE_TMPL_FP, "-o", diag_table_fp]
cla = fill_jinja_template.parse_args()
fill_jinja_template.main(cla)
```

We could certainly open PRs to the standalone tools to make them easier to deal with as imported scripts, if that's helpful here. Let me know what you think.

ush/set_thompson_mp_fix_files.py
```diff
@@ -0,0 +1,177 @@
...
+        ]
+    
+        if (EXTRN_MDL_NAME_ICS  != "HRRR" and EXTRN_MDL_NAME_ICS  != "RAP") or \
+           (EXTRN_MDL_NAME_LBCS != "HRRR" and EXTRN_MDL_NAME_LBCS != "RAP"):
+          thompson_mp_fix_files.append(thompson_mp_climo_fn)
+    
+        FIXgsm_FILES_TO_COPY_TO_FIXam.extend(thompson_mp_fix_files)
+    
+        num_files=len(thompson_mp_fix_files)
+        for i in range(num_files):
```
christinaholtNOAA: This could simplify to

```
for fix_file in thompson_mp_fix_files:
    mapping = f"{fix_file | fix_file}"
    ...
```

ush/setup.py
```diff
@@ -0,0 +1,2157 @@
...
+    #
+    # Make sure different variables are set to their corresponding valid value
+    #
+    #-----------------------------------------------------------------------
+    #
+    global VERBOSE
+    if DEBUG == True and VERBOSE == False:
+        print_info_msg('''
+            Resetting VERBOSE to \"TRUE\" because DEBUG has been set to \"TRUE\"...''')
+        VERBOSE=False
```
christinaholtNOAA, line 122: This should be `VERBOSE = True` based on the comment.

ush/setup.py
```diff
@@ -0,0 +1,2157 @@
...
+    try:
+      mng_extrns_cfg_fn = os.readlink(mng_extrns_cfg_fn)
+    except:
+      pass
+    property_name="local_path"
+    #
+    # Get the path to the workflow scripts
+    #
+    external_name="regional_workflow"
+    HOMErrfs = get_manage_externals_config_property( \
```
christinaholtNOAA: It would likely take less time overall (and file system stress) to interrogate this file only once, and return its contents as a dict. Then the variables could be set by getting the item from the dict. Maybe there is a separate python utility (or maybe the existing one gets modified)?

```
mng_extrns_cfg = load_manage_externals_config(mng_extrns_cfg_fn)
HOMErrfs = mng_extrns_cfg.get(external_name, {}).get(property_name)
...
UFS_WTHR_MDL_DIR = mng_extrns_cfg.get(external_name, {}).get(property_name)
...
```

ush/setup.py
```diff
@@ -0,0 +1,2157 @@
...
+    JOBSDIR=f"{HOMErrfs}/jobs"
+    SORCDIR=f"{HOMErrfs}/sorc"
+    SRC_DIR=f"{SR_WX_APP_TOP_DIR}/src"
+    PARMDIR=f"{HOMErrfs}/parm"
+    MODULES_DIR=f"{HOMErrfs}/modulefiles"
+    EXECDIR=f"{SR_WX_APP_TOP_DIR}/bin"
+    TEMPLATE_DIR=f"{USHDIR}/templates"
+    VX_CONFIG_DIR=f"{TEMPLATE_DIR}/parm"
+    METPLUS_CONF=f"{TEMPLATE_DIR}/parm/metplus"
+    MET_CONFIG=f"{TEMPLATE_DIR}/parm/met"
```
christinaholtNOAA: Since `os.path.join` helps us maintain platform independence (MacOS has historically been supported with SRW), it might be better to use it for dealing with path manipulation.

ush/setup.py
```diff
@@ -0,0 +1,2157 @@
...
+        Please ensure this value has been set for your desired platform. """)
+    
+    if not (FIXgsm and FIXaer and FIXlut and TOPO_DIR and SFC_CLIMO_INPUT_DIR):
+      print_err_msg_exit(f'''
+        One or more fix file directories have not been specified for this machine:
+          MACHINE = \"{MACHINE}\"
+          FIXgsm = \"{FIXgsm:-\"\"}
+          FIXaer = \"{FIXaer:-\"\"}
+          FIXlut = \"{FIXlut:-\"\"}
+          TOPO_DIR = \"{TOPO_DIR:-\"\"}
```
christinaholtNOAA: Is this syntax for f-strings, or leftover from bash?

ush/setup.py
```diff
@@ -0,0 +1,2157 @@
...
+    #-----------------------------------------------------------------------
+    #
+    global NH0,NH3,NH4
+    NH0=0
+    NH3=3
+    NH4=4
+
+    # export env vars
+    EXPORTS = ["NH0","NH3","NH4"]
+    export_vars(env_vars = EXPORTS)
```
christinaholtNOAA, line 1435: Sorry that this is so late in the game, but it only just occurred to me. What if `export_vars()` accepted a dictionary of `env_vars`...would that free up the requirements for Python globals? Here's a quick version of what I was thinking the call might look like:


```
NH0=0
NH3=3
NH4=4
exports = dict(NHO=NHO, NH3=NH3, NH4=NH4)
export_vars(env_vars = exports)

File: ush/setup.py [hunk 18]
```diff
@@ -0,0 +1,2157 @@
...
+    
+        RUN_TASK_VX_ENSGRID=False
+    
+        msg+=f'''
+            RUN_TASK_VX_ENSGRID = \"{RUN_TASK_VX_ENSGRID}\"
+        '''
+    
+        print_info_msg(msg)
+    
+    #
```
christinaholtNOAA, line 1326: Would you mind setting the indentation of the comments to the same level as the code? It makes it easier to visually follow the logic.

christinaholtNOAA (changes requested):
@danielabdi-noaa Thanks again for all this work. A few more similar comments below. I think I made it through all of it this time...

ush/setup.py
```diff
@@ -0,0 +1,2157 @@
...
+    
+        msg+=f'''
+            RUN_TASK_MAKE_SFC_CLIMO = \"{RUN_TASK_MAKE_SFC_CLIMO}\"
+            SFC_CLIMO_DIR = \"{SFC_CLIMO_DIR}\"
+        '''
+    
+        print_info_msg(msg)
+    
+      if RUN_TASK_VX_GRIDSTAT == True or \
+         RUN_TASK_VX_GRIDSTAT == False:
```
christinaholtNOAA: This logic seems unnecessary.

ush/setup.py
```diff
@@ -0,0 +1,2157 @@
...
+            RUN_TASK_MAKE_OROG = \"{RUN_TASK_MAKE_OROG}\"
+            OROG_DIR = \"{OROG_DIR}\"
+        '''
+    
+        print_info_msg(msg)
+    
+    
+      if RUN_TASK_MAKE_SFC_CLIMO == True or \
+         ( RUN_TASK_MAKE_SFC_CLIMO == False and \
+           SFC_CLIMO_DIR != nco_fix_dir ):
```
christinaholtNOAA, line 1240: Since these variables are boolean values, this logic statement becomes fairly simple in python:

```
if RUN_TASK_MAKE_SFC_CLIMO or \
  ( not RUN_TASK_MAKE_SFC_CLIMO and SFC_CLIMO_DIR != nco_fix_dir):
```

The comparison to a bool isn't necessary.

christinaholtNOAA (approved):
@danielabdi-noaa This is looking really nice. I have just a couple of super minor comments below. 

I think things are ready.

ush/set_cycle_dates.py
gsketefian: I will defer to @chan-hoo on this one since he added this bit.

gsketefian:
@danielabdi-noaa This generally looks good to me.  Since it's hard to look through every line of this new code, it would be nice to at least try it out.  Do you have any instructions on how to run the python version of the workflow generation?  Say I want to run a single experiment with a given experiment configuration file.

danielabdi-noaa:
> 

@gsketefian To use the python workflow generation path, the only thing different is you execute `generate_FV3LAM_wflow.py` instead of the corresponding shell script file `generate_FV3LAM_wflow.sh`. The python workflow path is setup to access a shell script config file `config.sh` by default, even though it can use YAML format too, so there is no other changes you need to make.
It may be better to run both workflow patchs for a specific test case to compare the files generated (xml file, nml file etc). I've run the full regression test on Hera after modifying `run_W2E2_tests.sh` to use `generate_FV3LAM_wflow.py` and got same failures and passes as the shell script path. It would be good to have the test run on other systems.

gsketefian:
@danielabdi-noaa Ok thanks.  I will try it out on Hera and at least one more platform.

gsketefian:
@danielabdi-noaa I've been running tests on both Hera and Orion.  I decided that it's best (and enough) to run the `wflow_features` tests (i.e. no need to run the tests in `grids_extrn_mdls_suites_community` and `grids_extrn_mdls_suites_nco`).  There are a total of 35 of those.  I've run 10 tests on Hera (including the bash counterparts) with the same results as with bash (2 failures), so I'm not worried about the failures so far.  I'm planning on running the rest today.

I've also run 6 tests on Orion with some failures, although I haven't run the bash counterparts there yet.  I had to make some adjustments to get things to run on Orion, but nothing directly related to this PR.  I'm working through those failures now and will do more tests today.

One thing I noticed is that for a pre-existing file or directory that needs to be renamed, the python version adds the date and time to the end of the name instead of "oldNN".  That is good since it identifies when it was renamed.  But can we also add the string "old" somewhere in the rename?  E.g. instead of renaming `specify_DOT_OR_USCORE` to say `specify_DOT_OR_USCORE_20220407_100006`, rename it to `specify_DOT_OR_USCORE_old_20220407_100006`.  That's just to make it clear that it's an older copy.  

I'll keep you posted with more results soon.  Thanks.

danielabdi-noaa:
@gsketefian Thank you for putting the effort to test it, I know it is an unusually big PR!
I will make the changes you suggested for renaming old folders.
One thing I should mention is that the python code could be a couple of commits behind so that could make some difference.
It was upto date last week but haven't checked since then.

I just got access to the wrfruc project jet and niagra so I will do the rull regression tests there too.
I've run a couple of tests on jet and looks to run there just fine.

Thanks again!

gsketefian:
@danielabdi-noaa You're welcome.  And sorry it takes a while for me to get to these big PRs.

I did notice that the python scripts are not completely up-to-date, so I'm testing with slightly older hashes.  In particular, I am not using the latest hashes that went in PR #[233](https://github.com/ufs-community/ufs-srweather-app/pull/233).  If you want to update to the latest, that would be great (I guess you'll have to do that anyway before merging).  Let me know if you do.  There are some tests (specifically with the `GFS_2017_gfdlmp[_regional]` suite that are failing that would be fixed if you merged in the latest from `develop`.

Also, on Orion, I need to update the data that's available to run several of the tests since it's not up-to-date with the latest data.

gsketefian:
@danielabdi-noaa One thing I'm finding on Orion is that after a test finishes (either successfully or not), the cron job for it is not removed from the user's cron table.  The simplest case to demo this is with the `deactivate_tasks` test (which completes successfully).  I ran it with the bash scripts, and the cron job was removed.  Then I tried with python, and it wasn't.  

Do you have an account on Orion?  I imagine it would be hard to debug if you don't.  The interesting thing is that this doesn't happen on Hera, which is very similar to Orion.

Note that on Orion, you have to be on login node 1 to be able to use `cron`.  I can provide more details about this if you like.

danielabdi-noaa:
> @danielabdi-noaa One thing I'm finding on Orion is that after a test finishes (either successfully or not), the cron job for it is not removed from the user's cron table. The simplest case to demo this is with the `deactivate_tasks` test (which completes successfully). I ran it with the bash scripts, and the cron job was removed. Then I tried with python, and it wasn't.
> 
> Do you have an account on Orion? I imagine it would be hard to debug if you don't. The interesting thing is that this doesn't happen on Hera, which is very similar to Orion.
> 
> Note that on Orion, you have to be on login node 1 to be able to use `cron`. I can provide more details about this if you like.

I haven't used my Orion account in a long time, but I will try to reactivate it and see if I can reproduce the problem.
Thanks

gsketefian:
@danielabdi-noaa I imagine this PR will affect the documentation that EPIC is working on one way or another.  I am tagging @gspetro-NOAA , who's doing the documentation for the release.

danielabdi-noaa:
@gsketefian  I have got access to Orion now. Please let me know how I can get login node 1 because all I got was login node 2 which doesn't allow a cron job.

I have already fixed one bug where the temp files `/tmp/t1` and `/tmp/t2` needed for extracting variables set in shell script are shared by different users. You already created those so I was not even able to generate the workflow before the fix ...

gsketefian:
@danielabdi-noaa Ahh, I see.  Ok, one less bug then!  I used to be able to login directly into node 1 from my laptop, but that's no longer possible.  I first login to wherever it puts me (usually it seems to be node 2), then I issue this:
```
ssh -Y gketefia@orion-login-1.hpc.msstate.edu
```

danielabdi-noaa:
@gsketefian Thanks I am able to reproduce the problem now. It looks like deactive_tasks behave similarly on Hera too i.e. a cronjob line is left hanging after job is done. It should be easy to debug now ...

danielabdi-noaa:
@gsketefian I have fixed the crontab issue now, which was due to double quotes being lost in `called_from_cron="TRUE"` when the crontab line is written. So at the end of workflow run  `grep -v` was not able to match the line with CRONTAB_LINE so it was not able to remove it.

Question, there is this line in the workflow generating script which implies that `called_from_cron` could be set during workflow generation.
https://github.com/ufs-community/regional_workflow/blob/3f83cde36876741ec5cf249e70c3c5ee5fd699ec/ush/generate_FV3LAM_wflow.sh#L505
I have it to always take the default value in the python workflow path because I couldn't figure out where it could be set.

gsketefian:
@danielabdi-noaa FYI the `called_from_cron` is needed as an argument to the workflow (re)launch script `launch_FV3LAM_wflow.sh` because of the behavior of `crontab` on Cheyenne.  That's described in more detail in PR #675.  From that PR:

> Adds an optional argument to launch_FV3LAM_wflow.sh that specifies whether it is being called from a cron job. This is necessary because on Cheyenne, the version of crontab to use depends on whether or not the parent script is being called from a cron job.

Anyway, the only place that `called_from_cron` is set to "TRUE" is when `launch_FV3LAM_wflow.sh` is called from the command line.  You can see that from here:
https://github.com/ufs-community/regional_workflow/blob/3f83cde36876741ec5cf249e70c3c5ee5fd699ec/ush/setup.sh#L1441
So that argument is only set internally when a cron job line is being created that will then be written to the user's cron table.  Otherwise, the default value (FALSE) should be used.  So you should be ok, I think.

Can I now go ahead and update my version with your latest?  Thanks.

danielabdi-noaa:
@gsketefian Thanks for the explanation.

> Can I now go ahead and update my version with your latest? Thanks.

Yes, you can try it out now. I have also updated the PR to the latest develop but it is not well tested.
It did run the `deactivate_tasks` properly on Orion though.

Note that I rebased the PR to develop and force pushed so you may need to make a clean clone of the PR.

danielabdi-noaa:
@gsketefian I think Orion is working great now, maybe even better than Hera, so I am putting it in the "tested section".
I have run the whole regression test and got fewer failures than Hera's (only 20).
Moreover, there were no crontab lines hanging around once the test suite finished running.
My test directory on Orion is at:
```
/work/noaa/gsd-hpcs/Daniel.Abdi/benchmark-run
```
Jet, on the other hand, seems to have a lot of missing data especially with latest develop so I am going to put that test on hold ...

willmayfield:
@danielabdi-noaa Thanks for your work this is great!

I tried generating and running the GST case by hand on Cheyenne and ran into some minor issues: 1. while bringing in environment variables, for some reason the (NPL) from the Cheyenne workflow environment prompt ends up causing an error. 2. When EXPT_BASEDIR is an empty string in the configs (as it is by default) indexing EXPT_BASEDIR[0] gives an error in setup.py. 3. MACHINE="cheyenne" fails when it gets compared to the upper-case machine names list in valid_param_vals.yaml.

I also tried it on Orion and still got the same problems 2 and 3 there. If I avoid triggering these checks, both generated and ran successfully!

danielabdi-noaa:
@willmayfield Thank you for testing and finding out issues -- I will try and fix them later!

danielabdi-noaa:
> @danielabdi-noaa Thanks for your work this is great!
> 
> I tried generating and running the GST case by hand on Cheyenne and ran into some minor issues: 1. while bringing in environment variables, for some reason the (NPL) from the Cheyenne workflow environment prompt ends up causing an error. 2. When EXPT_BASEDIR is an empty string in the configs (as it is by default) indexing EXPT_BASEDIR[0] gives an error in setup.py. 3. MACHINE="cheyenne" fails when it gets compared to the upper-case machine names list in valid_param_vals.yaml.
> 
> I also tried it on Orion and still got the same problems 2 and 3 there. If I avoid triggering these checks, it generated and ran successfully on both platforms!

@willmayfield  I believe that I 've fixed the last two issues but can not test the Cheyenne issue because I do not have access to the machine.
I was wondering if you can give me more information to debug the problem. There is a brittle section of code in export/import vars that could be the cause of the problem:

https://github.com/danielabdi-noaa/regional_workflow/blob/python_workflow/ush/python_utils/environment.py#L239-L245

I skip python functions, lowercase environment variables that are unlikely to be variables defined by SRW app, which is what I want to import/export. It could be that something on cheyenne made it through and is causing problems?

JeffBeck-NOAA:
@willmayfield, do you have time to test @danielabdi-noaa's changes to this PR on Cheyenne?  @gsketefian, does this need to be retested on Hera?

danielabdi-noaa:
> @willmayfield, do you have time to test @danielabdi-noaa's changes to this PR on Cheyenne? @gsketefian, does this need to be retested on Hera?

@JeffBeck-NOAA I believe Will has tested and confirmed my latest changes fix the problem on Cheyenne. Our communications shifted to email a couple of days ago.

JeffBeck-NOAA:
@danielabdi-noaa, thanks!  I will go ahead and approve.  @gsketefian, did you want to test one more time on Hera?

JeffBeck-NOAA (approved):

venitahagerty:
Machine: jet
Compiler: intel
Job: WE
Repo location: /lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/876620093/20220425230518/ufs-srweather-app
Build was Successful
Rocoto jobs started
Experiment failed: grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v15p2
2022-04-25 23:42:15 +0000 :: fe2 :: Task get_extrn_ics, jobid=2642517, in state DEAD (FAILED), ran for 15.0 seconds, exit status=256, try=1 (of 1)
ci-jet-intel-WE

PR #717, opened 2022-03-28
Allow user-defined file names for input template files

chan-hoo:
## DESCRIPTION OF CHANGES: 
- Currently, the file names of the input template files such as 'diag_table.[ccpp]', 'field_table.[ccpp]', 'input.nml.FV3', 'model_configure', and 'nems.configure' should be the same as the designated input file names of the ufs weather model. If a user wants to use his/her own template files, the user should change the file names to the designated ones before generating a workflow.
- With this PR, a user does not have to change the file names. Instead, the user-defined template file names can be specified in the configuration file 'config.sh' with the parameter names such as 'DIAG_TABLE_TMPL_FN', 'FIELD_TABLE_TMPL_FN', 'NEMS_CONFIG_TMPL_FN', and 'MODEL_CONFIG_TMPL_FN'.
- Some users at EMC requested to add this feature to the app.

## TESTS CONDUCTED: 
1. WE2E tests on Hera:
- MET_ensemble_verification
- community_ensemble_2mems
- grid_RRFS_CONUS_25km_ics_HRRR_lbcs_HRRR_suite_RRFS_v1beta
- grid_RRFS_CONUS_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha
- nco_grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
- user_defined_templates (new test)

2. GST test using user-defined template files:
- Copied 'diag_table' and 'field_table' to 'diag_table_test' and 'field_table_test'.
- Added DIAG_TABLE_TMPL_FN="diag_table_test" and FIELD_TABLE_TMPL_FN="field_table_test" to 'config.sh'.


RatkoVasic-NOAA (approved):

jwolff-ncar:
For a starting point should these be set in the config.community.sh as well?

jwolff-ncar:
Thanks, Chan-Hoo!

JeffBeck-NOAA:
@chan-hoo, could you clarify why this feature is needed?  The workflow already has the ability to modify all of these files to meet the use cases of any set of CCPP suite, external model data used for ICs/LBCs, resolution, write component grid, etc., so I'm struggling to see what having a separate template file resolves.

chan-hoo:
@JeffBeck-NOAA, currently the template file name can be specified with the parameters DATA_TABLE_FN, DIAG_TABLE_FN, FIELD_TABLE_FN, but a user should set the designated file name of the ufs weather model such as 'data_table', 'field_table', 'filed_table'. Assume that a user wants to test multiple cases for several parameters and the user prepared multiple template files like field_table_case1, field_table_case2, field_table_case3. The user should replace the file name to field_table for each case. This PR allows a user to specify the file name in config.sh. Without this feature, I think we don't need the parameters DATA_TABLE_FN, DIAG_TABLE_FN, FIELD_FN, FV3_NML_BASE_SUITE, and NEMS_CONFIG_FN because the only one file name for each parameter is allowed.

gsketefian:
@chan-hoo This is a good feature to add, but I think a better way to do it would be to make the template file names, e.g. `DATA_TABLE_TMPL_FN`, user-specifiable.  So I would:
1. Define `DATA_TABLE_TMPL_FN` in `config_defaults.sh` and just set it there to a null string.
2. Remove `DATA_TABLE_FN` from `config_defaults.sh`.  In `setup.sh`, define it as `DATA_TABLE_FN="data_table"`.  This is convenient to have in case the name changes in the weather model.  That way you don't have to change it everywhere again.
3. In `setup.sh`, do `DATA_TABLE_TMPL_FN=${DATA_TABLE_TMPL_FN:-${DATA_TABLE_FN}}"` to pick up any user-defined values.

Similar procedure for the other variables.  This would also make clear that it is the names of the template variables that the user is specifying, not the names of the actual files that are read in.

JeffBeck-NOAA:
@chan-hoo, thanks for the explanation.  I certainly understand there are situations where the current template files have hard-wired settings that currently aren't a user-defined option in the config.sh file.  I assume that is the source of this request (create a separate template file with modifications that aren't currently possible to modify).  I'm wondering, if we had user-definable options for all settings, would the functionality in this PR still be needed?  Are there many hard-wired settings in all template files that we don't have represented already in the config_defaults.sh file?

chan-hoo:
@gsketefian, I'll follow your instructions. It looks better.

chan-hoo:
@JeffBeck-NOAA, I've added a user defined option for the input template files of 'diag_table', 'field_table', 'data_table', 'model_configure', and 'nems.configure' to the app. I think all the input template files can be user-defined with this PR.

JeffBeck-NOAA:
> @JeffBeck-NOAA, I've added a user defined option for the input template files of 'diag_table', 'field_table', 'data_table', 'model_configure', and 'nems.configure' to the app. I think all the input template files can be user-defined with this PR.

@chan-hoo, thanks.  Did you see my comment above?  If not, I'll paste it here.  I have no major problem with this feature, I just wanted to understand things a bit more:

"I certainly understand there are situations where the template files have hard-wired settings that currently aren't a user-defined option. I assume that is the source of this request (create a separate template file with modifications that aren't currently possible to modify)? I'm wondering, if we had user-definable options for all settings in each template file, would the functionality in this PR still be needed? Are there many hard-wired settings in all template files that EMC is trying to modify that we don't have represented already in the config_defaults.sh file?"

tests/WE2E/test_configs/wflow_features/config.user_defined_templates.sh
```diff
@@ -0,0 +1,30 @@
+#
```
gsketefian: @chan-hoo To be consistent with the naming used for tests for other user-specified parameters (`specify_...`) and to reduce confusion between template files and template variables (which are something completely different), can we rename this file `config.specify_template_filenames.sh`?  Thanks.

ush/config_defaults.sh
```diff
@@ -343,28 +343,31 @@ DOT_OR_USCORE="_"
...
-#
-# NEMS_CONFIG_FN:
-# Name of file that specifies ???
-#
 # FV3_EXEC_FN:
 # Name to use for the forecast model executable when it is copied from
 # the directory in which it is created in the build step to the executables
 # directory (EXECDIR; this is set during experiment generation).
 #
+# DIAG_TABLE_TMPL_FN:
```
gsketefian, line 351: Typo:  fille --> file

ush/setup.sh
```diff
@@ -1216,11 +1216,18 @@ fi
 #
 dot_ccpp_phys_suite_or_null=".${CCPP_PHYS_SUITE}"
 
-DATA_TABLE_TMPL_FN="${DATA_TABLE_FN}"
-DIAG_TABLE_TMPL_FN="${DIAG_TABLE_FN}${dot_ccpp_phys_suite_or_null}"
-FIELD_TABLE_TMPL_FN="${FIELD_TABLE_FN}${dot_ccpp_phys_suite_or_null}"
-MODEL_CONFIG_TMPL_FN="${MODEL_CONFIG_FN}"
-NEMS_CONFIG_TMPL_FN="${NEMS_CONFIG_FN}"
+# Designated name as the input files of the forecast model (ufs-weather-model)
```
gsketefian: @chan-hoo To be clearer about what these variables represent and if/when they should be changed, can you modify the comment to something like this?  Thanks.
```
# Names of input files that the forecast model (ufs-weather-model) expects 
# to read in.  These should only be changed if the input file names in the 
# forecast model code are changed.
```

ush/config_defaults.sh
```diff
@@ -343,28 +343,31 @@ DOT_OR_USCORE="_"
...
+# Name of a template fille that specifies the output fields of the forecast model
+# (ufs-weather-model: diag_table) followed by [dot_ccpp_phys_suite]. Its default
```
gsketefian: @chan-hoo `designated` is not clear to me.  Can we change `default value is the same as the designated file name` to `default value is the name of the file the weather model expects to read in`?  Same for the other three files.

Btw, thanks for filling in these descriptions!

gsketefian (approved):
I requested a few minor fixes, otherwise looks good!  Approving.

chan-hoo:
@JeffBeck-NOAA, I think so. This function will be useful when a user wants to test one user-defined variable with other possible values to find its optimal one. With this PR, a user doesn't need to change the one template file and remember all the values, but need to change the template name in `config.sh`.

PR #740, opened 2022-04-18
Add more parameters to CSV file containing WE2E test info

gsketefian:
## DESCRIPTION OF CHANGES: 
The script/function `get_WE2Etest_names_subdirs_descs.sh` (which is called from `run_WE2E_tests.sh` if needed) creates a CSV (Comma-Separated Value) file named `WE2E_test_info.csv` that contains information about the WE2E tests.  Currently, this CSV file contains only 3 columns: the test name, the names of any alternate names for the test, and the test description.  In order to have a more complete summary of the WE2E tests, this PR modifies `get_WE2Etest_names_subdirs_descs.sh` so that additional information is included in the CSV file.  This additional information consists of the values of the following experiment variables for each test:
```
PREDEF_GRID_NAME
CCPP_PHYS_SUITE
EXTRN_MDL_NAME_ICS
EXTRN_MDL_NAME_LBCS
DATE_FIRST_CYCL
DATE_LAST_CYCL
CYCL_HRS
INCR_CYCL_FREQ
FCST_LEN_HRS
LBC_SPEC_INTVL_HRS
NUM_ENS_MEMBERS
```
In addition, the script uses this information to calculate the number of times each test calls the forecast model (e.g. if the test uses 3 different cycle dates, then the forecast model will be called 3 times; if it is an ensemble test for a single cycle, the test will call the forecast model as many times as the number of ensemble members).  

## TESTS CONDUCTED: 
The script `run_WE2E_tests.sh` was called that in turn calls `get_WE2Etest_names_subdirs_descs.sh`.  This created a new CSV file that contained the new fields (columns).  The CSV file was imported into Google Sheets (using "|" as the field/column separator) and looked correct.

## DOCUMENTATION:
The documentation is for the most part already within the `get_WE2Etest_names_subdirs_descs.sh`.  This PR slightly modifies that documentation to update it.


scripts/exregional_make_grid.sh
```diff
@@ -410,7 +410,7 @@ if [ "${GRID_GEN_METHOD}" = "GFDLgrid" ]; then
 elif [ "${GRID_GEN_METHOD}" = "ESGgrid" ]; then
   CRES="C${res_equiv}"
 fi
-set_file_param "${GLOBAL_VAR_DEFNS_FP}" "CRES" "\"$CRES\""
+set_file_param "${GLOBAL_VAR_DEFNS_FP}" "CRES" "'$CRES'"
```
gsketefian, line 413: This line updates the value of `CRES` in `var_defns.sh`.  This needs to be updated from double quotes to single quotes because the rest of the variables in `var_defns.sh` have recently been modified to use single quotes.

chan-hoo (approved):

JeffBeck-NOAA (approved):

PR #743, opened 2022-04-20
Update vertical structure of NCO mode

chan-hoo:
## DESCRIPTION OF CHANGES: 
The vertical structure of the NCO mode is updated to meet the latest version of the NCO standards (version 11.0.0, 01/19/2022).

## TESTS CONDUCTED: 
- WE2E tests:
MET_ensemble_verification
community_ensemble_2mems
grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15_thompson_mynn_lam3km
grid_RRFS_CONUScompact_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_RRFS_v1beta
grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha
grid_RRFS_NA_3km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1alpha
nco_ensemble
nco_grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
nco_grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15_thompson_mynn_lam3km
nco_grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR


ush/config_defaults.sh
```diff
@@ -269,39 +280,24 @@ EXEC_SUBDIR="bin"
...
 #
-COMINgfs="/base/path/of/directory/containing/gfs/input/files"
+COMIN="/path/of/directory/containing/data/files/for/IC/LBCS"
 FIXLAM_NCO_BASEDIR=""
-STMP="/base/path/of/directory/containing/model/input/and/raw/output/files"
-NET="rrfs"
 envir="para"
-RUN="experiment_name"
+NET="rrfs"
+model_ver="v1.0"
```
christinaholtNOAA: @chan-hoo Will you provide some guidance on how this versioning is determined?

chan-hoo: @christinaholtNOAA, there is no specific guideline about this in the nco standards. @JacobCarley-NOAA, @BenjaminBlake-NOAA , do you have an answer to this?

BenjaminBlake-NOAA: @christinaholtNOAA  I believe it's related to the number of operational implementations that have occurred.  Since RRFS is not operational yet, model_ver = "v1.0".  For the HRRR, we are on version 4, and with each successive WCOSS1 update we went from hrrr.v4.0.0 to hrrr.v4.0.14.  But on WCOSS2, NCO is using hrrr.v4.1, so model_ver = "v4.1".  I'm not sure why they are now using 2 digits when they had been using 3 digits for the version numbers though.

chan-hoo: @BenjaminBlake-NOAA, Thank you for the explanation. You're correct. The exact wording in the nco standard is "version number of package in **three** digits".  I'll update the comment and default value to "v.1.0.0".

christinaholtNOAA: @chan-hoo @BenjaminBlake-NOAA Thanks for the replies. That honestly doesn't clear things up too much for me. I'm understand that a frozen version in operations is going to correspond to a given number, but that number is only really applicable to the develop branch for a brief moment in time (if that is even true). Do we have v1.0 until this goes into ops, then we bump it to be working on v2.0 since it would be newer than what is running in operations?

BenjaminBlake-NOAA: @christinaholtNOAA From my point of view I think the idea of using v1.0 until it goes into operations makes sense, and then we would increase the version to v2.0 after that.  But maybe that doesn't make as much sense for ongoing development.  Would it make more sense to increase the version number incrementally with time?  For example, v1.0 becomes v1.1, then v1.2, etc. until RRFS is implemented and then we move up to v2.0.

christinaholtNOAA: @BenjaminBlake-NOAA Would NCO accept that we were putting v1.17 (for example) in operations as the first implementation?

BenjaminBlake-NOAA: @christinaholtNOAA  I don't think they would accept that, so the final code we deliver to NCO will need to use v1.0, but I don't see an issue with using different version numbers during the development process.

BenjaminBlake-NOAA (approved):

RatkoVasic-NOAA (approved):

PR #747, opened 2022-04-26
Add github actions for python unittests.

danielabdi-noaa:
## DESCRIPTION OF CHANGES: 
- Adds github actions for python unittests in python_utils/ and ush/ directories.

## TESTS CONDUCTED: 

No tests needed other than to confirm it works.
Runs successfully on current develop, also tested that it fails when there is a bug.

Last successful run on my fork of regional_workflow:
https://github.com/danielabdi-noaa/regional_workflow/runs/6281838521?check_suite_focus=true

Last failure due to a bug
https://github.com/danielabdi-noaa/regional_workflow/runs/6281719721?check_suite_focus=true

It also runs on the NOAA-GSL repository, from which I forked, so it is probably because I haven't
forked directly from ufs-community that the tests are not running here.
https://github.com/NOAA-GSL/regional_workflow/runs/6322450639?check_suite_focus=true

## DEPENDENCIES:
None

## DOCUMENTATION:
None

## ISSUE (optional): 
None


.github/workflows/python_unittests.yaml
```diff
@@ -0,0 +1,32 @@
...
+        run: |
+          cd ush
+          python3 -m unittest -b python_utils/test_python_utils.py
+          python3 -m unittest -b check_ruc_lsm.py create_diag_table_file.py \
+                     create_model_configure_file.py link_fix.py set_cycle_dates.py \
+                     set_extrn_mdl_params.py set_FV3nml_sfc_climo_filenames.py \
+                     set_FV3nml_ens_stoch_seeds.py set_gridparams_ESGgrid.py \
+                     set_gridparams_GFDLgrid.py set_ozone_param.py \
+                     set_predef_grid_params.py set_thompson_mp_fix_files.py \
+                     get_crontab_contents.py
```
christinaholtNOAA: Will this work if there is just a pattern like `*.py` so we don't have to keep track of which files have tests in them?

christinaholtNOAA (approved):
Just a quick question/suggestion for a minor change. Otherwise, this looks great and it's nice to see a set of quick/easy tests on the workflow that can be run on a GHA runner!

christinaholtNOAA:
@danielabdi-noaa I was surprised to see that the tests didn't actually run for this branch. Will you need an admin to turn something on, or does the workflow file need to set a specific branch (`develop`) before it will run the test?

.github/workflows/python_unittests.yaml
danielabdi-noaa: I had issues with some python scripts in `ush` dir and unittests. On hera, for example, numpy and matplotlib needed by `mrms_pull_topofhour.py` and `UFS_plot_domains.py`, are missing even after activating `wflow_hera.env`, so unittest fails. Moreover, with the former there is an issue with scripts that accept arguments but don't have unittest class implemented -- which I have fixed now. `UFS_plot_domains.py` uses deprecated plotting libraray `Basemap` from `mpl_toolkits` so need to figure out a way to install that for github actions. Or I could simply include all python scripts except problematic ones with some sort of wildcard ...

danielabdi-noaa:
@christinaholtNOAA Yes there must be something blocking it because it is running on my fork, so it would be good if someone take a look at that. I will try and see if specifying a branch makes a difference ...

.github/workflows/python_unittests.yaml
danielabdi-noaa: Ok all is good now. It doesn't fail with `*.py` but no real unittests are done on python scripts not related to the workflow generation.

christinaholtNOAA:
It looks like because this PR is open from a fork and no other Actions exist, it won't run the Actions. 

https://github.community/t/first-pr-with-github-actions-wont-get-built/18183

@danielabdi-noaa I see that you ran it workflow in your fork. Maybe other here will be able to review that and make a decision here?

danielabdi-noaa:
@christinaholtNOAA Thanks for finding out the issue. Also, I forked from NOAA-GSL but did not do the PR there, so that is one more level of obstacle. I have updated the PR description with the link to the last successful run on my fork.

mkavulich (approved):
Looks good to me, I look forward to seeing these tests in action!

PR #749, opened 2022-04-27
[ens_design_RRFS] Modifications for user-specified MET INPUT/OUTPUT directories and breaking out MET's ensemble-stat tool

michelleharrold:
## DESCRIPTION OF CHANGES: 
In METv10.1, the ensemble-stat tool has been split into two distinct tools: gen-ens-prod (new tool; creates ensemble products) and ensemble-stat (calculates ensemble stats). These changes break out the current functionality (solely using ensemble-stat) to use both gen-ens-prod and ensemble-stat. 

In addition, functionality was added for the user to provide input and output locations for METplus verification via the config.sh.

Note, this PR is for the ens_design_RRFS branch.

## TESTS CONDUCTED: 
Ran test case (2019061500) with GFSv15.2 physics suite for a 3 member (SPPT) ensemble and compared baseline/current capability to new capability to show results are the same using the gen-ens-prod tool.

Ran a test case (2021053000) with 9-member RRFS ensemble data staged in a different directory than where the MET output was written to test MET INPUT/OUTPUT directory functionality. @willmayfield also ran a test with the 9-member RRFS ensemble data as a sanity check.

## DEPENDENCIES:
None.

## DOCUMENTATION:
When gen-ens-prod capability is included in develop, the documentation will be updated accordingly.

## ISSUE (optional): 
Related to issue #709 (eventually this will be merged into develop).

## CONTRIBUTORS (optional): 
@jwolff-ncar


ush/setup.sh
```diff
@@ -1163,6 +1171,9 @@ FIXam="${EXPTDIR}/fix_am"
 FIXclim="${EXPTDIR}/fix_clim"
 FIXLAM="${EXPTDIR}/fix_lam"
 
+MET_INPUT_DIR="${EXPTDIR}"
```
gsketefian: I think if you want it to behave as you describe in the comment above, this should be changed to:
```
MET_INPUT_DIR="${MET_INPUT_DIR:-$EXPTDIR}"
```
**AND** its value in `config_defaults.sh` should be set to an empty string, i.e.
```
MET_INPUT_DIR=""
```
(so that if the user doesn't define it, it gets set to `$EXPTDIR` instead of the random value `"${EXPT_BASEDIR}/${EXPT_SUBDIR}"` in `config_defaults.sh`).  Same for `MET_OUTPUT_DIR` below.

ush/config_defaults.sh
```diff
@@ -585,6 +593,8 @@ MODEL=""
 MET_INSTALL_DIR=""
 MET_BIN_EXEC="bin"
 METPLUS_PATH=""
+MET_INPUT_DIR="${EXPT_BASEDIR}/${EXPT_SUBDIR}"
```
gsketefian: To match what's happening in `setup.sh` and since `EXPT_BASEDIR` and `EXPT_SUBDIR` aren't yet defined to their final values in this file, set `MET_INPUT_DIR` to just a null string, i.e.
```
MET_INPUT_DIR=""
```
Same for `MET_OUTPUT_DIR`.

ush/setup.sh
```diff
@@ -1130,6 +1130,14 @@ check_for_preexist_dir_file "$EXPTDIR" "${PREEXISTING_DIR_METHOD}"
 # This is the directory that will contain the fixed files or symlinks to
 # the fixed files containing the grid, orography, and surface climatology
 # on the native FV3-LAM grid.
+
```
gsketefian: I'd remove this block of comments for `MET_INPUT_DIR` and `MET_OUTPUT_DIR` since they are already defined in `config_defaults.sh` (to keep their definitions in one place only).

michelleharrold: Done!

ush/config_defaults.sh
michelleharrold: Thank you for the suggested changes -- I went ahead and made the changes!

ush/setup.sh
michelleharrold: Thank you for the suggested changes -- I went ahead and made the changes!

willmayfield (approved):
Looks good to me!

scripts/exregional_run_pointstatvx.sh
```diff
@@ -99,12 +99,12 @@ export fhr_list
...
 if [[ ${DO_ENSEMBLE} == "FALSE" ]]; then
-  INPUT_BASE=${EXPTDIR}/${CDATE}/postprd
-  OUTPUT_BASE=${EXPTDIR}/${CDATE}
+  INPUT_BASE=${MET_INPUT_BASE}
+  OUTPUT_BASE=${MET_OUTPUT_BASE}
   LOG_SUFFIX=pointstat_${CDATE}
 elif [[ ${DO_ENSEMBLE} == "TRUE" ]]; then
-  INPUT_BASE=${EXPTDIR}/${CDATE}/${SLASH_ENSMEM_SUBDIR}/postprd
-  OUTPUT_BASE=${EXPTDIR}/${CDATE}/${SLASH_ENSMEM_SUBDIR}
+  INPUT_BASE=${MET_INPUT_BASE}/${SLASH_ENSMEM_SUBDIR}/postprd
```
gsketefian: `SLASH_ENSMEM_SUBDIR` already includes the leading "/" character, so toj avoid double-slashes (although it doesn't make a difference in functionality), you can change to:
```
INPUT_BASE=${MET_INPUT_BASE}${SLASH_ENSMEM_SUBDIR}/postprd
```
Same for `OUTPUT_BASE`.

scripts/exregional_run_ensgridvx.sh
```diff
@@ -103,6 +103,16 @@ export fhr_list
 
 NUM_PAD=${NDIGITS_ENSMEM_NAMES}
 
+#
+#-----------------------------------------------------------------------
+#
+# Create INPUT_BASE and OUTPUT_BASE to read into METplus conf files.
+#
+#-----------------------------------------------------------------------
+#
```
gsketefian, line 112: It seems like for all (many of?) the vx tasks, `MET_INPUT_BASE` is set in the xml to `MET_INPUT_DIR`, then `INPUT_BASE` is set to `MET_INPUT_BASE` here.  Why not just remove `MET_INPUT_BASE` from the xml and here and just set `INPUT_BASE` here to `MET_INPUT_DIR` (which is defined in `var_defns.sh` and is available here because `var_defns.sh` gets sourced at the beginning of this file (by the line `. ${GLOBAL_VAR_DEFNS_FP}`)?  `MET_INPUT_BASE` seems like an unnecessary middleman (much like CIRES is between me and NOAA ;) ).  Or I may have missed something.

Same for the `...OUT...` variables and all (most of?) the other vx tasks.

ush/setup.sh
```diff
@@ -1163,6 +1163,9 @@ FIXam="${EXPTDIR}/fix_am"
 FIXclim="${EXPTDIR}/fix_clim"
 FIXLAM="${EXPTDIR}/fix_lam"
 
+MET_INPUT_DIR="${MET_INPUT_DIR:-$EXPTDIR}"
+MET_OUTPUT_DIR="${MET_INPUT_DIR:-$EXPTDIR}"
```
gsketefian: Bug here in setting `MET_OUTPUT_DIR` (right-hand side should say `OUT`, not `IN`).  I fixed it in my copy of the branch and will push it.  Feel free to fix it too if you like, I'll resolve any conflicts on my end.

scripts/exregional_run_enspointvx.sh
```diff
@@ -98,14 +98,24 @@ export fhr_last
 fhr_list=`echo ${FHR} | $SED "s/ /,/g"`
 export fhr_list
 
+#
+#-----------------------------------------------------------------------
+#
+# Create INPUT_BASE to read into METplus conf files.
+#
+#-----------------------------------------------------------------------
+#
```
gsketefian, line 107: This line caused an error during my testing because `MET_INPUT_BASE` is not defined for this task in the xml.  I commented out this line in my changes (that I'm going to push to your branch), but I'm not sure if instead `MET_INPUT_BASE` should be defined in the xml (depends on if `INPUT_BASE` is used in this task).  You may have to undo my changes...

gsketefian: I just noticed that below, you added `INPUT_BASE` to the list of variables that get exported, so I assume it is needed.  I won't comment out this line and leave it up to you to define it in the xml.

scripts/exregional_run_ensgridvx.sh
michelleharrold: I can definitely make the change to remove INPUT/OUTPUT dir from the XML. I had some initial concerns in my head about how to implement this functionality when I first got started on this, but I think what you lay out makes sense.

ush/setup.sh
gsketefian: Thanks!

gsketefian:
@michelleharrold Are you planning on merging this into the `develop` branch in time for the release?  I have an issue here (#751) where I'm tracking which PRs that are/have been going into `develop` also need to be merged into the release branch (which has already been created).  I can add this PR to the list if you like.

michelleharrold:
@gsketefian -- this PR is not intended for develop before the release. Eventually, we will get these features into develop, but for the time being they are for the ens_design_RRFS branch we are using for the DTC ensemble task.

JeffBeck-NOAA (approved):

PR #744, opened 2022-04-20
Standardize static data across Tier-1 platforms; fix and improve IC and LBC data retrieval

mkavulich:
## DESCRIPTION OF CHANGES: 
This PR implements the new static input data directory structures as described in [UFS SRW App Issue 231](https://github.com/ufs-community/ufs-srweather-app/issues/231). 

In addition, it fixes some issues with retrieving staged model input data from disk, and gives a way to specify whether or not to check alternate sources of data if not found on disk (HPSS, AWS). These flags have different defaults for different platforms based on data availability (for example, Cheyenne does not have NOAA HPSS access, and has no internet access on compute nodes, so both alternate sources have been turned off by default), and for WE2E tests that expect data staged on disk, these flags have been turned off.

Finally, some quality-of-life changes have been implemented, such as typo fixes, and reducing the number of tries for retrieving data from NOMADS from 20 failures (!) to 2.

## TESTS CONDUCTED: 
All tests run to completion on Hera (aside from known failures as documented in #728, #729, and #731). 

Tests on Cheyenne, Orion, and Jet are pending.

## DEPENDENCIES:
There are no PR dependencies for this PR, but data will need to be staged on disk on each platform before tests will run successfully. So far, data has been staged on Hera and Cheyenne, and should be staged

I will need assistance in getting data to other platforms from others who have access to those platforms.

## DOCUMENTATION:
Documentation for these changes and the new directory structure can be found in the [ufs-srweather-app issue #231](https://github.com/ufs-community/ufs-srweather-app/issues/231)

I was not sure what the best place to include this for version control is, but I wrote up a README file for the new "input_model_data" directory, attached here: 
[README_input_model_data.txt](https://github.com/ufs-community/regional_workflow/files/8523247/README_input_model_data.txt)

## ISSUE (optional): 
 - Solves #700 
 - Solves #715 
 - Solves #741
 - Solves #742
 - Solves https://github.com/ufs-community/ufs-srweather-app/issues/231

## CONTRIBUTORS (optional): 
@christinaholtNOAA

PR #750, opened 2022-04-28
Fix typo and crontab issue on wcoss dell in workflow python scripts

chan-hoo:
## DESCRIPTION OF CHANGES: 
- Fix typo in 'setup.py'.
- Fix crontab issue on wcoss_dell_p3 in the workflow python scripts.
- Remove capture_output as PR #744.

## TESTS CONDUCTED: 
- 6h GST tests with FV3GFS_v15p2 and FV3GFS_v16 on wcoss_dell_p3

### DIFF
--- a/ush/get_crontab_contents.py
+++ b/ush/get_crontab_contents.py
@@ -38,12 +38,9 @@ def get_crontab_contents(called_from_cron):
     print_input_args(locals())
   
     #import all env vars
-    IMPORTS = ["MACHINE"]
+    IMPORTS = ["MACHINE", "USER"]
     import_vars(env_vars=IMPORTS)
 
-    # Get username on current machine
-    global USER
-    USER = os.getlogin()
     #
     # Make sure called_from_cron is set to a valid value.
     #
@@ -82,5 +79,5 @@ class Testing(unittest.TestCase):
         self.assertEqual(crontab_cmd, "crontab")
     def setUp(self):
         define_macos_utilities();
-        set_env_var('DEBUG','FALSE')
+        set_env_var('DEBUG',False)
         set_env_var('MACHINE', 'HERA')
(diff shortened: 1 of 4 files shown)


ush/get_crontab_contents.py
```diff
@@ -40,7 +40,10 @@ def get_crontab_contents(called_from_cron):
     #import all env vars
     IMPORTS = ["MACHINE"]
     import_vars(env_vars=IMPORTS)
-  
+
+    # Get username on current machine
+    global USER
```
danielabdi-noaa: A minor comment: `USER` doesn't have to be global unless we want to export its value here and use it in other files as well. Maybe it is a good idea to define env variables that come pre-defined with shell, such as `USER`, `PATH` etc in one central place, similar to the way `SED` is defined in `define_macos_utilities.py`

danielabdi-noaa (approved):
@chan-hoo Thanks for fixes. It looks good to me so I will approve!
I suppose you were able to run some test cases after these fixes, right? WCOSS has not been tested before with the python workflow generator.

ush/get_crontab_contents.py
chan-hoo: @danielabdi-noaa, I'll try it and let you know if I have any problems.

chan-hoo: @danielabdi-noaa, I put "set_env_var('USER','USER')" in `define_macos_utilities.py`, but it didn't work, I got  "   (_,__crontab_contents__,_)=run_command(f'''cat "/u/{USER}/cron/mycrontab"''')
NameError: name 'USER' is not defined". I don't think I understood your comment correctly. Can you make this change on your end? I've sent a github collaborator invite to you.

danielabdi-noaa: I think `USER`, `HOME` etc were already in the environment so I just added it to the IMPORTS list at the beginning of `get_crontab_contents.py`. Let me know if that works now. Thanks!

chan-hoo: Thank you! I'll test it and let you know.

chan-hoo: @danielabdi-noaa, it worked! Thank you for your help!

gsketefian (approved):

PR #758, opened 2022-05-05
Change output file name of run_post to meet NCO standards

chan-hoo:
## DESCRIPTION OF CHANGES: 
- The output file name of the 'run_post' task is changed to meet the latest version of the NCO standards (Ver 11.0.0; January 19, 2022; page 6).
- The file name should be "{NET}.t{cyc}z.[var_info].fHHH.{domain}.grib2".
- The python scripts for workflow generation are updated with the recent changes in bash scripts.

## TESTS CONDUCTED: 
- WE2E tests:
community_ensemble_2mems
grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15_thompson_mynn_lam3km
grid_RRFS_CONUScompact_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_RRFS_v1beta
grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha
grid_RRFS_NA_3km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1alpha
MET_ensemble_verification
MET_verification
nco_grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
nco_grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15_thompson_mynn_lam3km
nco_grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR


RatkoVasic-NOAA (approved):

ush/config_defaults.sh
```diff
@@ -256,6 +256,10 @@ EXEC_SUBDIR="bin"
 #   Name of model run (third level of com directory structure).
 #   In general, same as $NET
 #
+#   FN_OUT_DOMAIN:
```
gsketefian: Two comments about the name of this variable:
1) Since this does not represent the full file name, it would be better to remove the "FN" from the name of this variable.
2) It would be clearer if there was something in the name that identifies what file name this is used in, i.e. that it is used in naming the UPP output files.

So maybe something like `UPP_OUTPUT_DOMAIN_NAME`.

Also, in the description of the variable, say it is the output file from UPP.  Might be obvious to you but probably not for a novice user.  Thanks!

chan-hoo: @gsketefian, agreed. I'll modify this soon.

scripts/exregional_run_fcst.sh
```diff
@@ -539,15 +538,15 @@ if [ ${WRITE_DOPOST} = "TRUE" ]; then
     post_mn=${post_time:10:2}
     post_mn_or_null=""
     post_fn_suffix="GrbF${fhr_d}"
-    post_renamed_fn_suffix="f${fhr}${post_mn_or_null}.${tmmark}.grib2"
+    post_renamed_fn_suffix="f${fhr}${post_mn_or_null}.${FN_OUT_DOMAIN}.grib2"
```
JeffBeck-NOAA, line 542: @chan-hoo, is there a reason to have "post_fn_suffix" and "post_renamed_fn_suffix"?

scripts/exregional_run_post.sh
```diff
@@ -314,7 +303,7 @@ fids=( "prslev" "natlev" )
 for fid in "${fids[@]}"; do
   FID=$(echo_uppercase $fid)
   post_orig_fn="${FID}.${post_fn_suffix}"
-  post_renamed_fn="${NET}.t${cyc}z.${fid}${post_renamed_fn_suffix}"
+  post_renamed_fn="${NET}.t${cyc}z.${fid}.${post_renamed_fn_suffix}"
   mv_vrfy ${tmp_dir}/${post_orig_fn} ${post_renamed_fn}
```
JeffBeck-NOAA, line 307: Same question as above, but this time post_orig_fn vs post_renamed_fn.

scripts/exregional_run_fcst.sh
JeffBeck-NOAA, line 542: Do you have an example of what a full post file name looks like after this change?  Does it still have the "prslev" and "natlev" names in the files?

ush/config_defaults.sh
chan-hoo: @gsketefian, it was renamed to `POST_OUTPUT_DOMAIN_NAME` for UPP as well as inline post.

scripts/exregional_run_fcst.sh
chan-hoo: @JeffBeck-NOAA, when I added this part for the inline post option, I copied the corresponding part in 'exregional_run_post.sh' as you pointed below. I am not the original developer of the script, but I can interpret them as follows: (1) The output file name of UPP and inline post is 'post_orig_fn' including 'post_fn_suffix'. (2) The output file name required by NCO is 'post_renamed_fn' including 'post_renamed_fn_suffix'. (3) File (1) is renamed to File (2), and the original file name (1) is soft-linked to file (2).

scripts/exregional_run_post.sh
chan-hoo: Explained above.

scripts/exregional_run_fcst.sh
chan-hoo: It should be `${NET}.t{cyc}z.[var_info].fHHH.${POST_OUTPUT_DOMAIN_NAME}.grib2` (described in 'config_default.sh'). Yes, "prslev" or "natlev" is placed in [var_info]. For example, "rrfs.t00z.prslev.f003.conus_3km.grib2"

ush/config_defaults.sh
gsketefian: Thanks @chan-hoo!

ush/setup.sh
```diff
@@ -1174,6 +1177,9 @@ else
   COMROOT=""
   COMOUT_BASEDIR=""
 fi
+
+POST_OUTPUT_DOMAIN_NAME="${POST_OUTPUT_DOMAIN_NAME:-${PREDEF_GRID_NAME}}"
+POST_OUTPUT_DOMAIN_NAME=$(echo_lowercase ${POST_OUTPUT_DOMAIN_NAME})
```
gsketefian, line 1191: @chan-hoo If someone is using a custom grid (in which case `PREDEF_GRID_NAME` will be set to an empty string) and forgets to explicitly set `POST_OUTPUT_DOMAIN_NAME` in `config.sh`, then this code will leave `POST_OUTPUT_DOMAIN_NAME` empty.  It would be good to put in a check for that after this and if it is empty, tell the user to set `POST_OUTPUT_DOMAIN_NAME` in their `config.sh` file (unless they are neither running inline post nor the `run_post` metatask). [I guess all comments for the bash scripts also apply to the python ones, but I'm not sure if you're going to change the python or @danielabdi-noaa.]

ush/setup.sh
```diff
@@ -2499,6 +2505,14 @@ RELATIVE_LINK_FLAG='${RELATIVE_LINK_FLAG}'
 #
 #-----------------------------------------------------------------------
 #
+# Domain name used in naming the output files by UPP or inline post
+#
+#-----------------------------------------------------------------------
+#
+POST_OUTPUT_DOMAIN_NAME='${POST_OUTPUT_DOMAIN_NAME}'
```
gsketefian: @chan-hoo I think you don't need this because `POST_OUTPUT_DOMAIN_NAME` is already included in `var_defns.sh` (because all variables defined in `config_defaults.sh` get written to `var_defns.sh` (but with updated values)).  You can check your runs and with this line still included, you'll see two definitions of `POST_OUTPUT_DOMAIN_NAME` in `var_defns.sh` (you just want the first one).

ush/setup.sh
chan-hoo: @gsketefian, if `POST_OUTPUT_DOMAIN_NAME` is not set in `config.sh', the `PREDEF_GRID_NAME` will be used as the domain name by default.

ush/setup.sh
chan-hoo: @gsketefian, agreed. I'll remove it.

chan-hoo: Done.

ush/setup.sh
chan-hoo: @gsketefian, I got it. I'll add a check for the case.

chan-hoo: Done.

ush/setup.sh
gsketefian: Thanks!

scripts/exregional_run_fcst.sh
```diff
@@ -539,15 +538,15 @@ if [ ${WRITE_DOPOST} = "TRUE" ]; then
     post_mn=${post_time:10:2}
     post_mn_or_null=""
     post_fn_suffix="GrbF${fhr_d}"
-    post_renamed_fn_suffix="f${fhr}${post_mn_or_null}.${tmmark}.grib2"
+    post_renamed_fn_suffix="f${fhr}${post_mn_or_null}.${POST_OUTPUT_DOMAIN_NAME}.grib2"
```
gsketefian, line 541: @chan-hoo From this code, it looks like `POST_OUTPUT_DOMAIN_NAME` is used regardless of whether running in NCO more or community mode.  Is that correct?  If so, can you move where it's defined in `config_defaults.sh` so that it's not in the NCO-only section?  Thanks.

chan-hoo: @gsketefian, yes. I'll move it to common area.

chan-hoo: Done.

ush/setup.sh
danielabdi-noaa: @gsketefian I am working on a branch that will remove the shell workflow generation. In the mean time, it would be very helpful if developers run both the shell and python workflow generator when running test cases -- although I will run the regression tests on platforms I have access too. Thanks @chan-hoo for making changes to both workflows!

gsketefian: @chan-hoo Instead of 
```
if [ -z "${POST_OUTPUT_DOMAIN_NAME}" ] && [ -z "${PREDEF_GRID_NAME}" ]; then
```
you can just use
```
if [ -z "${POST_OUTPUT_DOMAIN_NAME}" ]; then
```
since at this point, `POST_OUTPUT_DOMAIN_NAME` will be null only if `PREDEF_GRID_NAME` is also null.

Also, for the message, it would be good to give a bit more info to the user, something like:
```
The domain name used in naming the UPP output files (POST_OUTPUT_DOMAIN_NAME)
has not been set:
  POST_OUTPUT_DOMAIN_NAME = \"${POST_OUTPUT_DOMAIN_NAME}\""
If this experiment is not using a predefined grid (i.e. if PREDEF_GRID_NAME 
is set to a null string), POST_OUTPUT_DOMAIN_NAME must be set in the SRW 
App's configuration file (\"${EXPT_CONFIG_FN}\")."
```

gsketefian: @danielabdi-noaa Sounds good, thanks for letting us know!

gsketefian (approved):

danielabdi-noaa (approved):

danielabdi-noaa:
> @danielabdi-noaa, idx = EXTRN_MDL_SOURCE_BASEDIR_ICS.find("$") doesn't work correctly. @gsketefian, how can we interpret "${EXTRN_MDL_SOURCE_BASEDIR_ICS%%$*}"? print(os.path.exists(EXTRN_MDL_SOURCE_BASEDIR_ICS)) => true, however, print(os.path.exists(EXTRN_MDL_SOURCE_BASEDIR_ICS[:idx])) => false

What is the example path you are using for EXTRN_MDL_SOURCE_BASEDIR_ICS ? My understanding from the comment is that the path is a "template" so it could have "$" sign in it. For example

``` Python
>>> XX="/path/to/$some/where"
>>> idx=XX.find("$")
>>> print(XX[:idx])
/path/to/
```

And equivalent bash
```Bash
$ XX="/path/to/\$some/where"
$ echo ${XX%%\$*}
/path/to/
```

danielabdi-noaa:
@chan-hoo I think the problem is because the path you are testing is not templated and doesn't have the "$" sign.
So we need to do a check on the value of `idx==-1`, in which case `idx=len(EXTRN_MDL_SOURCE_BASEDIR_ICS)` instead.

PR #764, opened 2022-05-11
Feature/remove shell workflow

danielabdi-noaa:
## DESCRIPTION OF CHANGES: 
This PR removes the shell workflow.

## TESTS CONDUCTED: 
I have run the tests successfully on **HERA** and **ORION** and **JET**. The set of tests are  taken a recent shell workflow generation run done in #744 

Out of 56 tests

- community_ensemble_008mems
- community_ensemble_2mems
- deactivate_tasks
- grid_CONUS_25km_GFDLgrid_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
- grid_CONUS_3km_GFDLgrid_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
- grid_RRFS_AK_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
- grid_RRFS_AK_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
- grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
- grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_2017_gfdlmp
- grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_2017_gfdlmp
- grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v15p2
- grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_HRRR
- grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_RRFS_v1beta
- grid_RRFS_CONUScompact_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
- grid_RRFS_CONUScompact_13km_ics_HRRR_lbcs_RAP_suite_HRRR
- grid_RRFS_CONUScompact_13km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha
- grid_RRFS_CONUScompact_13km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
- grid_RRFS_CONUScompact_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
- grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_HRRR
- grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_RRFS_v1beta
- grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR
- grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha
- grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
- grid_RRFS_CONUScompact_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
- grid_RRFS_CONUScompact_3km_ics_HRRR_lbcs_RAP_suite_GFS_v15p2
- grid_RRFS_CONUScompact_3km_ics_HRRR_lbcs_RAP_suite_HRRR
- grid_RRFS_CONUScompact_3km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha
- grid_RRFS_CONUScompact_3km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
- grid_RRFS_NA_13km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta
- grid_RRFS_SUBCONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
- grid_RRFS_SUBCONUS_3km_ics_HRRR_lbcs_RAP_suite_GFS_v15p2
- grid_SUBCONUS_Ind_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
- grid_SUBCONUS_Ind_3km_ics_HRRR_lbcs_RAP_suite_HRRR
- grid_SUBCONUS_Ind_3km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
- inline_post
- MET_ensemble_verification
- MET_verification
- nco_ensemble
- nco_grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR
- custom_ESGgrid
- custom_GFDLgrid
- custom_GFDLgrid__GFDLgrid_USE_GFDLgrid_RES_IN_FILENAMES_eq_FALSE
- custom_GFDLgrid__GFDLgrid_USE_GFDLgrid_RES_IN_FILENAMES_eq_TRUE
- pregen_grid_orog_sfc_climo
- specify_DOT_OR_USCORE
- specify_DT_ATMOS_LAYOUT_XY_BLOCKSIZE
- specify_RESTART_INTERVAL
- specify_template_filenames

**HERA**
0 failures 

**ORION**
1 failure
```
nco_grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR (run_fcst segfault)
```

**JET**
only 5 failures which is a lot of improvement since last time I tried.
```
MET_ensemble_verification (run_gridstatvx_)
MET_verification (run_gridstatvx)
grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_HRRR (make_lbcs)
nco_ensemble (run_fcst)
pregen_grid_orog_sfc_climo (run_fcst)
```

## DEPENDENCIES:
None

## DOCUMENTATION:
None

## ISSUE (optional): 
None

## CONTRIBUTORS (optional): 
@christinaholtNOAA @venitahagerty


JeffBeck-NOAA:
WE2E using the config.community.sh succeeded on Hera.

scripts/exregional_make_grid.sh
```diff
@@ -589,9 +587,9 @@ halo failed."
 #
 #-----------------------------------------------------------------------
 #
-link_fix \
-  verbose="$VERBOSE" \
-  file_group="grid" || \
+python3 -S $USHDIR/link_fix.py \
+  --path-to-defns $EXPTDIR/var_defns.sh \
+  --file-group "grid" || \
```
christinaholtNOAA, line 592: Would you mind leaving a quick comment on why the -S flag is needed for link fix?

scripts/exregional_run_fcst.sh
```diff
@@ -446,7 +436,9 @@ fi
 
 if [ "${DO_ENSEMBLE}" = TRUE ] && ([ "${DO_SPP}" = TRUE ] || [ "${DO_SPPT}" = TRUE ] || [ "${DO_SHUM}" = TRUE ] \
    [ "${DO_SKEB}" = TRUE ] || [ "${DO_LSM_SPP}" =  TRUE ]); then
-  set_FV3nml_ens_stoch_seeds cdate="$cdate" || print_err_msg_exit "\
+  python3 $USHDIR/set_FV3nml_ens_stoch_seeds \
```
christinaholtNOAA: This ush script needs a .py suffix.

ush/create_model_configure_file.py
```diff
@@ -151,7 +154,7 @@ def create_model_configure_file(cdate,run_dir,sub_hourly_post,dt_subhourly_post_
     # to, or smaller than one hour.
     #
     if sub_hourly_post:
-        nsout=dt_subhourly_post_mnts*60 / dt_atmos
+        nsout=(dt_subhourly_post_mnts*60) // dt_atmos
```
christinaholtNOAA, line 157: Just a quick question...we want this to be integer division?

ush/create_model_configure_file.py
```diff
@@ -196,6 +199,55 @@ def create_model_configure_file(cdate,run_dir,sub_hourly_post,dt_subhourly_post_
...
+
+    parser.add_argument('-r', '--run-dir',
+                        dest='run_dir',
+                        required=True,
+                        help='Run directory.')
+
+    parser.add_argument('-c', '--cdate',
+                        dest='cdate',
+                        required=True,
+                        help='Date string.')
```
christinaholtNOAA: I would be helpful to have the expected format of the date string in the help comment.

ush/create_model_configure_file.py
```diff
@@ -196,6 +199,55 @@ def create_model_configure_file(cdate,run_dir,sub_hourly_post,dt_subhourly_post_
...
+
+    parser.add_argument('-s', '--sub-hourly-post',
+                        dest='sub_hourly_post',
+                        required=True,
+                        help='Sub hourly post.')
```
christinaholtNOAA: What type of argument is required here?

ush/launch_FV3LAM_wflow.sh
```diff
@@ -393,35 +392,14 @@ script for this experiment:
...
-
-    if [ "$MACHINE" = "WCOSS_DELL_P3" ]; then
-      echo "${crontab_contents}" > "/u/$USER/cron/mycrontab"
+    if [ "${called_from_cron}" = "TRUE" ]; then
+       MACHINE=$MACHINE CRONTAB_LINE=$CRONTAB_LINE \
+           python3 $USHDIR/get_crontab_contents.py --delete --called-from-cron
     else
-      echo "${crontab_contents}" | ${crontab_cmd} 
+       MACHINE=$MACHINE CRONTAB_LINE=$CRONTAB_LINE \
+           python3 $USHDIR/get_crontab_contents.py --delete
```
christinaholtNOAA, line 402: This is nice.

ush/link_fix.py
```diff
@@ -320,7 +322,7 @@ def link_fix(verbose, file_group):
            not GFDLgrid_USE_GFDLgrid_RES_IN_FILENAMES:
           target=f"{cres}{DOT_OR_USCORE}grid.tile{TILE_RGNL}.halo{NH4}.nc"
           symlink=f"C{GFDLgrid_RES}{DOT_OR_USCORE}grid.tile{TILE_RGNL}.nc"
-          create_symlink_to_file(target,symlink,relative)
+          create_symlink_to_file(target,symlink,True)
```
christinaholtNOAA: Does relative always equal True? Is it platform dependent?

ush/predef_grid_params.yaml
```diff
@@ -187,41 +187,6 @@
...
-  WRTCMP_stdlat2: 38.5
-  WRTCMP_nx: 393
-  WRTCMP_ny: 225
-  WRTCMP_lon_lwr_left: -121.70231097
-  WRTCMP_lat_lwr_left: 22.57417972
-  WRTCMP_dx: 13000.0
-  WRTCMP_dy: 13000.0
-#
-#-----------------------------------------------------------------------
-#
```
christinaholtNOAA, line 224: Did we get rid of this domain in another PR? Sorry if I missed that.

ush/python_utils/__init__.py
```diff
@@ -1,22 +1,25 @@
...
+  from .run_command import run_command
+  from .get_charvar_from_netcdf import get_charvar_from_netcdf
+  from .xml_parser import load_xml_file, has_tag_with_value
+  from .config_parser import load_shell_config, cfg_to_shell_str, \
+                             load_json_config, cfg_to_json_str, \
+                             load_ini_config, cfg_to_ini_str, \
+                             get_ini_value, load_config_file, \
+                             load_yaml_config, cfg_to_yaml_str
+except:
+  pass
```
christinaholtNOAA, line 25: This seems a bit dangerous. Why wouldn't we have this list of expected modules?

ush/python_utils/config_parser.py
```diff
@@ -14,7 +14,10 @@
 """
 
 import argparse
-import yaml
+try:
+    import yaml
+except:
+    pass
```
christinaholtNOAA, line 20: I'm also concerned here. If we don't have yaml, how will we be able to use this script at all?

ush/python_utils/misc.py
```diff
@@ -55,3 +55,21 @@ def find_pattern_in_file(pattern, file_name):
                 return match.groups()
     return None
 
+def flatten_dict(dictionary,keys=None):
+    """ Faltten a recursive dictionary (e.g.yaml/json) to be one level deep
```
christinaholtNOAA, line 59: Faltten -> Flatten

ush/set_gridparams_ESGgrid.py
```diff
@@ -93,18 +96,18 @@ def test_set_gridparams_ESGgrid(self):
...
+          0.0134894006,
+          -1760,
+          -1050)
         )
 
     def setUp(self):
-        set_env_var('DEBUG',True)
-        set_env_var('VERBOSE',True)
+        set_env_var('RADIUS_EARTH',6371200.0)
+        set_env_var('DEGS_PER_RADIAN',57.2957795131)
```
christinaholtNOAA, line 112: Do we not want to use these values from `constants`?

ush/setup.py
```diff
@@ -119,7 +125,7 @@ def setup():
     if DEBUG and not VERBOSE:
         print_info_msg('''
             Resetting VERBOSE to \"TRUE\" because DEBUG has been set to \"TRUE\"...''')
-        VERBOSE=False
+        VERBOSE=True
```
christinaholtNOAA, line 128: Would `VERBOSE = DEBUG or VERBOSE` work here?

ush/setup.py
```diff
@@ -849,6 +852,20 @@ def setup():
...
+    # Then change it to lowercase.  Finally, ensure that it does not end up 
+    # getting set to an empty string.
+    #
+    #-----------------------------------------------------------------------
+    #
+    POST_OUTPUT_DOMAIN_NAME = POST_OUTPUT_DOMAIN_NAME or PREDEF_GRID_NAME
+    POST_OUTPUT_DOMAIN_NAME = lowercase(POST_OUTPUT_DOMAIN_NAME)
+
     if POST_OUTPUT_DOMAIN_NAME is None:
       if PREDEF_GRID_NAME is None:
```
christinaholtNOAA: Should this logic be changed?  Line 867 will error if either of these are `None`, and it doesn't check for empty strings.

ush/setup.py
```diff
@@ -925,9 +938,11 @@ def setup():
     if DATA_TABLE_TMPL_FN is None:
        DATA_TABLE_TMPL_FN = DATA_TABLE_FN
     if DIAG_TABLE_TMPL_FN is None:
-       DIAG_TABLE_TMPL_FN = f"{DIAG_TABLE_FN}{dot_ccpp_phys_suite_or_null}"
+       DIAG_TABLE_TMPL_FN = DIAG_TABLE_FN
+    DIAG_TABLE_TMPL_FN = f"{DIAG_TABLE_TMPL_FN}{dot_ccpp_phys_suite_or_null}"
```
christinaholtNOAA: This whole section might be cleaner/faster/less repetitive with logic like: 
```
DIAG_TABLE_TMPL_FN = f"{DIAG_TABLE_TMPL_FN or DIAG_TABLE_FN}{dot_ccpp_phys_suite_or_null}"
```

christinaholtNOAA (commented):
@danielabdi-noaa Thanks for your huge, ongoing effort on this! 

I left a few comments/questions below.

scripts/exregional_make_grid.sh
danielabdi-noaa: This was a hack for a logging error I was encountering when python does `import site` by default.
This happens when a task does not activate `regional_workflow` environment, and we fall back to using the system python. On hera though the fall back is the python in the minconda module loaded by `build_hera_intel`, which I have no idea why it is there in the first place.
```
module use /scratch1/NCEPDEV/nems/emc.nemspara/soft/modulefiles
module load miniconda3/3.7.3
```
I suspect there is some issues with this particular installation of conda/python. Using `minoconda/4.5.12` or completely removing those two lines from `build_hera_intel`, thus using the default system python, the problem goes away. That means I don't have to skip `import site` with `python -S`.
I think the miniconda3 version for `build_hera_intel` is now updated to use 4.5.12 so we probably don't need the `-S` now.

scripts/exregional_run_fcst.sh
danielabdi-noaa: Thanks for catching that. I have no idea why this part of the code was not triggered among the 56 test cases I run.

ush/python_utils/__init__.py
danielabdi-noaa: This seems to be the only way to get around with calling any of the python utility functions when `regional_workflow` environment is not activated and `pyyaml` is absent. Note also that yaml is imported very last so that other modules are imported properly before the exception happens. Otherwise, we will not be able to use python versions of `link_fix`, `create_model/diag tables`, use python version to delete crontab line, and a lot of stuff we need for the `yaml config` PR etc. For example, `var_defns` is loaded as a `shell or ini` file, and it is needed by all tasks including those that do not activate `regional_workflow`

ush/python_utils/config_parser.py
danielabdi-noaa: As it turns out, we can use all other functionality of the module, even when yaml is absent. See previous comment.

ush/set_gridparams_ESGgrid.py
danielabdi-noaa: A recent commit has changed the way constants are handled, now they are part of `var_defns` file.
Also, I decided to not have a `constants.py`, which I had before, since they are now available via `var_defns`.

ush/link_fix.py
danielabdi-noaa: Note that `link_fix.py` is consistent with `link_fix.sh`. I  had a bug in `link_fix.py` using `relative` when I should have passed `True` just like `link_fix.sh`

ush/create_model_configure_file.py
danielabdi-noaa: No idea whether it should be or not, but bash does integer divisions by default, so the python version is consistent with
```
nsout=$(( dt_subhourly_post_mnts*60 / dt_atmos ))
```
It was actually not working when I had float division for some test case IIRC.

ush/predef_grid_params.yaml
danielabdi-noaa: I had duplicate RRFS_CONUS_13km in the yaml file, and it was causing a problem without yaml complaining.

ush/setup.py
danielabdi-noaa: I think empty strings are ok (python doesn't complain `''.lower()`), but if both are `None` it is a problem. This part of the code is being worked on at the moment, and I am sure I had a different logic before but changed it again to be consistent with the new `setup.sh`. I will fix it once a pending PR related to this is merged.

scripts/exregional_run_fcst.sh
danielabdi-noaa: Fixed

ush/create_model_configure_file.py
danielabdi-noaa: I have added additional comments to the help string in this particular file. The file does not contain description of the argumetns so it is mostly my best guess.

ush/setup.py
danielabdi-noaa: Done

ush/setup.py
danielabdi-noaa: Yes, but I would still need to test whether to print the info message or not, so it would be redundunt.

ush/python_utils/misc.py
danielabdi-noaa: I will fix this in the YAML PR since it is moved to another file.
**Edit** It was actually already fixed there.

ush/launch_FV3LAM_wflow.sh
danielabdi-noaa: This was the first issue I encountered with passing ENV variables to python subshells. In cases where a lot of ENV variables are used in the python script, `var_defns` has to be passed and sourced in the python script, as you probably saw in many task scripts.

scripts/exregional_make_grid.sh
danielabdi-noaa: I have removed the "-S" option now after testing it doesn't have any effect on Hera anymore.
Hopefully it is not needed on other systems.

JeffBeck-NOAA (approved):

christinaholtNOAA (approved):
LGTM!

PR #766, opened 2022-05-12
update IC/LBC file paths to match new IC/LBC file structure

EdwardSnyder-NOAA:
## DESCRIPTION OF CHANGES: 
Updated the IC/LBC file paths, model start time, and forecast length in the config.community.sh for the "out-of-the-box" case. The IC/LBC file paths were changed to match the recent standardize static data locations in PR #744. The forecast length was reduced to increase the likelihood of a successful model run, and the start time was changed to capture more severe weather.  

## TESTS CONDUCTED: 
Built the SRW App using the develop branch on Orion and ran using the proposed changes in the config.community.sh file. The test ran successfully. 

## DEPENDENCIES:
No dependencies.

## DOCUMENTATION:
The variables in the config.community.sh file are listed and explained here: https://ufs-srweather-app.readthedocs.io/en/develop/BuildRunSRW.html#user-specific-configuration-config-sh. Since some of these variables are changing, minor edits are needed to the documentation.


ush/config.community.sh
```diff
@@ -15,12 +15,12 @@ DO_ENSEMBLE="FALSE"
...
 
 CCPP_PHYS_SUITE="FV3_GFS_v16"
-FCST_LEN_HRS="48"
+FCST_LEN_HRS="12"
 LBC_SPEC_INTVL_HRS="6"
 
 DATE_FIRST_CYCL="20190615"
 DATE_LAST_CYCL="20190615"
-CYCL_HRS=( "00" )
+CYCL_HRS=( "18" )
```
chan-hoo, line 23: @EdwardSnyder-NOAA, is there any reason why you change this cycle hour? In the staged data, `2019061500` has much more data in it while `2019061518` only has two files.

EdwardSnyder-NOAA: @chan-hoo - We changed the cycle hour because we reduced the forecast runtime from 48 hours to 12 hours, and we wanted to ensure there was enough interesting weather during the model run. The reasons why we reduced the forecast runtime was to ensure users were able to run the forecast without using too many resources, reduce the wait time for the model to complete, and to increase the likelihood for a successful model run. Since this is the "out-of-the-box" case, we wanted to ensure the model ran quickly and to provide a great user experience, which in our mind was to reduce the forecast runtime.

chan-hoo: @EdwardSnyder-NOAA, thanks for the explanation. It makes sense.

chan-hoo (approved):

ush/config.community.sh
```diff
@@ -59,8 +59,7 @@ RUN_TASK_VX_ENSPOINT="FALSE"
...
 #
-#EXTRN_MDL_SOURCE_BASEDIR_ICS="/scratch2/BMC/det/UFS_SRW_app/v1p0/model_data/FV3GFS"
-#EXTRN_MDL_FILES_ICS=( "gfs.pgrb2.0p25.f000" )
-#EXTRN_MDL_SOURCE_BASEDIR_LBCS="/path/to/model_data/FV3GFS"
-#EXTRN_MDL_FILES_LBCS=( "gfs.pgrb2.0p25.f006" "gfs.pgrb2.0p25.f012" "gfs.pgrb2.0p25.f018" "gfs.pgrb2.0p25.f024" \
-#                       "gfs.pgrb2.0p25.f030" "gfs.pgrb2.0p25.f036" "gfs.pgrb2.0p25.f042" "gfs.pgrb2.0p25.f048" )
+#EXTRN_MDL_SOURCE_BASEDIR_ICS="/scratch2/BMC/det/UFS_SRW_App/develop/input_model_data/FV3GFS/grib2/2019061518"
+#EXTRN_MDL_FILES_ICS=( "gfs.t18z.pgrb2.0p25.f000" )
+#EXTRN_MDL_SOURCE_BASEDIR_LBCS="/scratch2/BMC/det/UFS_SRW_App/develop/input_model_data/FV3GFS/grib2/2019061518"
+#EXTRN_MDL_FILES_LBCS=( "gfs.t18z.pgrb2.0p25.f006" )
```
christinaholtNOAA: Will you not need one more LBC for a 12 hour forecast?

christinaholtNOAA (changes requested):
Just a quick question below.

ush/config.community.sh
EdwardSnyder-NOAA: I think you're correct. I tested this PR and when I look at the experiment's `test_community/2019061518/FV3GFS/for_LBCS/extrn_mdl_var_defns.sh`, it shows the same 6 hour forecast file twice for `EXTRN_MDL_FNS` even though the `EXTRN_MDL_FHRS=( 6 12 )`. What is strange is that the `run_get_lbcs.sh` didn't throw any errors or download the 12 hour forecast, so I had thought one LBC was sufficient. It is not. Good catch! We will convert back to 00Z forecast start time to avoid downloading this one file to all staged data locations.

EdwardSnyder-NOAA: Instead of converting back to 00Z forecast start time, we decided to reduce the forecast length time from 12 hours to 6 hours.

christinaholtNOAA: Okay, the `FCST_LEN_HRS` above is still set to 12, though.

christinaholtNOAA (changes requested):

ush/config.community.sh
EdwardSnyder-NOAA: I did a commit two days ago that switched the `FCST_LEN_HRS` to 6 hours.

gspetro-NOAA (approved):

ush/config.community.sh
gspetro-NOAA: @christinaholtNOAA I see the changes on my end. Did you refresh the page?

ush/config.community.sh
```diff
@@ -15,12 +15,12 @@ DO_ENSEMBLE="FALSE"
 NUM_ENS_MEMBERS="2"
 
 CCPP_PHYS_SUITE="FV3_GFS_v16"
-FCST_LEN_HRS="48"
+FCST_LEN_HRS="6"
```
gsketefian: @JeffBeck-NOAA Do you remember why we had decided to set this to 48 hours originally?

gsketefian (changes requested):
Asking @JeffBeck-NOAA a question before reducing from 48 hr to 6 hr.

ush/config.community.sh
JeffBeck-NOAA: The GST was chosen as 20190615 with a 00Z initialization and 48-hr forecast length to cover two full diurnal cycles, including two days with significant severe storm reports in various parts of the country.  There is an increased demand by NWS and NSSL for CAM systems to be run into day two (24-48 hrs) time frames for forecast guidance relative to hazardous weather.  The preliminary configuration of the RRFS (the main operational instantiation of the SRW App) will be run out to 60 hrs, four times a day.

Since the GST in the config.community.sh script uses a 25-km domain, I would advocate to keep the 00Z initialization with a 48-hr forecast.  That's very computationally inexpensive.  It also provides the general user with two days of forecasts to plot/compare.

gsketefian: @EdwardSnyder-NOAA Have you been getting failures later (after 6 hours) into the forecast?  Just trying to understand the comment in the PR:

> The forecast length was reduced to increase the likelihood of a successful model run.

gsketefian: Oh, I just saw your explanation to @chan-hoo.

ush/config.community.sh
JeffBeck-NOAA: In the config.community.sh file, the 20190615 GST starting at 00Z is a 25-km configuration, so the computational resources are very small for a 48-hr forecast (smaller than a 6-hr 3-km forecast with the same CONUS domain).  The test is very fast, and provides a more representative use case of the App (aside from resolution) covering two days of severe storm activity (the RRFS will initially run out 60 hours).  It's also a fundamental WE2E test case, so should never fail in develop.  I would advocate for keeping the original GST configuration.

JeffBeck-NOAA:
> Asking @JeffBeck-NOAA a question before reducing from 48 hr to 6 hr.

Please see my comments [here](https://github.com/ufs-community/regional_workflow/pull/766#discussion_r877596267) and [here](https://github.com/ufs-community/regional_workflow/pull/766#discussion_r877606594).

ush/config.community.sh
gspetro-NOAA: @JeffBeck-NOAA @gsketefian I can understand wanting the "out-of-the-box" case to be longer than 6 hours (I had proposed 12 or 24), but I do think it should be reduced from 48 hours. It takes about 2 hours to run through the whole 48-hour experiment, and if it fails 90 minutes in, that's a lot of time for a beginner to spend only to have it not work. I have had cases where I hit the wall time for the run_fcst task using the default 48-hour forecast, so the forecast task went dead, and I had to restart it. Users should be able to run the "out-of-the-box" case in config.community.sh without making any adjustments (e.g., to wall time) so that they are starting with a config file that they _know_ works. We could add a line to the config.community.sh file to increase the wall time for run_fcst, but I don't think we should. One hour (plus the time for the pre-processing tasks) is already a LONG time for a beginner to wait before figuring out they've done something wrong, even if it's not a long time in the vast scheme of weather forecasting. From an educational perspective, frequent feedback is preferable, and a shorter forecast would provide feedback more quickly. For the SRW short course, we were using a 12-hour forecast, which is why we were leaning toward using that in the config file, too. That said, I think one of the data files is missing on one or more systems. @EdwardSnyder-NOAA I know it's a pain, but I think we should add that file and maybe up the forecast to at least 12 hours.

JeffBeck-NOAA: @gspetro-NOAA, I was curious to know where the `WTIME_RUN_FCST="01:00:00"` setting in the config.community.sh file came from.  It looks like it was added [during the inclusion](https://github.com/ufs-community/regional_workflow/commit/68122ac05ec78a8dc878b370493344fb8f5e04b7#diff-10e1b947da3bcd21d54396c2f6f6dcd770ea537177fd8fe955d2205ae2fa378c) of the 48-hr GST in 2020.  I'm actually wondering if that limit was not originally used for the 6-hr forecast seen in the old config.community.sh file in that commit and it was never updated for the 48-hr GST.  Although, I could have sworn we were able to run that configuration in under an hour for the first release.  It's also possible that the model run time has changed that much in two years due to a number of major dycore and physics changes that have gone into the repos.  I can see on Hera that hourly output for the GST at 25-km is taking three minutes on average, which equates to ~144 minutes of run time.  Obviously, we want a wall clock request in the config.community.sh file that works!  I can also certainly understand the desire for new users to have quick turnaround for results on their first try.  All that said, I'm happy with a compromise using @EdwardSnyder-NOAA's proposed 18Z init, but with a 12-hr forecast.  Will that work for everyone?  At least with 12 hours, we'll get the full severe storm outbreak on day 1.  We should update the wallclock request to ~2 hours, just to be safe.

JeffBeck-NOAA (approved):

ush/config.community.sh
JeffBeck-NOAA, line 18:
```suggestion
FCST_LEN_HRS="12"
```

ush/config.community.sh
christinaholtNOAA: Weird. I had navigated here fresh yday when I took a new look. Maybe I clicked "outdated" and didn't realize? Either way, looks good on my end now. Sorry for the delay.

christinaholtNOAA (approved):

JeffBeck-NOAA (approved):

PR #778, opened 2022-05-23
[develop]: Use yaml format for specifying experiment parameters

danielabdi-noaa:
## DESCRIPTION OF CHANGES: 
Use structured yaml configs for specifying experiment parameters. This helps in better management of the recent proliferation of tasks and their associated parameters. However, one can still use shell scripts for configuring experiments by changing config script extension to `sh` in `config_defaults.yaml`
 ```yaml
 EXPT_CONFIG_FN: "config.sh"
 ```

Furthermore, a utility for converting shell config scripts to yaml config is provided. 

```bash
$ ./config_utils.py --help
usage: config_utils.py [-h] --cfg CFG [--output-type OUT_TYPE] [--flatten] [--template-cfg TEMPLATE] [--keys KEYS [KEYS ...]] [--validate-cfg VALIDATE]

Utility for managing different config formats.

optional arguments:
  -h, --help            show this help message and exit
  --cfg CFG, -c CFG     Config file to parse
  --output-type OUT_TYPE, -o OUT_TYPE
                        Output format: can be any of ["shell", "yaml", "ini", "json", "xml"]
  --flatten, -f         Flatten resulting dictionary
  --template-cfg TEMPLATE, -t TEMPLATE
                        Template config file used to structure a given config file
  --keys KEYS [KEYS ...], -k KEYS [KEYS ...]
                        Include only these keys of dictionary for processing. Keys can be python regex expression.
  --validate-cfg VALIDATE, -v VALIDATE
                        Validation config file used to validate a given config file

```

To convert a shell config script to yaml config script, for example `config.community.sh`
```yaml
$ ./config_utils.py -c $PWD/config.community.sh -t $PWD/config_defaults.yaml -o yaml
user:
  RUN_ENVIR: community
  MACHINE: hera
  ACCOUNT: an_account
platform:
  MODEL: FV3_GFS_v16_CONUS_25km
  MET_INSTALL_DIR: path/to/MET
  METPLUS_PATH: path/to/METPlus
  CCPA_OBS_DIR: /path/to/processed/CCPA/data
  MRMS_OBS_DIR: /path/to/processed/MRMS/data
  NDAS_OBS_DIR: /path/to/processed/NDAS/data
workflow:
  EXPT_SUBDIR: test_community
  CCPP_PHYS_SUITE: FV3_GFS_v16
  DATE_FIRST_CYCL: '20190615'
  DATE_LAST_CYCL: '20190615'
  CYCL_HRS:
    - 18
  FCST_LEN_HRS: 12
  PREEXISTING_DIR_METHOD: rename
  VERBOSE: true
  COMPILER: intel
workflow_switches:
  RUN_TASK_MAKE_GRID: true
  RUN_TASK_MAKE_OROG: true
  RUN_TASK_MAKE_SFC_CLIMO: true
  RUN_TASK_GET_OBS_CCPA: false
  RUN_TASK_GET_OBS_MRMS: false
  RUN_TASK_GET_OBS_NDAS: false
  RUN_TASK_VX_GRIDSTAT: false
  RUN_TASK_VX_POINTSTAT: false
  RUN_TASK_VX_ENSGRID: false
  RUN_TASK_VX_ENSPOINT: false
task_run_fcst:
  WTIME_RUN_FCST: 02:00:00
  QUILTING: true
  PREDEF_GRID_NAME: RRFS_CONUS_25km
task_get_extrn_ics:
  EXTRN_MDL_NAME_ICS: FV3GFS
  FV3GFS_FILE_FMT_ICS: grib2
task_get_extrn_lbcs:
  EXTRN_MDL_NAME_LBCS: FV3GFS
  LBC_SPEC_INTVL_HRS: 6
  FV3GFS_FILE_FMT_LBCS: grib2
global:
  DO_ENSEMBLE: false
  NUM_ENS_MEMBERS: 2
```
To check validity of existing yaml or shell config
```bash
$ ./config_utils.py -c $PWD/config.community.yaml -v $PWD/config_defaults.yaml
SUCCESS
```

## TESTS CONDUCTED: 
I have run tests successfully on **HERA**, **JET**, **GAEA** and **CHEYNNE gnu+intel**

**GAEA** and **CHEYNNE** tested through Jenkins shown [in this page](https://jenkins-epic.woc.noaa.gov/blue/organizations/jenkins/ufs-srweather-app%2Fpipeline/detail/PR-309/6/pipeline/145/).

Out of 56 tests

- community_ensemble_008mems
- community_ensemble_2mems
- deactivate_tasks
- grid_CONUS_25km_GFDLgrid_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
- grid_CONUS_3km_GFDLgrid_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
- grid_RRFS_AK_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
- grid_RRFS_AK_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
- grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
- grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_2017_gfdlmp
- grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_2017_gfdlmp
- grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v15p2
- grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_HRRR
- grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_RRFS_v1beta
- grid_RRFS_CONUScompact_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
- grid_RRFS_CONUScompact_13km_ics_HRRR_lbcs_RAP_suite_HRRR
- grid_RRFS_CONUScompact_13km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha
- grid_RRFS_CONUScompact_13km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
- grid_RRFS_CONUScompact_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
- grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_HRRR
- grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_RRFS_v1beta
- grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR
- grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha
- grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
- grid_RRFS_CONUScompact_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
- grid_RRFS_CONUScompact_3km_ics_HRRR_lbcs_RAP_suite_GFS_v15p2
- grid_RRFS_CONUScompact_3km_ics_HRRR_lbcs_RAP_suite_HRRR
- grid_RRFS_CONUScompact_3km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha
- grid_RRFS_CONUScompact_3km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
- grid_RRFS_NA_13km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta
- grid_RRFS_SUBCONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
- grid_RRFS_SUBCONUS_3km_ics_HRRR_lbcs_RAP_suite_GFS_v15p2
- grid_SUBCONUS_Ind_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
- grid_SUBCONUS_Ind_3km_ics_HRRR_lbcs_RAP_suite_HRRR
- grid_SUBCONUS_Ind_3km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
- inline_post
- MET_ensemble_verification
- MET_verification
- nco_ensemble
- nco_grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR
- custom_ESGgrid
- custom_GFDLgrid
- custom_GFDLgrid__GFDLgrid_USE_GFDLgrid_RES_IN_FILENAMES_eq_FALSE
- custom_GFDLgrid__GFDLgrid_USE_GFDLgrid_RES_IN_FILENAMES_eq_TRUE
- pregen_grid_orog_sfc_climo
- specify_DOT_OR_USCORE
- specify_DT_ATMOS_LAYOUT_XY_BLOCKSIZE
- specify_RESTART_INTERVAL
- specify_template_filenames

**HERA**
0 failures 

**JET**
5 failures 

## DEPENDENCIES:
None

## DOCUMENTATION:
None

## ISSUE: 
To be opened later

## CONTRIBUTORS: 
@christinaholtNOAA


tests/WE2E/run_WE2E_tests.sh
```diff
@@ -1276,12 +1273,25 @@ exist or is not a directory:
...
 #
+# Set the full path to the configuration file that the experiment 
+# generation script reads in.  Then write the contents of expt_config_str 
+# to that file.
+#
+#-----------------------------------------------------------------------
+#
+  expt_config_fp="$ushdir/${EXPT_CONFIG_FN}"
+  ext="${EXPT_CONFIG_FN##*.}"
+  config_to_str "${ext}" "${temp_file}" -t "$ushdir/config_defaults.yaml" >"${expt_config_fp}"
```
christinaholtNOAA, line 1284: Is this a bash utility? Will that change with a transition to python?

tests/WE2E/test_configs/grids_extrn_mdls_suites_community/config.grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_HRRR.sh
```diff
@@ -3,15 +3,15 @@
 # ------------------------
 #
 # This test is to ensure that the workflow running in community mode 
-# completes successfully on the RRFS_CONUS_13km grid using the GFS_v16
+# completes successfully on the RRFS_CONUS_13km grid using the HRRR
```
christinaholtNOAA: This one didn't make it to yaml.

tests/WE2E/test_configs/grids_extrn_mdls_suites_community/config.grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta.sh
```diff
@@ -3,15 +3,15 @@
 # ------------------------
 #
 # This test is to ensure that the workflow running in community mode 
-# completes successfully on the RRFS_CONUS_13km grid using the GFS_v15p2 
+# completes successfully on the RRFS_CONUS_13km grid using the RRFS_v1beta
```
christinaholtNOAA: This one either.

tests/WE2E/test_configs/grids_extrn_mdls_suites_community/config.grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_RRFS_v1beta.yaml
```diff
@@ -0,0 +1,24 @@
...
+  DATE_LAST_CYCL: '20210615'
+  CYCL_HRS:
+  - '00'
+  FCST_LEN_HRS: '6'
+  PREEXISTING_DIR_METHOD: rename
+task_run_fcst:
+  PREDEF_GRID_NAME: RRFS_CONUS_25km
+task_get_extrn_ics:
+  EXTRN_MDL_NAME_ICS: NAM
+  USE_USER_STAGED_EXTRN_FILES: 'TRUE'
```
christinaholtNOAA: In general, should we use boolean `True` in these files instead of string `"TRUE"`? What about strings vs integers?

tests/WE2E/test_configs/grids_extrn_mdls_suites_community/config.grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_HRRR.sh
```diff
@@ -3,15 +3,15 @@
 # ------------------------
 #
 # This test is to ensure that the workflow running in community mode 
-# completes successfully on the RRFS_CONUS_3km grid using the GFS_v16
+# completes successfully on the RRFS_CONUS_3km grid using the HRRR
```
christinaholtNOAA: Should this one be yaml?

tests/WE2E/test_configs/grids_extrn_mdls_suites_community/config.grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - '00'
+  FCST_LEN_HRS: '24'
+  PREEXISTING_DIR_METHOD: rename
+task_run_fcst:
+  PREDEF_GRID_NAME: RRFS_CONUScompact_25km
+task_get_extrn_ics:
+  EXTRN_MDL_NAME_ICS: HRRR
+  USE_USER_STAGED_EXTRN_FILES: 'TRUE'
+  EXTRN_MDL_FILES_ICS:
+  - '{yy}{jjj}{hh}00{fcst_hr:02d}00'
```
christinaholtNOAA: This might need to be indented with 2 more spaces.

tests/WE2E/run_WE2E_tests.sh
danielabdi-noaa: Look into `bash_utils/source_config.sh` file I added for handling different config formats. The file is just a wrapper for whatever functionality is present in main of `python_utils/config_parser.py`.

tests/WE2E/test_configs/grids_extrn_mdls_suites_community/config.grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_RRFS_v1beta.yaml
danielabdi-noaa: This is a result of automatic conversion but we can replace the 'TRUE's by actual boolean 'True' just fine. For conversion between different formats, I disable interpretation of values (and just return string), to be safe. One problem I was having was when for example a date like '20190101' is laded it becomes a datetime object, later when it is converted to yaml via yaml.dumps() it uses `str()` to convert it to string instead of my custom `date_to_str()`, so it will have '2019-01-01 00:00:00' instead of '20190101'. A custom datetime class can probably do the job but I wanted to keep it simple. For the static test_configs though, I could change them to use `True/False`

tests/WE2E/test_configs/grids_extrn_mdls_suites_community/config.grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha.yaml
danielabdi-noaa: I am using yaml.dumps() to write the file, so will look for config options that will do that.

tests/WE2E/test_configs/grids_extrn_mdls_suites_community/config.grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_HRRR.sh
danielabdi-noaa: It must have been a merge issue. There were 5 new ones added recently and this is one of them.

tests/WE2E/test_configs/grids_extrn_mdls_suites_community/config.grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_RRFS_v1beta.yaml
danielabdi-noaa: I am actually able to do the date conversion properly now with something similar to `yaml.add_constructor`.
```python
def date_representer(dumper, d):
    return dumper.represent_scalar(u'!datetime', date_to_str(d))
yaml.add_representer(datetime, date_representer)
```
It produces the familar format you expect for yaml
```yaml
workflow:
  USE_CRON_TO_RELAUNCH: true
  CRON_RELAUNCH_INTVL_MNTS: 2
  EXPT_BASEDIR: /scratch1/BMC/zrtrr/Daniel.Abdi/rrfs/benchmark-run
  EXPT_SUBDIR: deactivate_tasks
  DATE_FIRST_CYCL: !datetime '20190701'
  DATE_LAST_CYCL: !datetime '20190701'
  CYCL_HRS:
  - 0
  FCST_LEN_HRS: 6
  PREEXISTING_DIR_METHOD: rename
  VERBOSE: false
  COMPILER: intel
workflow_switches:
  RUN_TASK_GET_EXTRN_ICS: false
  RUN_TASK_GET_EXTRN_LBCS: false
  RUN_TASK_MAKE_ICS: false
  RUN_TASK_MAKE_LBCS: false
  RUN_TASK_RUN_FCST: false
  RUN_TASK_RUN_POST: false
```
However, I can't convert this file to json for example. It complains
```
TypeError: Object of type datetime is not JSON serializable
```
Also, the `CYCL_HRS` is now truncated from '00' to 0 (because it is converted to an integer and then back to string), which I am sure has caused me problems before!
So I feel very uneasy using the non-string version, so I will not change it for now. 
Note that, whichever way you write the yaml file, it will work because when the variables are imported to a shell script, actual type conversion is carried out, so there is no problem whether i use `'TRUE'`, `true`, `'yes'` etc.

danielabdi-noaa: I think i've now solved this more or less. Only datetime objects are quoted now, others like boolean and ints are not.
I will need to run the regression test fully though to make sure everything is working.

tests/WE2E/test_configs/grids_extrn_mdls_suites_community/config.grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha.yaml
danielabdi-noaa: I've fixed this with a custom yaml dumper. Although no indentation is required for lists, adding indentation does improve readability. It is only `pyyaml` that does not indent lists, but other packages do.

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220714203511/ufs-srweather-app
Build was Successful
Rocoto jobs started
ci-hera-intel-WE

venitahagerty:
Machine: jet
Compiler: intel
Job: WE
Repo location: /lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/944706962/20220714203518/ufs-srweather-app
Build was Successful
Rocoto jobs started
ci-jet-intel-WE

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220715150516/ufs-srweather-app
Build was Successful
Rocoto jobs started
ci-hera-intel-WE

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220722173516/ufs-srweather-app
Build was Successful
Rocoto jobs started
ci-hera-intel-WE

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220722193516/ufs-srweather-app
Build was Successful
Rocoto jobs started
ci-hera-intel-WE

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220722205017/ufs-srweather-app
Build was Successful
Rocoto jobs started
ci-hera-intel-WE

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220722215010/ufs-srweather-app
Build was Successful
Rocoto jobs started
ci-hera-intel-WE

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220727160524/ufs-srweather-app
Build was Successful
Rocoto jobs started
ci-hera-intel-WE

.github/workflows/python_unittests.yaml
```diff
@@ -1,5 +1,5 @@
 name: Python unittests
-on: [push, pull_request]
+on: [push, pull_request, workflow_dispatch]
```
christinaholtNOAA, line 2: Does this mean we have a button to run unittests on GitHub runners any time we want from any branch?

ush/calculate_cost.py
```diff
@@ -4,103 +4,128 @@
...
-        lon_of_t6_ctr=GFDLgrid_LON_T6_CTR, \
-        lat_of_t6_ctr=GFDLgrid_LAT_T6_CTR, \
-        res_of_t6g=GFDLgrid_NUM_CELLS, \
-        stretch_factor=GFDLgrid_STRETCH_FAC, \
-        refine_ratio_t6g_to_t7g=GFDLgrid_REFINE_RATIO, \
-        istart_of_t7_on_t6g=GFDLgrid_ISTART_OF_RGNL_DOM_ON_T6G, \
-        iend_of_t7_on_t6g=GFDLgrid_IEND_OF_RGNL_DOM_ON_T6G, \
-        jstart_of_t7_on_t6g=GFDLgrid_JSTART_OF_RGNL_DOM_ON_T6G, \
-        jend_of_t7_on_t6g=GFDLgrid_JEND_OF_RGNL_DOM_ON_T6G)
+        (
```
christinaholtNOAA: Oof! These output lists are a lot to take in here. You know their downstream use much more than I do, but I'm curious if it would make sense to have the `set_gridparams_*` functions return a dict that is passed to `export_vars()` down below? It could help readability and supportability of an extensive tuple return.

ush/calculate_cost.py
```diff
@@ -4,103 +4,128 @@
...
-        set_env_var('LAYOUT_Y',36)
-        set_env_var('BLOCKSIZE',28)
-        set_env_var('QUILTING',False)
-        set_env_var('RADIUS_EARTH',6371200.0)
-        set_env_var('DEGS_PER_RADIAN',57.2957795131)
-
+        set_env_var("DEBUG", False)
+        set_env_var("PREDEF_GRID_NAME", "RRFS_CONUS_3km")
+        set_env_var("DT_ATMOS", 36)
+        set_env_var("LAYOUT_X", 18)
```
christinaholtNOAA, line 105: Remind me why we don't want to use Python interface to set an environment variable? I worry about providing a wrapper to a very common standard library that doesn't really give us much. We will inevitably run into some instances using `set_env_var` and others calling `os.environ` directly with no great way to "police" it.

ush/get_crontab_contents.py
```diff
@@ -30,141 +37,165 @@ def get_crontab_contents(called_from_cron):
...
+        Copying contents of user cron table to backup file:
+          crontab_backup_fp = \"{crontab_backup_fp}\"''',
+        verbose=VERBOSE,
+    )
 
     global called_from_cron
-    try: called_from_cron
-    except: called_from_cron = False
+    try:
+        called_from_cron
```
christinaholtNOAA, line 92: Is this to check to see if the variable is set? Would you mind adding a specific exception to the `except` block to make sure we are only catching that case? If something else goes wrong, we want to make sure we treat that appropriately.

ush/mrms_pull_topofhour.py
```diff
@@ -5,84 +5,165 @@
...
+    # Copy and rename the file for future ease
+    difference = abs(closest_timestamp - valid)
+    if difference.total_seconds() <= 900:
+        filename1 = (
+            MRMS_PRODUCT
+            + level
+            + closest_timestamp.strftime("%Y%m%d-%H%M%S")
+            + ".grib2.gz"
+        )
+        filename2 = MRMS_PRODUCT + level + valid.strftime("%Y%m%d-%H") + "0000.grib2.gz"
```
christinaholtNOAA, line 101: These might be more readable as f-strings.

ush/mrms_pull_topofhour.py
```diff
@@ -5,84 +5,165 @@
...
+                + valid.strftime("%Y%m%d")
+                + "/dcom/us007003/ldmdata/obs/upperair/mrms/conus/"
+                + MRMS_PRODUCT
+                + "/"
+                + filename1
+                + " "
+                + VALID_DIR
+                + "/"
+                + filename2
+            )
```
christinaholtNOAA, line 117: These might be more readable by setting the source and destination files to variables and then an f-string like:
``` 
print(f`{cp} {source} {destination}`)
```

Also, should cp_vrfy be used here?

christinaholtNOAA (commented):
@danielabdi-noaa Thanks for this contribution!

I'm curious if in your testing, you have compared that you get identical output from the yaml-defined test configure files and the bash-defined test configure files? I have not been super thorough here in comparing the original config files to the new ones (read: I skipped over all of them), and it seems that may be the way to get the definitive answer to whether that transition has been accomplished without errors.


I only got about halfway through (mrms_pull_topofhour.py). I will return to finish up asap.

danielabdi-noaa:
> @danielabdi-noaa Thanks for this contribution!
> 
> I'm curious if in your testing, you have compared that you get identical output from the yaml-defined test configure files and the bash-defined test configure files? I have not been super thorough here in comparing the original config files to the new ones (read: I skipped over all of them), and it seems that may be the way to get the definitive answer to whether that transition has been accomplished without errors.
> 
Yes, i have run such test and still have the result here, with four different input formats.
```
/scratch2/BMC/gsd-hpcs/Daniel.Abdi/expt_dirs
```
Those are the 10 tests from the automated test done 4 times each with different config formats. This PR affects only how the "var_defns" file is generated and if that is identical, the rest should be the same too. If I do a diff between them only the `EXPT_CONFIG_FN` entry differs in the "var_defns" file.
```Bash
51c51
< EXPT_CONFIG_FN='config.json'
---
> EXPT_CONFIG_FN='config.ini'
```

> I only got about halfway through (mrms_pull_topofhour.py). I will return to finish up asap.

Sorry I added linting with "black --safe" for all python files, so it may look like I made a change to that file but nothing is changed there. Same goes for `retrieve_data.py` and other python files which this PR should not change.

christinaholtNOAA:
Perfect, @danielabdi-noaa. Thank you! 

I'm going to try to knock out the second half now...

ush/python_utils/__init__.py
```diff
@@ -1,25 +1,62 @@
...
+    cfg_to_xml_str,
+    flatten_dict,
+    structure_dict,
+    check_structure_dict,
+    update_dict,
+    cfg_main,
+    load_config_file,
+)
+
+# yaml may not be available
```
christinaholtNOAA: Is it advisable to proceed here if yaml isn't available?

ush/python_utils/__init__.py
```diff
@@ -1,25 +1,62 @@
...
-  from .config_parser import load_shell_config, cfg_to_shell_str, \
-                             load_json_config, cfg_to_json_str, \
-                             load_ini_config, cfg_to_ini_str, \
-                             get_ini_value, load_config_file, \
-                             load_yaml_config, cfg_to_yaml_str
+    from .config_parser import (
+        load_yaml_config,
+        cfg_to_yaml_str,
+    )
 except:
```
christinaholtNOAA: The standard is to list a specific exception. Does it make sense to have this one only pass on a `ModuleNotFoundError`?

ush/python_utils/config_parser.py
```diff
@@ -89,127 +135,373 @@ def load_shell_config(config_file):
...
+    """
+    flat_dict = {}
+    for k, v in dictionary.items():
+        if not keys or k in keys:
+            if isinstance(v, dict):
+                r = flatten_dict(v)
+                flat_dict.update(r)
+            else:
+                flat_dict[k] = v
+    return flat_dict
```
christinaholtNOAA, line 349: This one was in misc.py. Will the move cause problems for the imports in other scripts? I'm guessing no given all the tests pass. Maybe I missed that magic.

ush/python_utils/environment.py
```diff
@@ -42,26 +44,30 @@ def date_to_str(d, format="%Y%m%d%H%M"):
...
     return v
 
-def str_to_type(s, just_get_me_the_string = False):
-    """ Check if the string contains a float, int, boolean, or just regular string.
+
+def str_to_type(s, return_string=0):
+    """Check if the string contains a float, int, boolean, datetime, or just regular string.
     This will be used to automatically convert environment variables to data types
     that are more convenient to work with. If you don't want this functionality,
-    pass just_get_me_the_string = True
```
christinaholtNOAA, line 49: You didn't like original name?  😏

ush/python_utils/filesys_cmds_vrfy.py
```diff
@@ -14,22 +15,32 @@ def cmd_vrfy(cmd, *args):
     """
 
     for a in args:
-        cmd += ' ' + str(a)
+        cmd += " " + str(a)
```
christinaholtNOAA: Suggesting a more pythonic solution:
```
cmd + " ".join([str(a) for a in args])
```

ush/python_utils/interpol_to_arbit_CRES.py
```diff
@@ -22,22 +23,21 @@ def interpol_to_arbit_CRES(RES, RES_array, prop_array):
...
 
     if RES <= RES_array[i_min]:
         prop = prop_array[i_min]
     elif RES > RES_array[i_max]:
         prop = prop_array[i_max]
     else:
-        for i in range(0,num_valid_RESes-1):
-            if RES > RES_array[i] and RES <= RES_array[i+1]:
+        for i in range(0, num_valid_RESes - 1):
+            if RES_array[i] < RES <= RES_array[i + 1]:
```
christinaholtNOAA, line 34: Nice. I always forget about this notation!

christinaholtNOAA (commented):
Just a few other random comments/questions.

ush/python_utils/__init__.py
danielabdi-noaa: It was actually not necessary to have the exception because none is thrown, so I've removed it no. I did modify `config_parser.py` to catch only the specific exception. We will get a runtime error when we try to use `load_yaml_config` and `cfg_to_yaml_str` but pyyaml is not loaded.

ush/python_utils/__init__.py
danielabdi-noaa: Yes. The other formats ini/json/xml/shell will not work if we stop. For example, there are many instances when the `regional_workflow` conda env is not loaded but we want to load `var_defns.sh` file, or in the case where the user config format is ini/json etc.

ush/python_utils/interpol_to_arbit_CRES.py
danielabdi-noaa: I didn't know about it either, the linter came up with it :)

ush/python_utils/config_parser.py
danielabdi-noaa: I am importing from `python_utils` not from `misc` in scripts that use it so the only place I needed to fix it was `__init_.py`. This really helps to reorganize functions in python_utils without breaking something elsewhere!

ush/mrms_pull_topofhour.py
danielabdi-noaa: I am actually going through this process not for this PR but for the NCO transition where `cp_vrfy` have to be used. There is also `cpfs` to prevent a partially copied file from being used, but I am not sure where those are.

ush/get_crontab_contents.py
danielabdi-noaa: The old `generate_FV3LAM_wflow.sh` had something like
```
  called_from_cron=${called_from_cron:-"FALSE"}
  get_crontab_contents called_from_cron=${called_from_cron} \
                       outvarname_crontab_cmd="crontab_cmd" \
                       outvarname_crontab_contents="crontab_contents"
```
So the python code mimics the first line. The `generate_FV3LAM_wflow.sh` does not get any input arguments so `called_from_cron` can only come from the environment. I asked about this a while back since `called_from_cron` is always FALSE unless you set it to TRUE before calling generate, which never happens since we don't generate the workflow itself from a cron job ...

ush/calculate_cost.py
danielabdi-noaa: The wrapper takes care of list to string conversion in a way that can be used other shell scripts.
```
os.environ[param] = list_to_str(value)
```
Also regular python types such as `bool` are `datetime` need to be converted to a proper formatted string 
that is not just `str(x)`

ush/mrms_pull_topofhour.py
danielabdi-noaa: Let me know if I need to remove the formatting for these files which this PR does not alter.
```
mrms_pull_topofhour.py
UFS_plot_domains.py
retrieve_data.py
python_utils/fv3write_parms_lamber.py
```
I will try to reorg the commits so that the linter is applied last, but not sure if I can get through the git conflicts ...

ush/calculate_cost.py
christinaholtNOAA: Thanks!

ush/calculate_cost.py
danielabdi-noaa: This is an excellent idea! It has simplified the code a lot so will push the changes later.

ush/get_crontab_contents.py
christinaholtNOAA: Hmmm. Yeah, I guess I'm just not familiar with the use case where we are putting just a variable in a try clause to check whether it's been set. Typically we'd be passing this variable and have it set to None, for example. This makes sense if we are doing magic to pull it from the environment.

ush/mrms_pull_topofhour.py
christinaholtNOAA: This can totally address this later. I'd say this PR is big enough that just letting the linter do it's work as-is is fine. I have a couple of additional ideas for other formatting next steps we can talk through off line.

ush/python_utils/filesys_cmds_vrfy.py
danielabdi-noaa: Done

ush/mrms_pull_topofhour.py
christinaholtNOAA: Nice! 

As I mentioned above, we can leave the formatting work for a subsequent PR and adopt the linter version here.

ush/python_utils/__init__.py
christinaholtNOAA: Okay. Still a little tiny bit shady to me without full understanding, but it makes enough sense with the explanation. Would you mind adding a bit of that explanation in the comment? Just the why it's okay that yaml may not be loaded?

ush/python_utils/__init__.py
christinaholtNOAA: Thanks!

ush/python_utils/config_parser.py
christinaholtNOAA: Nice.

ush/python_utils/__init__.py
danielabdi-noaa: Done.

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220817172016/ufs-srweather-app
Build was Successful
Rocoto jobs started
ci-hera-intel-WE

.github/workflows/python_unittests.yaml
danielabdi-noaa: That is the idea. I did check that I can manually re-run a workflow on my fork, but things could be different once it is merged here.

panll:
I did some runs on Hera. The results from shell file and yaml file are not byte-for-byte match. The directory for shell file run:
/scratch2/BMC/det/lpan/review/09012022/test/expt_dirs, and the directory for yaml file run: /scratch2/BMC/det/lpan/review/09012022/test1/expt_dirs.
Is byte-for-byte match required? @danielabdi-noaa 
We may also keep the original shell config files. Thanks!

danielabdi-noaa:
@panll Thanks for testing. Could you please tell me what is the hash of the commit you used for the develop branch  i.e. for the shell workflow runs? I have done 10 tests using this PR but 5 different formats including shell config files and they are bit-for-bit identical.
Here are my results:
```
/scratch2/BMC/gsd-hpcs/Daniel.Abdi/expt_dirs-save
```
There is a `compare.sh` script there you can use like this:
```
./compare.sh grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2/2019070100/ grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2_old_20220715_151808/2019070100/
```
It should output nothing if they match.

In the mean time I will re-run the community test case. Thanks

panll:
I use develop branch. The details can be found in the file /scratch2/BMC/det/lpan/review/09012022/test/ufs-srweather-app/Externals.cfg.

diff test/expt_dirs/test_community/2019061518/dynf000.nc test1/expt_dirs/test_community/2019061518/dynf000.nc
Binary files 2019061518/dynf000.nc and ../../../test1/expt_dirs/test_community/2019061518/dynf000.nc differ

danielabdi-noaa:
@panll The community test case also matches bit for bit whether i use shell or a yaml config or a shell config using this PR.
However, it does not match your yaml or shell result for some reason? Maybe different binaries for usf-weather-model?
I will run on develop branch shortly.

panll:
The weather model code used for test and test1 are same, which can be verified in the file Externals.cfg.

danielabdi-noaa:
@panll I am talking about my result for yaml not matching your result for yaml. I would expect that to match unless there is a difference in ufs-weather-model build or something of that sort.

gsketefian:
@danielabdi-noaa I ran 3 WE2E tests with a version of this PR from Wednesday.  They all passed, but the jobs in my crontab weren't cleared afterwards.  I will rerun them with the latest version, maybe I made a mistake.

tests/WE2E/get_WE2Etest_names_subdirs_descs.sh
```diff
@@ -995,80 +979,11 @@ configuration files of the primary WE2E tests...
...
-# space character".
-#
-# First remove leading whitespace.
-#
-      test_desc="${test_desc#"${test_desc%%[![:space:]]*}"}"
-#
-# Now remove trailing whitespace.
-#
-      test_desc="${test_desc%"${test_desc##*[![:space:]]}"}"
+      config_fn="config.${test_name}.yaml"
```
gsketefian, line 982: I generated a csv file containing test info and after opening it in [google sheets](https://docs.google.com/spreadsheets/d/1_wqJWatfSa8fQDVl59OjNiM4mWVLBoA0zH9850VuPGo/edit#gid=1413052772), I found a couple of issues.  At least in some places in the test descriptions, newlines are dropped and single quotes (apostrophes) are replaced with double quotes.  For example, for the test `custom_GFDLgrid__GFDLgrid_USE_NUM_CELLS_IN_FILENAMES_eq_TRUE`, the description in the yaml file is:
```
metadata:
  description: >-
    This test checks the capability of the workflow to have the user
    specify a new grid (as opposed to one of the predefined ones in the
    workflow) of GFDLgrid type.  Note that this test sets the workflow
    variable
      GFDLgrid_USE_NUM_CELLS_IN_FILENAMES
    to "FALSE"; see the UFS SRW App's User's Guide for a description of
    this variable.
```
but in the csv file, it becomes:
```
This test checks the capability of the workflow to have the user specify a new grid (as opposed to one of the predefined ones in the workflow) of GFDLgrid type.  Note that this test sets the workflow variable   GFDLgrid_USE_NUM_CELLS_IN_FILENAMES to "FALSE"; see the UFS SRW App"s User"s Guide for a description of this variable.
```
So the newline/indentation is ignored and the apostrophes in App's User's Guide are changed to double quotes.  Any chance we can fix that up?

gsketefian: Just randomly checking other descriptions in the google sheet, I see a similar issue for the test `deactivate_tasks`: extra spaces added and newlines ignored.  Is there a way to have yaml treat the description verbatim?

danielabdi-noaa:
@panll I run the community test case with the develop branch, the yaml branch with both yaml and shell config formats and they are all identical after you "fix" develop branch with ush/set_gridparams_ESGgrid.py
```Python
diff --git a/ush/set_gridparams_ESGgrid.py b/ush/set_gridparams_ESGgrid.py
index 722c2f27..1831ae86 100644
--- a/ush/set_gridparams_ESGgrid.py
+++ b/ush/set_gridparams_ESGgrid.py
@@ -72,8 +72,8 @@ def set_gridparams_ESGgrid(lon_ctr,lat_ctr,nx,ny,halo_width,delx,dely,pazi):
     #

     return (lon_ctr,lat_ctr,nx,ny,pazi,halo_width,stretch_factor,
-            del_angle_x_sg,
-            del_angle_y_sg,
+            round(del_angle_x_sg,10),
+            round(del_angle_y_sg,10),
```
The results are here:
```
/scratch2/BMC/gsd-hpcs/Daniel.Abdi/expt_dirs
```
The develop branch is half way through the forecast but all 6 netcdf files match bit-for-bit with the others.

danielabdi-noaa:
> @danielabdi-noaa I ran 3 WE2E tests on Hera with a version of this PR from Wednesday. They all passed, but the jobs in my crontab weren't cleared afterwards. I will rerun them with the latest version, maybe I made a mistake.

tests/WE2E/get_WE2Etest_names_subdirs_descs.sh
```diff
@@ -1689,49 +1605,8 @@ containing information on all WE2E tests:
...
-#   See description of ${DOT_OR_USCORE} in the configuration file.
-#
-# Then, if outvarname_test_descs is set to "some_array", the exact string 
-# we want to pass to eval is:
-#
-#   some_array=('Please see the User'"'"'s Guide.' 'See description of ${DOT_OR_USCORE} in the configuration file.')
-#
-    test_descs_esc_sq=()
-    for (( i=0; i<=$((num_tests-1)); i++ )); do
-      test_descs_esc_sq[$i]=$( printf "%s" "${test_descs[$i]}" | \
```
gsketefian, line 1730: I guess removal of this substitution is what's causing the apostrophes (single quotes) to become double quotes.

danielabdi-noaa:
> @danielabdi-noaa I ran 3 WE2E tests on Hera with a version of this PR from Wednesday. They all passed, but the jobs in my crontab weren't cleared afterwards. I will rerun them with the latest version, maybe I made a mistake.

It should clear them. I have run 68 we2e test cases simultaneously on Hera and my crontab was empty after all is done.
Note that this PR does not change anything related to crontab management so if there is an issue it must come from develop.

Sorry I closed this PR by accident, pressed "closed with comment" instead of "comment"

tests/WE2E/get_WE2Etest_names_subdirs_descs.sh
danielabdi-noaa: @gsketefian Yes, I replace single quotes with double quotes, not only in we2e test description but also in comments in crontab contents. Usually you get empty crontab content but Venita's automated testing setup had crontab contents with comments that had single quotes. Please take a look at PR #816 that describs the issue and how I solved it. I didn't find a better solution than that at the time ...

tests/WE2E/get_WE2Etest_names_subdirs_descs.sh
```diff
@@ -995,80 +979,11 @@ configuration files of the primary WE2E tests...
 # leading spaces, the hash symbol, and possibly another space and append
 # what remains to the local variable test_desc.
 #
-      config_fn="config.${test_name}.sh"
-      test_desc=""
-      while read -r line; do
```
gsketefian, line 1000: I think removing this section is what causes newlines to not show up properly in the spreadsheet and for indentation to be done improperly.

tests/WE2E/get_WE2Etest_names_subdirs_descs.sh
danielabdi-noaa: The replacement is done whenever you produce a shell string equivalent of other formats. Bash does not like single quotes since they have been causing me problems in unexpected places (two places I already mentioned but there were  of others too), so I decided to get rid of them in the config parser:
```Python
            # replace some problematic chars
            v1 = v1.replace("'", '"')
            v1 = v1.replace("\n", " ")
            # end problematic
            shell_str += f"{k}='{v1}'\n"
```
Also I replace newlines with space for the same reason, without which I would not be able to store the test description in `var_defns.sh`. It is debatable whether that is needed there but new lines and single quotes in strings does cause lots of problems in bash scripts.

tests/WE2E/test_configs/grids_extrn_mdls_suites_community/config.grid_CONUS_3km_GFDLgrid_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16.yaml
```diff
@@ -0,0 +1,25 @@
...
+  RUN_ENVIR: community
+workflow:
+  CCPP_PHYS_SUITE: FV3_GFS_v16
+  DATE_FIRST_CYCL: '20190701'
+  DATE_LAST_CYCL: '20190701'
+  CYCL_HRS:
+    - 0
+  FCST_LEN_HRS: 6
+  PREEXISTING_DIR_METHOD: rename
+task_run_fcst:
```
gsketefian: It would be nice to list the different task dictionaries in the order in which they appear in the xml, both in these test config files and in `config_defaults.yaml`, i.e.
```
task_make_grid
task_make_orog
task_make_sfc_climo
task_get_extrn_ics
task_get_extrn_lbcs
task_make_ics
task_make_lbcs
task_run_fcst
task_run_post
task_get_obs_ccpp
task_get_obs_mrms
task_get_obs_ndas
task_run_vx_... (the vx tasks are still changing)
```

tests/WE2E/get_WE2Etest_names_subdirs_descs.sh
danielabdi-noaa: Here is an example `var_denfs.sh` file with a test description
```Shell
# [metadata]
description='This test checks the capability of the workflow to have the user specify a new grid (as opposed to one of the predefined ones in the workflow) of GFDLgrid type.  Note that this test sets the workflow variable   GFDLgrid_USE_NUM_CELLS_IN_FILENAMES to "TRUE" (which is its default value); see the UFS SRW User"s Guide for a description of this variable. The difference between this test and the one named   custom_GFDLgrid__GFDLgrid_USE_NUM_CELLS_IN_FILENAMES_eq_TRUE is that this one uses almost no stretching by setting the workflow variable GFDLgrid_STRETCH_FAC very close to 1.  Setting it exactly to 1 used to cause the workflow to fail because it caused the GFDL grid generator to assume a global grid.  This bug should be rechecked, e.g. by setting GFDLgrid_STRETCH_FAC to exactly 1 below.  If the grid generation succeeds, then this test can be removed.'
version='1.0'

# [user]
RUN_ENVIR='community'
MACHINE='HERA'
MACHINE_FILE='/scratch2/BMC/gsd-hpcs/Daniel.Abdi/ufs-srweather-app/regional_workflow/ush/machine/hera.sh'
ACCOUNT='zrtrr'

# [platform]
WORKFLOW_MANAGER='rocoto'
NCORES_PER_NODE='40'
```
That multiline description would cause problems if it had "single quotes" or "new line" characters.

tests/WE2E/test_configs/grids_extrn_mdls_suites_community/config.grid_CONUS_3km_GFDLgrid_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16.yaml
danielabdi-noaa: @gsketefian I have made the change you suggested to both `config_defaults.yaml` and the we2e config files.
Will push later once I run some tests.

tests/WE2E/get_WE2Etest_names_subdirs_descs.sh
gsketefian: @danielabdi-noaa In this case of the test descriptions, it doesn't seem to make sense to do that replacement.  What you really want to do is take the description and put it into the csv file exactly as is.  Replacing apostrophes with double quotes would be very confusing to someone reading the test description in the csv.  Since this is separate from the treatment of crontab lines, is it possible to not do the replacement when dealing with these descriptions?  There must be a way to just take a literal string from the yaml file (even one that includes single quotes and newlines) and place it exactly as-is into the csv.

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220906230509/ufs-srweather-app
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220906230509/ufs-srweather-app/test/../bin_intel/bin/adjust_soiltq.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220906230509/ufs-srweather-app/test/../bin_intel/bin/check_imssnow_fv3lam.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220906230509/ufs-srweather-app/test/../bin_intel/bin/fv3lam_nonvarcldana.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220906230509/ufs-srweather-app/test/../bin_intel/bin/gen_annual_maxmin_GVF.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220906230509/ufs-srweather-app/test/../bin_intel/bin/gen_cs.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220906230509/ufs-srweather-app/test/../bin_intel/bin/gen_ensmean_recenter.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220906230509/ufs-srweather-app/test/../bin_intel/bin/lakesurgery.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220906230509/ufs-srweather-app/test/../bin_intel/bin/process_imssnow_fv3lam.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220906230509/ufs-srweather-app/test/../bin_intel/bin/process_larccld.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220906230509/ufs-srweather-app/test/../bin_intel/bin/process_Lightning.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220906230509/ufs-srweather-app/test/../bin_intel/bin/process_metarcld.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220906230509/ufs-srweather-app/test/../bin_intel/bin/process_NSSL_mosaic.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220906230509/ufs-srweather-app/test/../bin_intel/bin/process_updatesst.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220906230509/ufs-srweather-app/test/../bin_intel/bin/ref2tten.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220906230509/ufs-srweather-app/test/../bin_intel/bin/update_bc.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220906230509/ufs-srweather-app/test/../bin_intel/bin/update_GVF.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220906230509/ufs-srweather-app/test/../bin_intel/bin/update_ice.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/944706962/20220906230509/ufs-srweather-app/test/../bin_intel/bin/use_raphrrr_sfc.exe does NOT exist'
+ echo 'BUILD(S) FAILED'
+ msg=FAIL
+ echo FAIL
Build failed
ci-hera-intel-WE

tests/WE2E/get_WE2Etest_names_subdirs_descs.sh
danielabdi-noaa: @gsketefian I have made the we2e test description specifically (not `var_defns.sh` or other places) to keep existing single quotes and new lines. Pyyamls folds newlines when `>-` is used for multiline string, so I had to replace that with `|-`. Even then when you write the multiline string back with `yaml.dump` it will not keep the format. I needed to use a hack to make pyyaml use block format when dumping multiline strings, but seems to work fine now.

For the record though, I liked the way it was where the test description is one long line, because, in the excel sheet it will be wrapped around cell width and looks natural. While with this new fix that keeps the format first used in the config file, widening the cell has no effect. Also now it is not consistent with the test description in `var_defns.sh`, that must replace single quotes and newlines to work properly. In any case, please test it and let me know what you think about it.

gsketefian: @danielabdi-noaa Sorry I wasn't clear about the newlines (or suggested the wrong thing).  Yes, I like it to be one long string too (that's how it is now) because that's what works best in a spreadsheet, except when there is a blank line, which signifies a new paragraph or newline.  Currently, if you have this in the description in the test config file (in bash):
```
# Hello there.  This is a description of 
# this test.  Define a couple of variables:
#
#   MY_VAR1 = "abcd"
#   MY_VAR2 = "efgh${MY_VAR1}"
#   MY_VAR3 = 'klmn${MY_VAR2}'
#
# Done with description.
```
it will show up like this in a cell in the spreadsheet:
```
Hello there.  This is a description of this test.  Define a variable:

  MY_VAR1 = "abcd"
  MY_VAR2 = "efgh${MY_VAR1}"
  MY_VAR3 = 'klmn${MY_VAR2}'

Done with description.
```
So the lines without a blank line in between get concatenated into one long line, but the blank lines cause there to be a newline to be inserted.  Also, the indentations for the lines `MY_VAR... = ...` are kept, and no changes are made to single and double quotes.  Maybe there is a way to specify a newline character, e.g. `\n`, so it can be used only where required (i.e. where we want a blank line).  Anyway, I'll leave it up to you to decide what is best.  Feel free to change the `|-` back to `>-` (I actually prefer that).

tests/WE2E/get_WE2Etest_names_subdirs_descs.sh
gsketefian: Ok, I see.  I didn't know the test description goes in `var_defns.sh`.  It didn't used to and it's not necessary since it's not used by any tasks (as far as I know), but that's fine if you feel it should be there.  And yes, it would be a problem to have single quotes and newlines in there.  I wonder if you can escape them.  But I'll leave it up to you how you want to handle it (also see my other comment above in this file).

christinaholtNOAA (approved):

mkavulich:
@danielabdi-noaa We discussed this PR in the DTC meeting just now and we are comfortable merging the PR today provided we can ensure bit-for-bit equivalence. Is it possible to modify your code to do this fairly easily? My understanding in the problem is simply in a truncation of the digits in the regional_grid.nml; can you modify your code to use the full, un-truncated precision as currently occurs in develop? We understand that this precision isn't scientifically relevant, but given the size of the PR if we can ensure there is bit-for-bit equivalence that would make us much more comfortable with the merge.

ush/python_utils/process_args.py
```diff
@@ -1,10 +1,13 @@
 #!/usr/bin/env python3
 
 from textwrap import dedent
+
```
gsketefian, line 4: @danielabdi-noaa Just wondering why you still need this script.  The bash version of this script was something I had written to handle arguments for bash scripts/functions.  Doesn't `argparse` take care of arguments now?  (I would grep for this myself but Hera is down at the moment.)

ush/config_defaults.yaml
```diff
@@ -0,0 +1,2252 @@
+#----------------------------
```
gsketefian, line 1: @danielabdi-noaa Can you describe a bit how/where the different sections (dictionaries) in the config file (is it called `config.yaml`?) are accessed?  I think the idea behind dividing the variables into sections was to only use those sections that are necessary for a given rocoto task.  Where are those determined for each task?  Thx.

danielabdi-noaa: @gsketefian There is this now closed PR #789 that did what you described. The variables needed only for the specific task being run are sourced. I may revive it once the dust is settled on this PR.

panll: I tested it on Jet. (Hera is under maintenance today).  It's byte-for-byte match now. Thanks! @danielabdi-noaa @JeffBeck-NOAA @mkavulich

panll (approved):

JeffBeck-NOAA (approved):
Approving based on @linlin-pan's testing and @gsketefian's review.

PR #787, opened 2022-06-05
[develop] Rename variables related to GFDLgrid-type grids

gsketefian:
## DESCRIPTION OF CHANGES: 
For clarity, rename variables related to GFDLgrid-type grids:
1) Rename `GFDLgrid_RES` to `GFDLgrid_NUM_CELLS`.
2) Rename `GFDLgrid_USE_GFDLgrid_RES_IN_FILENAMES` to `GFDLgrid_USE_NUM_CELLS_IN_FILENAMES`.
3) Rename WE2E tests on GFDLgrid-type custom grids according to variable renaming above.

Related improvements:
* In `link_fix.sh`, remove creation of unnecessary symlink when using a GFDLgrid-type grid with `GFDLgrid_USE_NUM_CELLS_IN_FILENAMES` is set to `"FALSE"`.
* In WE2E test configuration files that use custom grids, move the definition of the variable `POST_OUTPUT_DOMAIN_NAME` to the beginning of the custom grid definition because it is really the name of the custom native grid, and it's most appropriate to place that at the beginning of the grid definition section.
* Reduce `WTIME_MAKE_OROG` from 1 hour to 20 minutes since on Hera, 1 hour causes the task to wait quite a while in the queue.  This used to be 20 minutes but was changed to 1 hour because 20 minutes was not sufficient to complete the task on `WCOSS_DELL_P3`.  Thus, reset it to 1 hour only in the machine file for `WCOSS_DELL_P3`.
* Minor fixes to comments in python plotting scripts.
* Edits to other comments.

## TESTS CONDUCTED: 
Ran the four WE2E tests that use custom grids (listed below) on Hera.  All were successful.
* `custom_ESGgrid`
* `custom_GFDLgrid`
* `custom_GFDLgrid__GFDLgrid_USE_NUM_CELLS_IN_FILENAMES_eq_FALSE` (previously `custom_GFDLgrid__GFDLgrid_USE_GFDLgrid_RES_IN_FILENAMES_eq_FALSE`)
* `custom_GFDLgrid__GFDLgrid_USE_NUM_CELLS_IN_FILENAMES_eq_TRUE` (previously `custom_GFDLgrid__GFDLgrid_USE_GFDLgrid_RES_IN_FILENAMES_eq_TRUE`)

## DEPENDENCIES:
PR #[310](https://github.com/ufs-community/ufs-srweather-app/pull/310) in `ufs-srweather-app`.

## DOCUMENTATION:
Updates to documentation are in PR #[310](https://github.com/ufs-community/ufs-srweather-app/pull/310) in `ufs-srweather-app`.


willmayfield (approved):

tests/WE2E/test_configs/wflow_features/config.custom_GFDLgrid__GFDLgrid_USE_GFDLgrid_RES_IN_FILENAMES_eq_FALSE.sh
```diff
@@ -31,6 +31,7 @@ LBC_SPEC_INTVL_HRS="3"
 #
```
JeffBeck-NOAA, line 31: Can the name of this test be updated?  "GFDLgrid" is used three times.

gsketefian: The first part, i.e. `custom_GFDLgrid`, is to indicate that this test is for creating a custom GFDLgrid.  The rest, i.e. `GFDLgrid_USE_GFDLgrid_RES_IN_FILENAMES_eq_FALSE` is to indicate that a variable named `GFDLgrid_USE_GFDLgrid_RES_IN_FILENAMES` is being set to `FALSE`.  In this variable name, the first `GFDLgrid` indicates that it is a variable that is only relevant for GFDLgrid type grids (there are a set of variables that start with `GFDLgrid_...` that only apply to GFDLgrid-type grids, and another set that start with `ESGgrid_...` that only apply to ESG-type grids).  The rest of the variable name, i.e. `USE_GFDLgrid_RES_IN_FILENAMES` indicates what the flag is about.  In this case, the flag toggles between using the C-resolution from the global grid (i.e. using the number of grid points per tile for the C-res in the file names, e.g. `C48`, `C96`, ..., or `C768`, which I call the GFDLgrid resolution, or `GFDLgrid_RES`; this number doesn't tell you anything about the actual resolution of your regional grid because it doesn't take into account any stretching or refinement) and the C-resolution obtained by finding the number of grid points (per tile) that an **unstretched** global cubed-sphere grid would need to have the same actual resolution in kms of the regional grid (e.g. `C3334`; from this, you can back-calculate what the resolution is in kms).  So we can avoid repeating "GFDLgrid" by renaming the variable `GFDLgrid_USE_GFDLgrid_RES_IN_FILENAMES`.  Let me know if you have suggestions.

JeffBeck-NOAA: >  So we can avoid repeating "GFDLgrid" by renaming the variable GFDLgrid_USE_GFDLgrid_RES_IN_FILENAMES.

Thank for the explanation!  The variable is already named `GFDLgrid_USE_GFDLgrid_RES_IN_FILENAMES`, though, right?

gsketefian: Yes.  But I'm open to renaming it as part of this PR if we can think of a good name, which would imply renaming the file.

JeffBeck-NOAA: I suppose you could call it `GFDLgrid_USE_EQUIV_CRES_IN_FILENAMES`, and set it to `true` in this case?

gsketefian: @JeffBeck-NOAA I renamed the variable from `GFDLgrid_USE_GFDLgrid_RES_IN_FILENAMES` to `GFDLgrid_USE_NUM_CELLS_IN_FILENAMES` and renamed the tests accordingly.  Now there are only two `GFDLgrid`s in the file names.  But the renaming as well as a bunch of improvements turned out to be a bigger change than the original intent of the PR, so I rephrased the PR title.  It will require a re-review by both you and @willmayfield, I think.

chan-hoo (approved):

JeffBeck-NOAA (approved):

ush/config_defaults.sh
```diff
@@ -878,14 +878,14 @@ GRID_GEN_METHOD=""
...
-# each tile of the parent global grid.  Note that the name of this parameter
-# is really a misnomer because although it has the stirng "RES" (for 
-# "resolution") in its name, it specifies number of grid cells, not grid
-# size (in say meters or kilometers).  However, we keep this name in order
-# to remain consistent with the usage of the word "resolution" in the 
-# global forecast model and other auxiliary codes.
+# GFDLgrid_NUM_CELLS:
+# Number of grid cells in each of the two horizontal directions (x and 
+# y) on each tile of the parent global grid.  Note that the name of this 
+# parameter is really a misnomer because although it has the stirng "RES"
```
willmayfield: Shouldn't the discussion of "RES" be removed if taking it out of the variable name? (Also there's a typo "stirng")

gsketefian: @willmayfield Yes, it should!  I'll fix it.

PR #791, opened 2022-06-08
Update UFS WM to include CCPP-v6 + fixes for CI/CD

mark-a-potts:
This PR fixes a number of issues related to both ci/cd and updating the UFS weather model to include CCPP-v6. 

References to lndp_each_step are removed due to changes in the input.nml on the UFS-WM side.

The miniconda version is now explicitly set for jet, so that the correct conda module gets loaded.

The "location" variable in ush/machine/hera.sh has been changed so that it does not point to an rstprod directory. (Maybe there is a better solution to this?) The role account on hera does not have rstprod access, so all e2e tests there were failing to launch previously. 

These changes will work in concert with ufs-srweather-app PR 293. https://github.com/ufs-community/ufs-srweather-app/pull/293


EdwardSnyder-NOAA (approved):

ush/config_defaults.sh
```diff
@@ -1910,7 +1910,6 @@ LSM_SPP_LSCALE=( "150000" "150000" "150000" "150000" "150000" "150000" "150000"
 ISEED_LSM_SPP=( "9" )
 LSM_SPP_VAR_LIST=( "smc" "vgf" "alb" "sal" "emi" "zol" "stc" )
 LSM_SPP_MAG_LIST=( "0.2" "0.001" "0.001" "0.001" "0.001" "0.001" "0.2" )
-LSM_SPP_EACH_STEP="true" #Sets lndp_each_step=.true.
 #
```
JeffBeck-NOAA, line 1921: The comments above related to LSM_SPP_EACH_STEP need to be removed:

`# The default perturbation frequency is determined by the fhcyc namelist entry. Since that parameter is set to zero in the SRW App, use LSM_SPP_EACH_STEP to perturb every time step.`

JeffBeck-NOAA (changes requested):

ush/config_defaults.sh
JeffBeck-NOAA: Information related to this namelist entry needs to be removed from the release documentation as well.  We'll also want to modify the develop branch documentation when we push this change there.  I can work on that when we update the hash of ufs-weather-model in develop. Attn: @mark-a-potts @gspetro-NOAA.

mark-a-potts: We probably need to get a bit more info from Clara Draper, who made the change in the ufs-wm. These current changes are all still for testing, but I think we are getting closer. I now have the changes to the weather model all rolled up and part of their own branch under ufs-community.

JeffBeck-NOAA: @mark-a-potts, yep, I was going to pull up her PR to get some more info on how things are handled now that this namelist entry has been removed.  It would be good to at least remove the comment above this namelist entry in config_defaults.sh.  I'm glad to hear you're much closer to getting the ufs-weather-model release branches finalized!

mark-a-potts: Done.

JeffBeck-NOAA (approved):

PR #775, opened 2022-05-20
Update scripts to work with the latest hashes of UFS_UTILS and UPP

chan-hoo:
## DESCRIPTION OF CHANGES: 
- Some scripts are updated to work with the latest hashes of UFS_UTILS and UPP.

## TESTS CONDUCTED: 
- WE2E tests:
grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_2017_gfdlmp_regional
grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_RAP_suite_HRRR
grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15_thompson_mynn_lam3km
grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_RRFS_v1beta
grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha

## DEPENDENCIES:
- ufs-community/ufs-srweather-app/pull/266

PR #795, opened 2022-06-14
[Release] Fixed file path for FIXgsm

EdwardSnyder-NOAA:
## DESCRIPTION OF CHANGES: 
Added back the `fix_dir_input_grid` variable to the `exregional_make_ics.sh` and `exregional_make_lbcs.sh` scripts.

## TESTS CONDUCTED: 
Ran test on Hera and more tests will be conducted through the Jenkins pipeline.

## DEPENDENCIES:
None.

## DOCUMENTATION:
No documentation changes needed.

## CONTRIBUTORS (optional): 
@mark-a-potts


mark-a-potts (changes requested):
This looks good. I am not sure we need the parameter in exregional_make_ics.sh, but we can test it to make sure it doesn't break anything. Can you also switch over all the "develop" paths in ush/machine/*.sh to "v2p0" and we can combine two fixes into one?

EdwardSnyder-NOAA:
> This looks good. I am not sure we need the parameter in exregional_make_ics.sh, but we can test it to make sure it doesn't break anything. Can you also switch over all the "develop" paths in ush/machine/*.sh to "v2p0" and we can combine two fixes into one?

Made the changes as requested.

mark-a-potts (approved):

scripts/exregional_make_lbcs.sh
```diff
@@ -443,6 +443,7 @@ list file has not specified for this external LBC model (EXTRN_MDL_NAME_LBCS):
 #
 settings="
 'config': {
+ 'fix_dir_input_grid': ${FIXgsm},
```
gsketefian, line 446: @EdwardSnyder-NOAA @mark-a-potts Just curious if you know whether this line was necessary to fix the `make_lbcs` failures.  If it was, it means something must have changed either in `chgres_cube` or the fix files.  (I always thought the external model data files already contained a description of the external model's grid.)  To keep the namelist as simple as possible, it would be good to remove it if it's not really needed.  Not necessarily for this PR; I'm thinking more about when we move these changes into `develop`.  Thanks.

mark-a-potts: From our testing, it was definitely needed. I am not sure that it is needed for the exregional_make_ics.sh script, though. Without that line in the namelist, chgres_cube was looking for NULL/latlon.grid3.nc (or something very similar) and segfaulting when it could not open the file.

gsketefian: After digging into this a bit more, looks like this line was removed very recently from both files in PR #775.  I guess that was needed due to a change in UFS_UTILS (not sure what), but the `release/public-v2` branch still needs the lines (although I haven't actually tested the `develop` branch with the two tests that were failing in the release branch to see if it works; maybe I'll give that a shot now).

Issue #818, opened 2022-08-08
ensemble with stochastic physics issue


danielabdi-noaa:
# Description

Current develop does not work for ensemble with stochastic physics that requires setting seeds due to a bug in the python script `set_FV3nml_ens_stoch_seeds.py`. Thanks to @JeffBeck-NOAA for spotting this problem.

## Steps to Reproduce
Set ensemble and stochastic physics options like below:
```
DO_ENSEMBLE="TRUE"
NUM_ENS_MEMBERS="2"
DO_SPPT="TRUE" #Any stochastic physics options causes the error, but I used SPPT here
```

## Additional Context

Machine=Hera

## Output

```
Creating links in the current run directory to cycle-independent model
input files in the main experiment directory...
usage: set_FV3nml_ens_stoch_seeds.py [-h] -c CDATE
set_FV3nml_ens_stoch_seeds.py: error: unrecognized arguments: --path-to-defns /scratch2/BMC/fv3lam/beck/FV3-LAM/expt_dirs/SPPT/var_defns.sh
```

PR #819, opened 2022-08-08
Bugfix for ensemble+stochastic physics issue and new WE2E test case for stochastic physics.

danielabdi-noaa:
## DESCRIPTION OF CHANGES: 
Fixes the problem described in issue #818 and more. 

- The path to `var_defns.sh` is passed to the script that sets seeds for ensemble members, namely `set_FV3nml_ens_stoch_seeds.py`
- Bug fix for DO_SKEB/SHUM stochastic seeding options
- Added a new WE2E test case to monitor future changes to ensemble + all stochastic physics options
- Some changes to printing of info/error messages to help future debugging of create_diag/model tables.

## TESTS CONDUCTED: 
Note that none of the WE2E tests have the required setup to reproduce this bug.

The following tests are run so far:

* community_ensemble_2mems: to make sure existing ensemble run without stochastic physics works ok
* community_ensemble_2mem tested with DO_SPPT turned on
* Original test case from Jeff that revealed the issue tested with the following changes:
1. DO_SPPT="TRUE"
2. DO_SKEB="TRUE"
3. DO_SHUM="TRUE"
4. DO_SPPT="TRUE" and DO_SPP="TRUE"

## DEPENDENCIES:
None

## DOCUMENTATION:
None

## ISSUE (optional): 
#818 

## CONTRIBUTORS (optional): 
@JeffBeck-NOAA


JeffBeck-NOAA:
Thanks for getting this fix put together, @danielabdi-noaa!  It's gotten past loading the var_defns.sh file, but is not stuck on the path to input.nml to modify the stochastic physics entries.  It's looking for a path without the cycle hour, so that needs to be added (bold highlighting is mine):

Creating links in the current run directory to cycle-independent model
input files in the main experiment directory...
usage: set_FV3nml_ens_stoch_seeds.py [-h] -o OUTFILE [-c [FILE, SECTION]]
                                     [-i INPUT_NML] [-n NML] [-t {nml,yaml}]
                                     [-u YAML STRING] [-q]
set_FV3nml_ens_stoch_seeds.py: error: argument -o/--outfile: /scratch2/BMC/fv3lam/beck/FV3-LAM/expt_dirs/new_SKEB/**20190615**/mem1/input.nml is not a writable path!
  File "/scratch2/BMC/fv3lam/beck/FV3-LAM/ufs-srweather-app_fix/regional_workflow/ush/set_FV3nml_ens_stoch_seeds.py", line 137, in <module>
    set_FV3nml_ens_stoch_seeds(str_to_type(args.cdate))
  File "/scratch2/BMC/fv3lam/beck/FV3-LAM/ufs-srweather-app_fix/regional_workflow/ush/set_FV3nml_ens_stoch_seeds.py", line 103, in set_FV3nml_ens_stoch_seeds
    print_err_msg_exit(dedent(f'''
  File "/scratch2/BMC/fv3lam/beck/FV3-LAM/ufs-srweather-app_fix/regional_workflow/ush/python_utils/print_msg.py", line 18, in print_err_msg_exit
    traceback.print_stack(file=sys.stderr)

            Call to python script set_namelist.py to set the variables in the FV3
            namelist file that specify the paths to the surface climatology files
            failed.  Parameters passed to this script are:
              Full path to base namelist file:
                FV3_NML_FP = "/scratch2/BMC/fv3lam/beck/FV3-LAM/expt_dirs/new_SKEB/input.nml"
              Full path to output namelist file:
                fv3_nml_ensmem_fp = "/scratch2/BMC/fv3lam/beck/FV3-LAM/expt_dirs/new_SKEB/20190615/mem1/input.nml"
              Namelist settings specified on command line (these have highest precedence):
                settings =
            nam_stochy:
  iseed_skeb: 20190615013
nam_spperts: {}

danielabdi-noaa:
@JeffBeck-NOAA The issue seems to be that it is getting wrong path to ens members `input.nml`. Somehow the hour 18 is missing from `20190615`. The logic in `date_to_str` in python_utils is such that when the hour and minute are zero it converts to YYYYMMDD. Why is the hour 0 instead of 18 because you have `2019061518` directly with data for ensemble members.

JeffBeck-NOAA:
@danielabdi-noaa, the cycle directory is always YYYYMMDDHH, so for this specific test, 2019061518 is correct.  I'm not sure why the cycle hour is missing from the `date_to_str` output in the Python version of the set_FV3nml_ens_stoch_seeds.py script.

danielabdi-noaa:
> @danielabdi-noaa, the cycle directory is always YYYYMMDDHH, so for this specific test, 2019061518 is correct. I'm not sure why the cycle hour is missing from the `date_to_str` output in the Python version of the set_FV3nml_ens_stoch_seeds.py script.

That is because it is forced to be short version YYYYMMDD in the python script. I will fix it and push a commit shortly.

danielabdi-noaa:
@JeffBeck-NOAA I have pushed a commit that forces hour to be appended to cycle dates even though hour is 0. Could you please try again?

scripts/exregional_run_fcst.sh
```diff
@@ -434,7 +434,7 @@ if [ ${WRITE_DOPOST} = "TRUE" ]; then
   cp_vrfy ${UPP_DIR}/parm/params_grib2_tbl_new .
 fi
 
-if [ "${DO_ENSEMBLE}" = TRUE ] && ([ "${DO_SPP}" = TRUE ] || [ "${DO_SPPT}" = TRUE ] || [ "${DO_SHUM}" = TRUE ] \
+if [ "${DO_ENSEMBLE}" = TRUE ] && ([ "${DO_SPP}" = TRUE ] || [ "${DO_SPPT}" = TRUE ] || [ "${DO_SHUM}" = TRUE ] || \
    [ "${DO_SKEB}" = TRUE ] || [ "${DO_LSM_SPP}" =  TRUE ]); then
```
JeffBeck-NOAA, line 438: Glad you found this too!  I was going to open a separate PR, but it's much easier to include it here.  Thanks!

JeffBeck-NOAA:
@danielabdi-noaa, thanks for the commits, things are now working!  One last minor fix: From lines 80-97, change `spperts` to `sppperts`.  Thanks!

danielabdi-noaa:
@JeffBeck-NOAA I have made tests with your test case and make sure it works for different ways of initializing seeds.

- DO_SPPT="TRUE"
- DO_SKEB="TRUE"
- DO_SHUM="TRUE"
- DO_SPPT="TRUE" and DO_SPP="TRUE"

My results are at:
```
   /scratch2/BMC/gsd-hpcs/Daniel.Abdi/expt_dirs
```
with test cases named SPPT/SHUM/SKEB/SPPT-SPP.

In the process I have discovered a bug which came from the shell workflow generation code. I noticed that SHUM/SKEB runs keep the default values of iseed_skeb/shum =2/3 because `set_FV3nml_ens_stoch_seeds.py` was not being called due to a "syntax error" that was not being reported as so for some reason.
```Bash
+++ b/scripts/exregional_run_fcst.sh
@@ -434,7 +434,7 @@ if [ ${WRITE_DOPOST} = "TRUE" ]; then
   cp_vrfy ${UPP_DIR}/parm/params_grib2_tbl_new .
 fi
 
-if [ "${DO_ENSEMBLE}" = TRUE ] && ([ "${DO_SPP}" = TRUE ] || [ "${DO_SPPT}" = TRUE ] || [ "${DO_SHUM}" = TRUE ] \
+if [ "${DO_ENSEMBLE}" = TRUE ] && ([ "${DO_SPP}" = TRUE ] || [ "${DO_SPPT}" = TRUE ] || [ "${DO_SHUM}" = TRUE ] || \
```
It looks to me like there is a missing "||" operator before the continuation mark, so I have added that. I have confirmed that default values do get changed after this fix, so please let me know if this is what you intended. I have no idea why it was not complaining without the or operator at the end?

Another potential issue I noticed is  the iseed_sppt/shum etc values exceed integer limits (2^32) now after adding the hour to seed calculation. If those values are passed to fortran/c subroutines with 32 bit integers, it may get truncated.

scripts/exregional_run_fcst.sh
danielabdi-noaa: Ah this was the issue you found earlier. Sorry I got in your way.

JeffBeck-NOAA: Not a problem at all!  Thanks for taking care of it!

JeffBeck-NOAA:
@danielabdi-noaa, thanks for conducting these tests!  I took a look at your directory on Hera and things look correct and mimic my own runs.  Thanks as well for fixing the "or" bug in the "if" statement which controlled the call to the set_FV3nml_ens_stoch_seeds.py script.  That bug effectively kept the workflow from creating the correct seeds for SKEB and LSM_SPP, unless they were used in combination with one of the other stochastic physics options (SPP, SPPT, or SHUM).  I believe the shell script wasn't failing because "set -e" is not included in the script.  I spoke with @gsketefian, who mentioned that bash will allow some errors unless this flag is set.  However, he said that setting "set -e" can also cause other undesirable results, which is why it wasn't included in our original shell workflow scripts.

Just to confirm, are we OK with the lengths of iseed_sppt/shum/skeb/spp/lsm_spp?   I confirmed that they are all 64-bit integers in the ufs-weather-model code.

JeffBeck-NOAA:
@danielabdi-noaa, since you already fixed the "or" bug in the set_FV3nml_ens_stoch_seeds.py script, would you also be willing to add a 1-hr, regional, 3-km Indianapolis WE2E test turning all stochastic physics options (SPP, SPPT, SHUM, and SKEB) on?  This will ensure that the seed script and all stochastic physics are tested in the full workflow in the future.  Thank you!

ush/set_FV3nml_ens_stoch_seeds.py
```diff
@@ -97,6 +97,11 @@ def set_FV3nml_ens_stoch_seeds(cdate):
         }
 
     settings_str = cfg_to_yaml_str(settings)
+
+    print_info_msg(dedent(f'''
+        The variable \"settings\" specifying seeds in \"{FV3_NML_FP}\"
```
gsketefian, line 102: @danielabdi-noaa Seems like you're missing the rest of this sentence, e.g. maybe after the first line of the message, add:
```
has been set as follows:
```

ush/set_FV3nml_ens_stoch_seeds.py
```diff
@@ -108,9 +113,8 @@ def set_FV3nml_ens_stoch_seeds(cdate):
                 FV3_NML_FP = \"{FV3_NML_FP}\"
               Full path to output namelist file:
                 fv3_nml_ensmem_fp = \"{fv3_nml_ensmem_fp}\"
-              Namelist settings specified on command line (these have highest precedence):
-                settings =
-            {settings_str}'''))
+              Namelist settings specified on command line (these have highest precedence):\n
```
gsketefian, line 118: @danielabdi-noaa Is the `\n` at the end of this line to add an extra blank line before printing out `settings = ...`?

ush/set_FV3nml_sfc_climo_filenames.py
```diff
@@ -80,17 +79,16 @@ def set_FV3nml_sfc_climo_filenames():
     try:
         set_namelist(["-q", "-n", fv3_nml_base_fp, "-u", settings_str, "-o", FV3_NML_FP])
     except:
-        print_err_msg_exit(f'''
+        print_err_msg_exit(dedent(f'''
```
gsketefian, line 82: @danielabdi-noaa Just curious, does `dedent` remove all leading whitespace, or does it remove only the portion that is common to all lines (and thus keep relative indentation of lines)?

danielabdi-noaa:
> @danielabdi-noaa, thanks for conducting these tests! I took a look at your directory on Hera and things look correct and mimic my own runs. Thanks as well for fixing the "or" bug in the "if" statement which controlled the call to the set_FV3nml_ens_stoch_seeds.py script. That bug effectively kept the workflow from creating the correct seeds for SKEB and LSM_SPP, unless they were used in combination with one of the other stochastic physics options (SPP, SPPT, or SHUM). I believe there are two reasons the shell script wasn't failing: 1) because "set -e" is not included in the script. I spoke with @gsketefian, who mentioned that bash will allow some errors unless this flag is set. However, he said that setting "set -e" can also cause other undesirable results, which is why it wasn't included in our original shell workflow scripts, and 2) if the portion of "if" statement with the bug wasn't being evaluated (if SKEB and LSM_SPP were run in combination with one of the other stochastic physics options (SPP, SPPT, or SHUM), or they weren't run at all), then it wouldn't fail.
> 
Thank you and @gsketefian for the explanation! I was flabbergasted as to why it didn't complain at all.

> Just to confirm, are we OK with the lengths of iseed_sppt/shum/skeb/spp/lsm_spp? Gerard and I double-checked that they are all 64-bit integers in the ufs-weather-model code.

Yes, i did too and it uses 64-bit ints as seeds.

danielabdi-noaa:
> @danielabdi-noaa, since you already fixed the "or" bug in the set_FV3nml_ens_stoch_seeds.py script, would you also be willing to add a 1-hr, regional, 3-km Indianapolis WE2E test turning on all stochastic physics options (SPP, SPPT, SHUM, and SKEB) on (LSM SPP is still untested, so for now we can keep that one out)? This will ensure that the seed script and all stochastic physics are tested in the full workflow in the future. If you're unable to create one, or don't have the time, just let me know. Thank you!

Ok I will add those test cases to WE2E.

ush/set_FV3nml_ens_stoch_seeds.py
danielabdi-noaa: Will do, thanks!

ush/set_FV3nml_sfc_climo_filenames.py
danielabdi-noaa: Yes it removes leading white spaces upto the leftmost line (least number of leading spaces). Normally we don't have to do `dedent` for `print_info/err_msg` since those functions already do that. However, in this particular case where a "settings_str" is contructed from yaml (which is out of our control), we need to dedent the info message separately from the `settings_str`.

ush/set_FV3nml_ens_stoch_seeds.py
danielabdi-noaa: Yes. It is still does not look pretty but I guess better than before where the indentation was messed up.

ush/set_FV3nml_sfc_climo_filenames.py
gsketefian: Ok, thanks!

danielabdi-noaa:
> 

@JeffBeck-NOAA I have added a WE2E test case for testing stochastic physics options and make sure it run correctly.
The run is located at: `/scratch1/BMC/zrtrr/Daniel.Abdi/rrfs/benchmark-test/community_ensemble_2mems_stoch`
I did not change FCT_LEN_HRS to be 1-hour because I was not sure whether setting `LBC_SPEC_INTVL_HRS` to 1 hour is ok. Also, almost all other WE2E test cases use 6-hr forecast so I left it as it is, but please let me know if you have suggestions for changes.

JeffBeck-NOAA:
Thanks for the WE2E test, @danielabdi-noaa!  I took a look and I see a few changes that will be necessary.  First we want to run with the "FV3_HRRR" CCPP SDF, since that is the suite that can run with all SPP options.  Also, if you change `LBC_SPEC_INTVL_HRS` to 1, and you change EXTRN_MDL_NAME_ICS/LBCS to "HRRR", you can allow the WE2E to run for just a single hour, since we have hourly HRRR data available.  Having a 1-hr test at 3 km is good due to the computational expensive of that resolution, and is also what I added to ufs-weather-model as the regional stochastic physics regression test.  However, how long did you run take for six hours?  Since it's a very small domain (smaller than the regional domain I used in the ufs-weather-model regression test), maybe we can get away with six hours?  If a six-hour run is tolerable, we can then do LBC_SPEC_INTVL_HRS = 3.

JeffBeck-NOAA:
@danielabdi-noaa, also, I think you can probably just run a single cycle.  Currently I see:

DATE_FIRST_CYCL="20190701"
DATE_LAST_CYCL="20190702"
CYCL_HRS=( "12" )

But we can probably just do 20190701 with a 00 or 12Z initialization.

danielabdi-noaa:
@JeffBeck-NOAA The run I had before took about 80 mins based on the `create_WE2E_resource_summary.py` script. Decreasing number of cycles and fcst len should bring it down to 10-20 mins i guess. 
Quickly trying your suggestions, I've encountered an issue with HRRR LBCS/ICS. `get_ics` fail with the following message on Hera
```Bash
+ print_err_msg_exit 'Call to retrieve_data.py failed with a non-zero exit status.

The command was:

python3 -u /scratch2/BMC/gsd-hpcs/Daniel.Abdi/ufs-srweather-app/regional_workflow/ush/retrieve_data.py   --debug   --anl_or_fcst anl   --config /scratch2/BMC/gsd-hpcs/Daniel.Abdi/ufs-srweather-app/regional_workflow/ush/templates/data_locations.yml   --cycle_date 2019070112   --data_stores disk disk   --external_model HRRR   --fcst_hrs 0   --output_path /scratch1/BMC/zrtrr/Daniel.Abdi/rrfs/benchmark-test/community_ensemble_2mems_stoch/2019070112/HRRR/for_ICS   --summary_file extrn_mdl_var_defns.sh      --input_file_path /scratch2/BMC/det/UFS_SRW_App/develop/input_model_data/HRRR/2019070112
```

JeffBeck-NOAA:
@danielabdi-noaa, try using 2020081000 instead of 2019070112.  Those data are on disk.  Then change LBC_SPEC_INTVL_HRS to 3.  That should work.

danielabdi-noaa:
@JeffBeck-NOAA There were a couple of issues, one was the format of the ics/lbcs file. I needed to add
```Bash
EXTRN_MDL_FILES_ICS=( '{yy}{jjj}{hh}00{fcst_hr:02d}00' )
EXTRN_MDL_FILES_LBCS=( '{yy}{jjj}{hh}00{fcst_hr:02d}00' )
```
Another is that FCST_LEN should be divisble by LBCS_LEN
```
The forecast length (FCST_LEN_HRS) must be evenly divisible by the lateral
boundary conditions specification interval (LBC_SPEC_INTVL_HRS):
  FCST_LEN_HRS = 1
  LBC_SPEC_INTVL_HRS = 3
  rem = FCST_LEN_HRS%LBC_SPEC_INTVL_HRS = 1
Exiting with nonzero status.
```
HRRR LBCS are every 3 hrs so i guess minimum FCST_LEN should be 3 hrs. However, I vaguely recall past discussion as to why this check is present in the code.

For now as a compromise, I will use 3 hrs for both until we resolve 1-hr forcecast issue. This test takes about 15 min
`/scratch1/BMC/zrtrr/Daniel.Abdi/rrfs/benchmark-test/community_ensemble_2mems_stoch`

JeffBeck-NOAA:
@danielabdi-noaa, oh yes, sorry, I meant to say that we'd still use FCST_LEN = 6 and LBC_SPEC_INTVL_HRS = 3.  We can use FCST_LEN = 3, that would be fine.  Yeah, we put a check in the original workflow to make sure they were evenly divisible to avoid having people trying to run forecasts for which there are not enough LBC data.  If a 3-hr forecast takes aboute 15 minutes with the Indy grid, I think that sounds like a good compromise.  Thanks again for your work on this PR.  Approving now!

JeffBeck-NOAA (approved):

gsketefian:
@danielabdi-noaa Can you mention in the PR message that you're adding a new WE2E test for an ensemble forecast with stochastic options all turned on?  Otherwise good.  Approving.

gsketefian (approved):

PR #820, opened 2022-08-10
Update default data locations to make "config.community.sh" case work correctly

mkavulich:
## DESCRIPTION OF CHANGES: 
On a few different platforms (Cheyenne, Hera, and Orion), the default case specified in the file `ush/config.community.sh` does not work correctly because the input data can not be found. This PR modifies the default input data location on those platforms so that this test case will succeed "out of the box" (user only needs to specify machine and account) by pointing to the staged data file directories on those respective platforms.

## TESTS CONDUCTED: 
Ran the case specified by `ush/config.community.sh` on Cheyenne, Hera, and Orion; all succeeded after these changes. Also ran it on Jet; this succeeded because the already-specified default location on Jet has the appropriate data.

Did not run any WE2E tests, although this should not change any of those results. Will run for Hera and Jet using CI tags to make sure nothing major has broken.

## DEPENDENCIES:
None

## DOCUMENTATION:
None

### DIFF
--- a/ush/machine/cheyenne.sh
+++ b/ush/machine/cheyenne.sh
@@ -13,7 +13,11 @@ function file_location() {
   case ${external_model} in
 
     "FV3GFS")
-      location='/glade/p/ral/jntp/UFS_SRW_App/develop/input_model_data/FV3GFS/grib2/${yyyymmdd}${hh}'
+      case $external_file_fmt in
+        "grib2")
+          location='/glade/p/ral/jntp/UFS_SRW_App/develop/input_model_data/FV3GFS/grib2/${yyyymmdd}${hh}'
+          ;;
+      esac
       ;;
 
   esac
--- a/ush/machine/cheyenne.sh
+++ b/ush/machine/cheyenne.sh
@@ -13,7 +13,7 @@ function file_location() {
   case ${external_model} in
 
     "FV3GFS")
-      location='/glade/p/ral/jntp/UFS_CAM/COMGFS/gfs.${yyyymmdd}/${hh}'
+      location='/glade/p/ral/jntp/UFS_SRW_App/develop/input_model_data/FV3GFS/grib2/${yyyymmdd}${hh}'
       ;;
 
   esac
--- a/ush/machine/hera.sh
+++ b/ush/machine/hera.sh
@@ -3,7 +3,6 @@
 function file_location() {
 
   # Return the default location of external model files on disk
-  # Hera does not currently have any files staged on disk.
 
   local external_file_fmt external_model location
 
@@ -11,6 +10,13 @@ function file_location() {
   external_file_fmt=${2}
 
   location=""
+  case ${external_model} in
+
+    "FV3GFS")
+      location='/scratch2/BMC/det/UFS_SRW_App/develop/input_model_data/FV3GFS/grib2/${yyyymmdd}${hh}'
+      ;;
+
+  esac
   echo ${location:-}
 
 }
--- a/ush/machine/hera.sh
+++ b/ush/machine/hera.sh
@@ -13,7 +13,11 @@ function file_location() {
   case ${external_model} in
 
     "FV3GFS")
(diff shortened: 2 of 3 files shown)


venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1022938469/20220810165016/ufs-srweather-app
Build was Successful
Rocoto jobs started
ci-hera-intel-WE

venitahagerty:
Machine: jet
Compiler: intel
Job: WE
Repo location: /lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1022938469/20220810165015/ufs-srweather-app
Build was Successful
Rocoto jobs started
ci-jet-intel-WE

JeffBeck-NOAA (approved):

christinaholtNOAA (changes requested):
I am curious why this is the mechanism that is being used instead of the staged_data_dir. I was under the impression that EXTRN_MDL_SYSBASEDIR_* was meant to identify supported real-time data streams.

EDIT: I think that setting the ush/config.community.sh to find the staged disk files in the same way the test configs do it would be a better solution. When these particular env variables are set for a platform, the scripts makes certain assumptions about the directory structure below -- that it's like those supported in real time.

natalie-perlin (approved):
Verified all the data locations on Cheyenne, Hera, Orion.

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1022938469/20220818165012/ufs-srweather-app
Build was Successful
Rocoto jobs started
ci-hera-intel-WE

venitahagerty:
Machine: jet
Compiler: intel
Job: WE
Repo location: /lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1022938469/20220818165012/ufs-srweather-app
Build was Successful
Rocoto jobs started
ci-jet-intel-WE

mkavulich:
@christinaholtNOAA 
This problem all came about because the config.community.sh case just fails on most platforms, because while the data is staged, there is no logic that points the workflow to look there by default. All the documentation says about "EXTRN_MDL_SYSBASEDIR_ICS" is that it's the "Base directory on the local machine containing external model files for generating ICs on the native grid. " Doesn't say anything about realtime streams. So if that is the case, it needs more definition. Based on the current info, on platforms *without* realtime streams, I thought it made sense to just point to the staged data so that this default case will work.

I am curious about your edit comment "I think that setting the ush/config.community.sh to find the staged disk files in the same way the test configs do it would be a better solution." Do you think that we should introduce platform-specific logic to this config file? I think that would not work well as an "example" config.sh file.

christinaholtNOAA:
@mkavulich I think it's worth documenting the assumption I just made if others are not opposed. Taking up this mechanism with staged data for canned cases seems redundant and provides no alternative for folks who want to know where supported data streams are on platforms where they may like to run in near-real time.

Jet, Hera, and potentially some user-defined cloud platforms will have data streams that support real time runs. 

I think that using the "staged file" approach wouldn't need machine-specific logic since all the supported platforms provide the necessary information under `TEST_EXTRN_MDL_SOURCE_BASEDIR`. You could continue to leave a machine-specific example in as a commented guideline if you wanted, I think. Or tell users to change this directory to one they choose. In all likelihood, real Community users are going to stage their own data anyway, and will need to know how to do that.

One more thought....a good life motto: if it's not tested, assume it's broken. That's exactly why we have this PR. Is there a possibility that users could be directed to the 80+ examples of how to do this using the currently tested WE2E tests, and this one could disappear altogether. Alternatively, can this example script (and its NCO mode counterpart) be tested as a WE2E test so that it doesn't end up broken for extended periods until a user needs it?

panll (approved):

mkavulich:
We will continue this discussion after the merge; see https://github.com/ufs-community/ufs-srweather-app/issues/341

PR #823, opened 2022-08-12
Bugfix with machine file wrapper.

danielabdi-noaa:
## DESCRIPTION OF CHANGES: 
Machine file wrapper `source_machine_file.sh` uses `save/restore_shell_opts` but those are not defined when it is called from 
`load_module_run_task` and `launch_FV3LAM_wflow` because bash utilities are sourced after the machine file.
A simple change in order of sourcing these files should fix it. After this fix, i was able to use `set -e` for all scripts.

## TESTS CONDUCTED: 
Run one test on hera manually, will run automated test later.

### DIFF
--- a/ush/launch_FV3LAM_wflow.sh
+++ b/ush/launch_FV3LAM_wflow.sh
@@ -95,9 +95,9 @@ fi
 #-----------------------------------------------------------------------
 #
 . $exptdir/var_defns.sh
+. $USHDIR/source_util_funcs.sh
 . $USHDIR/source_machine_file.sh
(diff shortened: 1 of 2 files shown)


venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1025384393/20220812212017/ufs-srweather-app
Build was Successful
Rocoto jobs started
ci-hera-intel-WE

gsketefian (approved):

Issue #825, opened 2022-08-24
Multiple issues related to METplus conf files


michelleharrold:
# Description

In PointStat_upper_air_mean.conf verification for DPT at 700 hPa is set incorrectly:

FCST_VAR5_NAME = DPT_P850_ENS_MEAN
FCST_VAR5_LEVELS = P700
FCST_VAR5_THRESH = >=263, >=286, >=273

In addition, while fixing this, it was discovered the VARn_LEVELS was inconsistently set in the mean and prob conf files for METplus grid-to-point verification. This issue doesn't affect output (MET doesn't actually read the level in this case), but it could lead to confusion down the road.

## Steps to Reproduce

Run the workflow in its current state, the the 700 hPa dew point verification for ensemble mean metrics will actually be for 850 hPa!

## Additional Context

None.

## Output

None.

mkavulich:
Resolved by #826

PR #826, opened 2022-08-24
Mods to METplus vx to address multiple issues

michelleharrold:
## DESCRIPTION OF CHANGES: 
Multiple changes were made to METplus configuration files as well as the ex-scripts for verification. In multiple testing and evaluation projects using the verification, several bugs and/or enhancements were recently discovered and are addressed in this PR:

1) Ensemble mean verification for 700 hPa DPT was incorrect (was actually verifying 850 hPa). Fixed this in PointStat_upper_air_mean.conf.
2) Variable level was being inconsistently set in the grid-to-point configuration files between mean probabilistic vx. MET doesn't actually use the level in this case, but it can lead to confusion. Changed levels in PointStat_conus_sfc_mean.conf, PointStat_conus_sfc_prob.conf, PointStat_upper_air_mean.conf, and PointStat_upper_air_prob.conf to be consistent.
3) TCDC was not being specified enough, so to ensure the correct field is being used, updates to the "options" were included: GRIB2_ipdtmpl_val=[255]
4) MET occasionally errors due to new directories being created inconsistently in the wrapper logic (see METplus [issue](https://github.com/dtcenter/METplus/issues/1657) for more details). To get around the errors, METplus output directories are now created in the ex-scripts.

## TESTS CONDUCTED: 
Ran WE2E tests on Hera for deterministic and ensemble verification. Inspected output logs and MET output.

## DEPENDENCIES:
N/A

## DOCUMENTATION:
N/A

## ISSUE (optional): 
Encompasses bug in Issue #825


scripts/exregional_run_ensgridvx_mean.sh
```diff
@@ -115,6 +115,21 @@ else
...
 fi
 
+#
+#-----------------------------------------------------------------------
+#
+# Create the directory(ies) in which MET/METplus will place its output
+# from this script.  We do this here because (as of 20220811), when
+# multiple workflow tasks are launched that all require METplus to create
+# the same directory, some of the METplus tasks can fail.  This is a
+# known bug and should be fixed by 20221000.  See https://github.com/dtcenter/METplus/issues/1657.
```
JeffBeck-NOAA: @gsketefian, did you mean 20221001 here?

gsketefian: Yes, I meant 20221001 (see discussion [here](https://github.com/dtcenter/METplus/discussions/1733)).  But as I think a bit more, it would probably be better if we take out this long comment and replace with just this:
```
#
#
# Make sure directories in which output files will be placed exist.
#
#
```
or something similar.  Would have to make this change in all the relevant scripts.

michelleharrold: I will go ahead and make/push that change to relevant files!

gsketefian:
@michelleharrold @JeffBeck-NOAA I just remembered/realized that some (many?) of the `mkdir -p` commands that @michelleharrold got from my version of the scripts on disk (that are specific to the ensemble task) may not work for the `develop` branch.  That's because I had to change `OUTPUT_BASE` (and other variables too) in several places in order to make the time-lagging work.  @michelleharrold, did you make any adjustments to the `mkdir` commands or are they as-is from my version?

michelleharrold:
@gsketefian - ah yes - I forgot to mention that - I did make changes to account for the fact that develop is slightly different than the ens_design_RRFS branch. It should be good to go.

JeffBeck-NOAA (approved):

ush/templates/parm/metplus/PointStat_conus_sfc_prob.conf
```diff
@@ -185,95 +185,95 @@ POINT_STAT_ONCE_PER_FIELD = False
 # Note: If FCST_VAR<n>_* is set, then a corresponding OBS_VAR<n>_* variable must be set
 # To use one variables for both forecast and observation data, set BOTH_VAR<n>_* instead
 FCST_VAR1_NAME = TMP_Z2_ENS_FREQ_ge268
-FCST_VAR1_LEVELS = (*,*)
+FCST_VAR1_LEVELS = Z2
```
gsketefian, line 188: @michelleharrold Out of curiosity, what was `(*,*)` doing previously?

scripts/exregional_run_ensgridvx_prob.sh
```diff
@@ -115,6 +115,15 @@ else
...
 fi
 
+#
+#-----------------------------------------------------------------------
+#
+# Make sure directory in which output files will be placed exist.
+#
+#-----------------------------------------------------------------------
+#
+mkdir_vrfy -p "${EXPTDIR}/${CDATE}/metprd/ensemble_stat_prob"  # Output directory for grid_stat tool.
```
gsketefian, line 125: This should be `.../grid_stat_prob`, no?  For example, the conf file `GridStat_APCP01h_prob.conf` runs only the GridStat tool and puts its output in `.../grid_stat_prob` (at least in my version of the workflow).

scripts/exregional_run_ensgridvx_mean.sh
```diff
@@ -115,6 +115,15 @@ else
...
 fi
 
+#
+#-----------------------------------------------------------------------
+#
+# Make sure directory in which output files will be placed exist.
+#
+#-----------------------------------------------------------------------
+#
+mkdir_vrfy -p "${EXPTDIR}/${CDATE}/metprd/ensemble_stat_mean"  # Output directory for grid_stat tool.
```
gsketefian, line 125: Should be `.../grid_stat_mean`?

scripts/exregional_run_enspointvx_mean.sh
```diff
@@ -108,6 +108,16 @@ export fhr_list
...
 
+#
+#-----------------------------------------------------------------------
+#
+# Make sure directories in which output files will be placed exist.
+#
+#-----------------------------------------------------------------------
+#
+mkdir_vrfy -p "${EXPTDIR}/metprd/pb2nc"                        # Output directory for pb2nc tool.
+mkdir_vrfy -p "${EXPTDIR}/${CDATE}/metprd/ensemble_stat_mean"  # Output directory for point_stat tool.
```
gsketefian, line 119: Should this be `.../point_stat_mean`?  E.g. `PointStat_upper_air_mean.conf` only runs PointStat, not EnsembleStat.

scripts/exregional_run_enspointvx_prob.sh
```diff
@@ -108,6 +108,16 @@ export fhr_list
...
 
+#
+#-----------------------------------------------------------------------
+#
+# Make sure directories in which output files will be placed exist.
+#
+#-----------------------------------------------------------------------
+#
+mkdir_vrfy -p "${EXPTDIR}/metprd/pb2nc"                        # Output directory for pb2nc tool.
+mkdir_vrfy -p "${EXPTDIR}/${CDATE}/metprd/ensemble_stat_prob"  # Output directory for point_stat tool.
```
gsketefian, line 119: Should be `.../point_stat_prob`?

ush/templates/parm/metplus/EnsembleStat_conus_sfc.conf
```diff
@@ -215,7 +215,7 @@ ENS_VAR3_OPTIONS = GRIB2_pdt = 0; ;; derive instantaneous 10-m wind from U/V com
 ENS_VAR4_NAME = TCDC
 ENS_VAR4_LEVELS = L0
 ENS_VAR4_THRESH = <25, >75
-ENS_VAR4_OPTIONS = GRIB_lvl_typ = 200; GRIB2_ipdtmpl_index = [ 14 ]; GRIB2_ipdtmpl_val = [ 0 ]; interp = { type = [ { method = NEAREST; width = 1; } ]; }
+ENS_VAR4_OPTIONS = GRIB_lvl_typ = 200; GRIB2_ipdtmpl_index=[27]; GRIB2_ipdtmpl_val=[255]; interp = { type = [ { method = NEAREST; width = 1; } ]; }
```
gsketefian, line 218: @michelleharrold In the `ens_design_RRFS` branch, there aren't any `ENS_VAR...` variables in this conf file.  Maybe you removed them at some point because they're redundant?  (That's just the way I inherited that branch.)

ush/templates/parm/metplus/PointStat_conus_sfc_prob.conf
gsketefian: Also, should all these changes in the conf files also go into other branches (e.g. `ens_design_RRFS`) before we make any new runs?

JeffBeck-NOAA: @gsketefian, yeah, we'll want to get these pushed to ens_design_RRFS to fix similar problems there too.  That might have to wait until after the merge though.

scripts/exregional_run_ensgridvx_prob.sh
michelleharrold: @gsketefian - in the ens_design_RRFS branch it is grid_stat_prob, grid_stat_mean, point_stat_prob, and point_stat_mean, but in develop, it is still just ensemble_stat_mean and ensemble_stat_prob (where output from grid-stat and point-stat get output into the same directory). When I made the I/O changes for METplusand added gen-ens-prod, I decided to split up the output directories to be a little more intuitive.

ush/templates/parm/metplus/EnsembleStat_conus_sfc.conf
michelleharrold: @gsketefian - Yes, you inherited that way (as part of the gen-ens-prod and ensemble-stat split)! Gen-ens-prod is the new MET tool that handles the creation of the ensemble fields, whereas ensemble-stat used to cover both the creation of the ensemble fields AND the ensemble statistic generation. With the introduction of gen-ens-prod, these ENS_* fields are no longer necessary in the ensmeble-stat METplus conf files.

ush/templates/parm/metplus/PointStat_conus_sfc_prob.conf
michelleharrold: @gsketefian - the `(*,*)` is the default way for MET to read netCDF data (where it is reading in all data). John HG wrote up an [MET issue](https://github.com/dtcenter/MET/issues/2235) related to this that explains it well. We chose to switch to something like Z2, e.g., because it is more intuitive to see and matches what the level is in the netCDF file.

scripts/exregional_run_ensgridvx_mean.sh
gsketefian: Thanks for simplifying!

scripts/exregional_run_ensgridvx_prob.sh
gsketefian: Oh I see.  I will resolve all my comments about this then.  Thanks.

ush/templates/parm/metplus/EnsembleStat_conus_sfc.conf
gsketefian: Ok, makes sense, thanks!

ush/templates/parm/metplus/PointStat_conus_sfc_prob.conf
gsketefian: @michelleharrold Ok, I understand after reading that issue.

Let me know if it would help to include the `(*,*)` related changes in the `ens_design_RRFS` branch as well.  Seems like these are just improvements (not  bug fixes), but it would be nice to be running the best code possible for the ensemble task.

ush/templates/parm/metplus/PointStat_upper_air_mean.conf
```diff
@@ -216,7 +216,7 @@ OBS_VAR4_NAME = DPT
 OBS_VAR4_LEVELS = P850
 OBS_VAR4_THRESH = >=273, >=278, >=283
 
-FCST_VAR5_NAME = DPT_P850_ENS_MEAN
+FCST_VAR5_NAME = DPT_P700_ENS_MEAN
```
gsketefian, line 219: Looks like just the name was wrong, but it was actually using 700?

gsketefian (approved):

